PR #1, opened 2016-01-20
update README.md to keep consistent with generate.go and MAKEFILE

wangkirin:
current generate.go support  the generation of more `specs` items than listed in README , and in MAKEFILE we have the install function , this PR  keep README consistent with generate.go and MAKEFILE

Signed-off-by: Wang Qilin <qilin.wang@huawei.com>


mrunalp:
Thanks! LGTM.

PR #2, opened 2016-01-20
generate: Drop devices and mounts

wking:
Cross-posted from mrunalp/ocitools#15.

Most of these are covered by the requirements that landed in
opencontainers/specs#164, and I don't think the additional mqueue
mount is worth the space between the spec requirements and the
ocitools-generated configs.  More details in the commit messages.

There was some [discussion in the 2016-01-20 meeting][1] about the
meaning of lists that didn't include default entries, but I think
we're all on the same page that unset values mean “give me the
defaults”.  So this should be mergable without waiting for that
discussion to reach consensus.

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-01-20-18.02.log.html#l-43

### DIFF
--- a/generate.go
+++ b/generate.go
@@ -619,32 +619,7 @@ func getDefaultTemplate() specs.LinuxSpec {
 				Cwd: "/",
 			},
 			Hostname: "shell",
-			Mounts: []specs.Mount{
-				{
-					Destination: "/proc",
-					Type:        "proc",
-					Source:      "proc",
-					Options:     nil,
-				},
-				{
-					Destination: "/dev/pts",
-					Type:        "devpts",
-					Source:      "devpts",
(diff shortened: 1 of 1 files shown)


mrunalp:
This one will be ready to go once we update the spec dependency here as runc is making these changes.

mrunalp:
Could you rebase this PR? Thanks!

wking:
On Mon, Mar 07, 2016 at 04:48:29PM -0800, Mrunal Patel wrote:
> Could you rebase this PR?

Rebased with 05c3827 → 080ee00.  It looks like the mqueue entry was
removed in 0fbe252 (update generate.go to 0.3.0, 2016-02-23, #20), so
this PR is down to a single commit.

mrunalp:
Thanks. LGTM.

PR #4, opened 2016-01-21
add bundle validate, using reflect to check omitempty

liangchenye:

Based on previous PR (https://github.com/mrunalp/ocitools/pull/10).
1. correct rootfs reading problem
 way to try it: 
    make a bundle 'testbundle for example' by ocitools generate & tar xzvf rootfs.tar.gz
    ocitools bvalidate --path=testbundle

2. add 'reflect' to check 'omitempty' in spec/*go file
    open it once spec version in Godeps updated with opencontainer/specs


bvalidate.go
```diff
@@ -0,0 +1,395 @@
...
+		}
+
+		defer sf.Close()
+
+		var spec specs.LinuxSpec
+		if err = json.NewDecoder(sf).Decode(&spec); err != nil {
+			logrus.Fatal(err)
+		} else {
+			if spec.Platform.OS != "linux" {
+				logrus.Fatalf("Operation system '%s' of the bundle is not support yet.", spec.Platform.OS)
```
mrunalp: s/support/supported/

bvalidate.go
```diff
@@ -0,0 +1,395 @@
...
+		var runtime specs.LinuxRuntimeSpec
+		if err = json.NewDecoder(rf).Decode(&runtime); err != nil {
+			logrus.Fatal(err)
+		}
+
+		rootfsPath := path.Join(inputPath, spec.Root.Path)
+		if fi, err := os.Stat(rootfsPath); err != nil {
+			logrus.Fatalf("Cannot find the rootfs: %v", rootfsPath)
+		} else if !fi.IsDir() {
+			logrus.Fatalf("Rootfs :%v is not a directory.", spec.Root.Path)
```
mrunalp: Rootfs:

bvalidate.go
```diff
@@ -0,0 +1,395 @@
...
+		}
+		bundleValidate(spec, runtime, rootfsPath)
+		logrus.Infof("Bundle test succeeded.")
```
mrunalp: s/test/validation/

bvalidate.go
```diff
@@ -0,0 +1,395 @@
...
+	} else if !isStruct(objT) {
+		return nil, true
+	}
+
+	valid = true
+	for i := 0; i < objT.NumField(); i++ {
+		t := objT.Field(i).Type
+		if isStructPtr(t) && objV.Field(i).IsNil() {
+			if !strings.Contains(objT.Field(i).Tag.Get("json"), "omitempty") {
+				msgs = append(msgs, fmt.Sprintf("'%s.%s' is should not be empty", objT.Name(), objT.Field(i).Name))
```
mrunalp: s/is should/should

bvalidate.go
```diff
@@ -0,0 +1,395 @@
...
+			}
+		}
+
+	}
+	return msgs, valid
+}
+
+func CheckMandatoryField(obj interface{}) {
+	if msgs, valid := checkMandatory(obj); !valid {
+		logrus.Fatalf("Mandatory information required: %s.", msgs)
```
mrunalp: Mandatory configuration missing:

bvalidate.go
```diff
@@ -0,0 +1,395 @@
...
+		"RLIMIT_SIGPENDING",
+		"RLIMIT_MSGQUEUE",
+		"RLIMIT_NICE",
+		"RLIMIT_RTPRIO",
+		"RLIMIT_RTTIME",
+	}
+)
+
+var bundleValidateCommand = cli.Command{
+	Name:  "bvalidate",
```
mrunalp: Let us use validate for this. I have removed existing unused code in #14

liangchenye:
Replace original validate.go with new one (bvalidate.go).

@mrunalp @zenlinTechnofreak 
so runtimevalidate will not be a sub-command but a test framework, right?

zenlinTechnofreak:
@liangchenye @mrunalp  I think it is nice to develop tools seperately, anyway, our repo named tools.
If in that way, I should rebase my PR  Add initial version of runtimetest command: #9  to a seperate tool, but not a sub command.

What is your design in the ocitools repo? It  should be nice if you can let me know. Thanks very much.

liangchenye:
@zenlinTechnofreak my opinion is we should provides three official sub-command in ocitools.
1. generate
   -- generate bundle according to user's input
2. bundle validate
  -- validate if a bundle is valid
   once the 'layered' bundle released, more validating codes will needed.
3. runtime validate
  -- validate if a runtime is valid
   pass all the test cases you provided.
   it is very similar to test_runtime.sh,  the original requirement is https://github.com/mrunalp/ocitools/issues/8

Other command/program/script like get-stage3.sh is important, but maybe suitable to put into a 'contrib' directory. @wking

wking:
On Tue, Jan 26, 2016 at 02:23:16AM -0800, 梁辰晔 (Liang Chenye) wrote:
> Other command/program/script like get-stage3.sh is important, but
> maybe suitable to put into a 'contrib' directory. @wking

get-stage3.sh is certainly internal tooling and not part of this
package's public interface.  Feel free to put it wherever you like ;).

mrunalp:
@liangchenye Yes, having these 3 commands makes sense.

ChengTiesheng:
@liangchenye @zenlinTechnofreak @mrunalp @wking,  a more command maybe needed: oci convert, which converts oci bundle to other container images(e.g. docker image or ACI image), any thoughts?

wking:
On Tue, Jan 26, 2016 at 05:53:44PM -0800, Tiesheng wrote:
> oci convert, which converts oci bundle to other container
> images(e.g. docker image or ACI image), any thoughts?

Conversions like that are going to be complicated and lossy (at least
in one direction, it seems unlikely that the feature sets are
identical for two tools).  I'd rather keep conversion tools like that
in different repositories to cut down on noise in the
conformance-testing tool.  But I agree that those tools will be
useful, so +1 to putting them under opencontainers/ if you can attract
enough OCI-member support for them.

liangchenye:
@ChengTiesheng I'll create a new PR to add 'contrib' directory.
The convert tools could be add to that directory.

liangchenye:
Use bundleValidateCommand (bvalidate) as the sub-command.

@zenlinTechnofreak  maybe you can use 
  runtimeValidateCommand(rvalidate) as another sub-command to do the runtime test.

PR #5, opened 2016-01-21
Use bash for get-stage3.sh

hqhq:
I got the following error with `make rootfs.tar.gz`:
./get-stage3.sh: 43: ./get-stage3.sh: Bad substitution

Change to bash fixed it.

Signed-off-by: Qiang Huang <h.huangqiang@huawei.com>

### DIFF
--- a/get-stage3.sh
+++ b/get-stage3.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
 # Download the current Gentoo stage3
 #


wking:
On Thu, Jan 21, 2016 at 12:05:08AM -0800, Qiang Huang wrote:
> I got the following error with `make rootfs.tar.gz`:
> ./get-stage3.sh: 43: ./get-stage3.sh: Bad substitution
> 
> Change to bash fixed it.

Ah, my bad.  I think we can do:

  CURRENT="${FILE/${DATE}/current}"

without leaving POSIX though (with ‘sed’ as a last resort).  I'll try
to work up a PR.

mrunalp:
Closing in favor of #8 @hqhq please retry and report if any issues. Thanks!

PR #13, opened 2016-01-22
Add make test for Makefile and use make test to fix problems.

ChengTiesheng:
Users can use make test to check gofmt/govet/golint errors.

Signed-off-by: ChengTiesheng <chengtiesheng@huawei.com>


mrunalp:
Thanks! LGTM

PR #17, opened 2016-01-27
add contrib directory and mv get-stage3 to that

liangchenye:


liangchenye:
add contrib and mv get-stage3 to a directory called 'rootfs-gentoo-builder'  (https://github.com/opencontainers/ocitools/pull/4)

@wking not sure if you could accept the name of 'rootfs-gentoo-builder'  :)

wking:
On Wed, Jan 27, 2016 at 01:11:18AM -0800, 梁辰晔 (Liang Chenye) wrote:
> @wking not sure if you could accept the name of
> 'rootfs-gentoo-builder' :)

I don't mind, but unless we expect to have multiple builders (and I
can't see why we'd need them), I'd just call the directory
‘rootfs-builder’.  Then the implementation can change from being
Gentoo-based to whatever without consumers needing to care.

mrunalp:
LGTM

PR #21, opened 2016-03-10
update Godeps to specs-go version; modify codes accordingly

liangchenye:
Update to new version (with specs-go directory)

Works in my test, and find the 'device' filed should be 'omitempty'.


liangchenye:
Also submit PR to add 'omitempty' to 'device'.
https://github.com/opencontainers/specs/pull/340

liangchenye:
update to 0.4.0

mrunalp:
Merging this. We need a follow on to fix generate. I will add that.

PR #23, opened 2016-03-16
Add Maintainer

liangchenye:
Since this project is based on ocitools and OCT, 
I propose to add Mrunal and me as the initial maintainers. 

There are other contributers  who also makes great work on it,
I think we can add more since ocitools will become bigger :) 
Signed-off-by: liangchenye liangchenye@huawei.


mrunalp:
Sounds good. We can add more maintainers per their contributions. Thanks!

PR #28, opened 2016-04-05
use opencontainer/runtime-spec

liangchenye:
Since OCI has two spec projects now, update ocitools by using opencontainers/runtime-spec and use 'rspecs' for a better differentiation. 


bvalidate.go
```diff
@@ -11,7 +11,7 @@ import (
 
 	"github.com/Sirupsen/logrus"
 	"github.com/codegangsta/cli"
-	"github.com/opencontainers/specs/specs-go"
+	rspecs "github.com/opencontainers/runtime-spec/specs-go"
```
mrunalp: Maybe call it `rspec` instead of `rspecs`? Otherwise looks good. Thanks!

mrunalp:
LGTM. Thanks!

PR #29, opened 2016-04-06
Add support for sysctls to generate

mrunalp:

### DIFF
--- a/generate.go
+++ b/generate.go
@@ -45,6 +45,7 @@ var generateFlags = []cli.Flag{
 	cli.StringFlag{Name: "cwd", Value: "/", Usage: "current working directory for the process"},
 	cli.StringSliceFlag{Name: "uidmappings", Usage: "add UIDMappings e.g HostID:ContainerID:Size"},
 	cli.StringSliceFlag{Name: "gidmappings", Usage: "add GIDMappings e.g HostID:ContainerID:Size"},
+	cli.StringSliceFlag{Name: "sysctl", Usage: "add sysctl settings e.g net.ipv4.forward=1"},
 	cli.StringFlag{Name: "apparmor", Usage: "specifies the the apparmor profile for the container"},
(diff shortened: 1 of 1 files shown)


liangchenye:
https://groups.google.com/forum/#!topic/golang-announce/JMs4_CGbXWk
```
Russ Cox 	Mar 1
Re: [golang-dev] Re: A friendly heads-up: Deletion of Go1.4-dependent code in x/tools repo
Other recipients: k.shapo...@gmail.com, golan...@googlegroups.com, g...@google.com

On Tue, Mar 1, 2016 at 5:53 AM, Konstantin Shaposhnikov <k.shapo...@gmail.com> wrote:
Will cmd/vet be deleted as well?
I've seen quite a lot of projects that use Go 1.4 and "go get" vet from the x/tools repository during the build.

Since it depends on those, yes, it will.

Go 1.4 stopped being supported when Go 1.5 was released six months ago. (See https://golang.org/doc/devel/release.html#policy for details.)

Because we knew this deletion would affect use cases such as the one you mentioned, we waited an extra six months (one release cycle) to give users additional time to update, and we pre-announced the change in October (see Robert's original email for the link). That's a significant grace period with significant advance notice.

We can't keep the code running indefinitely. People who need to keep running Go 1.4 can in the worst case make their own copy of the code, but we'd also love to hear from them about why they're still on Go 1.4 so that we can address those reasons.

Thanks.
Russ
```

liangchenye:
Go community suggests to update the usage of `vet` package.

liangchenye:
Seems the x/tools repo was removed just between PR#28 and this one.
Fixed in https://github.com/opencontainers/ocitools/pull/28.

PS runc also affected by this.

mrunalp:
@liangchenye Thanks for the info! I have merged your PR and rebuilt this one. I have also opened a PR to fix this in runc.

liangchenye:
LGTM

PR #30, opened 2016-04-09
Add man pages for ocitools

rhatdan:
Also fix up Makefile so that it will install ocitools and man pages.

### DIFF
--- a/man/ocitools-generate.1.md
+++ b/man/ocitools-generate.1.md
@@ -1,0 +1,365 @@
+% OCI(1) Oci User Manuals
+% OCI Community
+% APRIL 2016
+# NAME
+ocitools-generate - Generate a OCI spec file
+
+# SYNOPSIS
+**ocitools generate**
+[**--arch**[=*[]*]
+[**--apparmor**[=*[]*]]
+[**--args**[=*[]*]]
+[**--bind**[=*[]*]]
+[**--cap-add**[=*[]*]]
+[**--cap-drop**[=*[]*]]
+[**--cwd**[=*[]*]]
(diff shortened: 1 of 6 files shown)


man/ocitools-generate.1.md
```diff
@@ -0,0 +1,356 @@
...
+
+Then exit and check the journal.
+
+    # exit
+
+    # journalctl -b | grep Hello
+
+This should list the message sent to logger.
+
+## Attaching to one or more from STDIN, STDOUT, STDERR
```
mrunalp: This does not exist in ocitools.

man/ocitools-generate.1.md
```diff
@@ -0,0 +1,356 @@
...
+```
+ $ ocitools generate -it --ipc=container:ed735b2264ac shm ipcs -m
+ $ sudo ipcs -m
+
+ ------ Shared Memory Segments --------
+ key        shmid      owner      perms      bytes      nattch     status
+ 0x0000162e 0          root       666        27         1
+```
+
+## Mapping Ports for External Usage
```
mrunalp: Also, ports have to be handled at a higher level. We can't manage them in ocitools.

rhatdan: I was just cutting and pasting docker-run man page.  On an early Saturday/Sunday morning. Will remove these sections.

rhatdan:
Fixed up the SELInux example.

rhatdan:
Removed the bogus examples.

BTW Why did you call it bvalidate rather then validate?

mrunalp:
I think we can rename it.

PR #32, opened 2016-04-11
Add bash completions

rhatdan:


completions/bash/ocitools
```diff
@@ -0,0 +1,389 @@
...
+
+# suppress trailing whitespace
+__ocitools_nospace() {
+	# compopt is not available in ancient bash versions
+	type compopt &>/dev/null && compopt -o nospace
+}
+
+__ocitools_complete_propagations() {
+	COMPREPLY=( $( compgen -W "
+		PRIVATE
```
mrunalp: we are using lowercase here. I can change it but will do it after this goes in.

mrunalp:
LGTM

PR #35, opened 2016-04-13
Add support for using a base template

mrunalp:

### DIFF
--- a/generate.go
+++ b/generate.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"runtime"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


mrunalp:
@liangchenye @rhatdan PTAL

liangchenye:
LGTM

rhatdan:
Why not make the default template use this method and ship it separate from the runc implementation.

mrunalp:
@rhatdan That would mean going elsewhere to get the first template. With this approach one can start by creating a template and then use it as the base for other templates.

rhatdan:
How do I see the default template other then reading the code?  Can I easily generate the default template so I can edit it.

mrunalp:
@rhatdan Just running ocitools generate will give you the default template.

rhatdan:
Ok I guess we should document this, If you want to customize the default template, run ostree generate.  Edit config.json, or run ocitools to create a config.json, which you want as your customer "template."

mrunalp:
Yeah, I will follow up with a PR with updated documentation.

PR #37, opened 2016-04-18
Only set boolean parameters if they are passed as flags

mrunalp:

### DIFF
--- a/generate.go
+++ b/generate.go
@@ -121,7 +121,9 @@ func loadTemplate(path string) (spec *rspec.Spec, err error) {
 
 func modify(spec *rspec.Spec, context *cli.Context) error {
 	spec.Root.Path = context.String("rootfs")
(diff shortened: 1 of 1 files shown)


liangchenye:
I think we do not need to check the bool flag before use it. If a bool flag is not set, context.Bool will return `false` as a default.

mrunalp:
This is for when we set a bool to true in a template and then try to use it.
For e.g.
```
# Create a config.json to be used as template with terminal true
$ ocitools generate --tty
$ mv config.json config.json.template

# Try using that template to create another config.json
# Without this PR, the terminal will be set to false even though it wasn't specified.
$ ocitools generate --args /bin/bash --template config.json.template

liangchenye:
Make sense.
LGTM

mrunalp:
Thanks!

PR #42, opened 2016-04-28
man/ocitools-generate: Use [OPTIONS] in the synopsis

wking:
There are so many of these, I think it's time to follow grep(1) and
friends and punt the individual details to the OPTIONS section.

I stared off trying to adjust the metavars (to use the [Python
nomenclature][1]) to match the OPTIONS section, but there are so many
;).  And useful metavars in the synopsis would make the synopsis even
more unwieldy than it already was.

[1]: https://docs.python.org/3/library/argparse.html#metavar


mrunalp:
LGTM

PR #47, opened 2016-04-28
README: Punt detailed command description to the man pages

wking:
This makes it easier to get a project overview from the README, and
avoids the tedium of keeping parallel sets of detailed docs up to
date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,8 @@ ocitools is a collection of tools for working with the [OCI runtime specificatio
 
 ## Generating an OCI runtime spec configuration files
 
-[`ocitools generate`][generate.1] is used to generate a `config.json` ([OCI spec][runtime-spec] file) to be used to instantiate an OCI container.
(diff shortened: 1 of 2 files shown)


liangchenye:
It looks much better. 
TODO: We can add the runtime test description to the `Testing OCI runtimes` part just like what you to do generete and validate.

LGTM

wking:
On Wed, Apr 27, 2016 at 11:55:42PM -0700, 梁辰晔 (Liang Chenye) wrote:
“TODO: We can add the runtime test description to the `Testing OCI
runtimes` part just like what you to do generete and validate.”

I think it would be better to make test_runtime.sh a first-class
citizen (‘ocitools certify’?) with a man page and everything.  But
that seemed out of scope for this PR ;).

liangchenye:
Yes. And we need to discuss with @zenlinTechnofreak about his PR and his busy these day aroud.

mrunalp:
@wking Agree that we need to convert the script into a first class option. Just needs to be done :)
LGTM

PR #48, opened 2016-04-28
generate: Change --mount-cgroups default to 'no'

wking:
Since a937f7a4 (Add support for using a base template, 2016-04-13,
`--mount-cgroups` is not `no`, that will lead to additional cgroup
entries in the mount array (even if there was already a cgroup entry
in the mount array).  Instead of writing logic to find and clobber
existing cgroup entries, just make the default the no-op setting.


wking:
Looks like my commit message is missing a line.  Will fix once I get back to a keyboard.

mrunalp:
LGTM

wking:
Restored missing line with 8819fe2 → 87101c0, and adjusted my `core.commentchar` to avoid future occurrences ;).

wking:
Ah, too late ;).  Oh, well.

PR #49, opened 2016-04-28
validate: add Hooks validation

Mashimiao:

### DIFF
--- a/man/ocitools-validate.1.md
+++ b/man/ocitools-validate.1.md
@@ -18,6 +18,9 @@ Validate an OCI bundle
 **--path=PATH
    Path to bundle
 
+**--hooks**
+  Check specified hooks exist and are executable on the host.
+
 # SEE ALSO
 **ocitools**(1)
 
--- a/validate.go
+++ b/validate.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"path/filepath"
 	"reflect"
 	"regexp"
 	"strings"
@@ -19,6 +20,7 @@ import (
 
 var bundleValidateFlags = []cli.Flag{
 	cli.StringFlag{Name: "path", Usage: "path to a bundle"},
+	cli.BoolFlag{Name: "hooks", Usage: "Check specified hooks exist and are executable on the host."},
 }
 
 var (
@@ -76,17 +78,19 @@ var bundleValidateCommand = cli.Command{
 			logrus.Fatalf("root path %q is not a directory.", spec.Root.Path)
 		}
 
-		bundleValidate(spec, rootfsPath)
+		hooksCheck := context.Bool("hooks")
+		bundleValidate(spec, rootfsPath, hooksCheck)
 		logrus.Infof("Bundle validation succeeded.")
 	},
 }
 
-func bundleValidate(spec rspec.Spec, rootfs string) {
+func bundleValidate(spec rspec.Spec, rootfs string, hooksCheck bool) {
 	checkMandatoryField(spec)
 	checkSemVer(spec.Version)
 	checkPlatform(spec.Platform)
 	checkProcess(spec.Process, rootfs)
 	checkLinux(spec.Linux, spec.Hostname, rootfs)
+	checkHooks(spec.Hooks, hooksCheck)
 }
 
 func checkSemVer(version string) {
@@ -118,6 +122,30 @@ func checkPlatform(platform rspec.Platform) {
 		}
 	}
 	logrus.Fatalf("Operation system %q of the bundle is not supported yet.", platform.OS)
+}
+
+func checkHooks(hooks rspec.Hooks, hooksCheck bool) {
+	checkEventHookPaths("pre-start", hooks.Prestart, hooksCheck)
+	checkEventHookPaths("post-start", hooks.Poststart, hooksCheck)
+	checkEventHookPaths("post-stop", hooks.Poststop, hooksCheck)
+}
+
+func checkEventHookPaths(hookType string, hooks []rspec.Hook, hooksCheck bool) {
+	for _, hook := range hooks {
+		if !filepath.IsAbs(hook.Path) {
+			logrus.Fatalf("The %s hook %v: is not absolute path", hookType, hook.Path)
+		}
+
+		if hooksCheck {
+			fi, err := os.Stat(hook.Path)
+			if err != nil {
+				logrus.Fatalf("Cannot find %s hook: %v", hookType, hook.Path)
+			}
+			if fi.Mode()&0111 == 0 {
+				logrus.Fatalf("The %s hook %v: is not executable", hookType, hook.Path)
+			}
+		}
+	}
 }
 
 func checkProcess(process rspec.Process, rootfs string) {


validate.go
```diff
@@ -106,6 +107,29 @@ func checkMounts(mounts []rspec.Mount, rootfs string) {
 	}
 }
 
+func checkHooks(process rspec.Hooks, rootfs string) {
```
wking: `process` is a strange name for this variable.  How about `hooks`?  And hooks are in the host mount namespace, so I don't think we need `rootfs`.

wking:
This sort of validation would be easier if we could recycle the
process validator [1].

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/Cn5L6g1prgA
     Subject: Unify container-process and hook-process structures
     Date: Mon, 4 Jan 2016 21:56:37 -0800
     Message-ID: <20160105055637.GG6362@odin.tremily.us>

validate.go
Mashimiao: @wking  change `process` to `hooks` is fine.
`hooks are in the host mount namespace` I'm not sure what it means.
Do you mean hooks is not in OCI bundle?

wking: On Thu, Apr 28, 2016 at 07:54:26PM -0700, Ma Shimiao wrote:
> `hooks are in the host mount namespace` I'm not sure what it means.
> Do you mean hooks is not in OCI bundle?

The may be in the bundle, but if they are, the path will be
/path/to/bundle/and/hook, not /and/hook.  Specs are [1]:

  “Hook paths are absolute and are executed from the host's
  filesystem.”

Although I'd be happier if the wording used “runtime namespace” [2] so
it was explicit about other namespaces as well (e.g. hooks are also in
the host network namepace, the host ipc namespace, …).

[1]: https://github.com/opencontainers/runtime-spec/blob/v0.5.0/config.md#hooks
[2]: https://github.com/opencontainers/runtime-spec/blob/v0.5.0/glossary.md#runtime-namespace

liangchenye:
Thanks @Mashimiao, 
if we want to check hook, we can just check `hook.Path` directly, there is no need to join the rootfs. 
```
Hook paths are absolute and are executed from the host's filesystem.
```

Also, since the hook command exists in the host's filesystem, we can not check that in bundle validation.go.   (But we reuse this code in runtime test.)

wking:
On Tue, May 03, 2016 at 02:36:04AM -0700, 梁辰晔 (Liang Chenye) wrote:
> Also, since the hook command exists in the host's filesystem, we can
> not check that in bundle validation.go.

We can check “Does the hook command exist in the host's filesystem?”,
it's just that the check will not be portable.  I think it's still
worth performing the check (e.g. see opencontainers/runtime-spec#418),
but it's probably worth putting that sort of thing behind a flag
(--host-checks?).  I don't really have a preference for the flag name
or whether it's opt-in or opt-out.

Mashimiao:
@wking  @liangchenye  
This validation, as @wking said we can check  “Does the hook command exist in the host's filesystem?”.
`config.json` belongs to OCI bundle. I think we check hook command exist or not is to check whether config.json is configured correct or not, is to check OCI bundle.
@wking  I'm not sure what portability problem exists. If there was no hooks configured, the `hook validation` would not be executed.

liangchenye:
The hook configuration is a little tricky, it is not portable.
If it was verified on hostA, it could still fail on hostB.
If it fails on hostA, it is hard to say this is an invalid bundle.
But it is helpful to a user who tries to use it, so agree with WKing, we can add a flag to bundle validate.

Mashimiao:
@wking  @liangchenye  PR updated

man/ocitools-validate.1.md
```diff
@@ -20,6 +21,9 @@ Validate an OCI bundle
 **--path="PATH"
    Path to bundle
 
+**--hooks-check**
+  Specify check hooks on host, default is false.
```
wking: I'd have gone with `--hook-paths` or some such, but I expect folks will get the idea regardless.  It's probably worth detailing what the hook checks are and why this is optional.  The only check that needs to be toggle-able is “Does the `path` exist (and point to an executable)?”, and there is other validation that you can do beyond that (e.g. “Are there equals signs in all the `env` values?”).

validate.go
```diff
@@ -72,18 +73,23 @@ var bundleValidateCommand = cli.Command{
...
+func bundleValidate(spec rspec.Spec, rootfs string, hooksCheck bool) {
 	checkMandatoryField(spec)
 	checkSemVer(spec.Version)
 	checkPlatform(spec.Platform)
 	checkProcess(spec.Process, rootfs)
 	checkMounts(spec.Mounts, rootfs)
 	checkLinux(spec.Linux, rootfs)
+
+	if hooksCheck {
+		checkHooks(spec.Hooks)
```
wking: The only check that has to be optional is [whether the paths exist][1].  If we punt on other checks (fine with me), I'd recommend `--hook-paths`, `hookPaths`, and `checkHookPaths()`, or some such.

validate.go
```diff
@@ -128,6 +134,26 @@ func checkPlatform(platform rspec.Platform) {
 	logrus.Fatalf("Operation system '%s' of the bundle is not supported yet.", platform.OS)
 }
 
+func checkHooks(hooks rspec.Hooks) {
+	for _, prestart := range hooks.Prestart {
+		if _, err := os.Stat(prestart.Path); err != nil {
+			logrus.Fatalf("Cannot find pre-start hook: %v", prestart.Path)
+		}
+	}
```
wking: Probably worth pulling this out into a helper function so it can get more complex without needing copy/pastes to the blocks handling the other hook arrays.  `checkEventHookPaths(hooks []Hook)`?  And this should also check that at least one of the three [executable bits is set][1].  I don't see constants for the bits in `ModePerm`, and I'm barely literate in Go, but `fileInfo.Mode() & 0111` is probably close.

And having validation die at the first error is a bit lazy, but fixing that is probably out of scope for this PR ;).

[1]: https://golang.org/pkg/os/#FileMode

validate.go
liangchenye: I'm fine with checkHooks. I think it looks more consistent with the functions above.

validate.go
liangchenye: Agree with the checkEventHookPaths idea.
Besides this, LGTM.

validate.go
wking: On Tue, May 10, 2016 at 06:27:30AM -0700, 梁辰晔 (Liang Chenye) wrote:
> > +	if hooksCheck {
> > +		checkHooks(spec.Hooks)
> 
> I'm fine with checkHooks. I think it looks more consistent with the
> functions above.

The additional specificity of checkHookPaths() is to support:

  if hookPaths {
    checkHookPaths(spec.Hooks)
  }
  checkHookEnvs(spec.Hooks)

and such if we want to validate other parts of the hook structure [1].
We can always punt on that distinction until later, but we'd be
breaking a public API if we rename ‘--hooks-check’ down the road, so
I'd rather pick a specific name (like ‘--hook-paths’) now [1].
There's no breakage with internal function renames, so I care less
about what we use there ;).

[1]: https://github.com/opencontainers/ocitools/pull/49/files/29a4d8e05ecc789646b65de133754273cd710f2e#r62610834

validate.go
Mashimiao: add func `checkEventHookPaths`

Mashimiao:
ping @liangchenye

man/ocitools-validate.1.md
```diff
@@ -6,6 +6,9 @@ ocitools-validate - Validate a OCI bundle
 
 # SYNOPSIS
 **ocitools validate**  *[OPTIONS]*
+[**--help**]
+[**--path**[=*PATH*]
+[**--hooks-check**[=*true*|*false*]
```
wking: Listing specific options here rolls back part of c1df0ceb (Fixup man pages for consistency, 2016-05-07, #65).  I like the shorter *[OPTIONS]* for `generate` (see #42) because it has *lots* of options.  I don't care either way for `validate`, but we probably want to pick one approach and not flip back and forth without motivation ;).

man/ocitools-validate.1.md
```diff
@@ -18,6 +21,9 @@ Validate an OCI bundle
 **--path=PATH
    Path to bundle
 
+**--hooks-check**
```
wking: I [still prefer][1] the more-explicit `--hook-paths` (but I'm not a maintainer).  Whatever the option name ends up being, I'd advise against including `-check` (so just `--hooks` if you don't want an option that mentions paths).  I'd guess almost all future options to `validate` are going to be about enabling/disabling particular checks (since checking fields is all `validate` does ;).  And in that case, the fact that a hook enables/disables a check is pretty obvious even without the `-check` qualifier.

[1]: https://github.com/opencontainers/ocitools/pull/49#discussion_r62611097

validate.go
```diff
@@ -128,6 +134,24 @@ func checkPlatform(platform rspec.Platform) {
...
+}
+
+func checkEventHookPaths(hookType string, hooks []rspec.Hook) {
+	for _, hook := range hooks {
+		fi, err := os.Stat(hook.Path)
+		if err != nil {
+			logrus.Fatalf("Cannot find %s hook: %v", hookType, hook.Path)
+		}
+		if fi.Mode()&0111 != 0 {
+			logrus.Fatalf("The %s hook %v: is not executable", hookType, hook.Path)
```
wking: I think the comparison should be `== 0` (i.e. “if none of the executable bits are set”).

man/ocitools-validate.1.md
Mashimiao: @wking  I can't agree with you.
We are not sure if there will be other options to validate  going to enable/disable particular checks, as hooks is a lit sepcial.
By now, there is only hooks need to enablle/disable a check and I think an obvious option name is better.

wking: On Sun, May 15, 2016 at 10:38:52PM -0700, Ma Shimiao wrote:
> We are not sure if there will be other options to validate going to
> enable/disable particular checks, as hooks is a lit sepcial.  By
> now, there is only hooks need to enablle/disable a check and I think
> an obvious option name is better.

Falling into this same host-path category is root.path (since
opencontainers/runtime-spec#394, currently unreleased).  So maybe the
option should be --host-paths?  Bind mounts and such can also
reference out-of-bundle paths.

I'm sure there are several features like that.  Looking for other
classes of host-specific checks, there's at least “Does the specified
platform match the local system?”.

Mashimiao: Maybe there will be other particular checks, but I still think `--hooks-check` is better and other particular checks in the future should be with `-check`.  it obviously tell people what it is used for.

liangchenye:
Thanks @Mashimiao 

LGTM

wking:
Ah, I'd missed the responses to my earlier comments.  f54700db looks
good to me too.

Mashimiao:
ping @mrunalp

liangchenye:
Hi @Mashimiao, 
I added another task to https://github.com/opencontainers/ocitools/issues/66
 'Hooks paths are absolute'.
It is different but related to this PR. We need to check if a path is absolute  by default (without hooksCheck set to true).

Mashimiao:
@liangchenye  OK, I can implement it in this PR.

wking:
51d1cc61 still looks good to me.

wking:
Ah, actually 51d1cc61 does not look good to me ;).  The absolute-path
check needs to happen regardless of whether folks have set the
host-specific switch [1].  Maybe we want to lump all of the
host-specific checks behind a --host-specific option?

[1]: https://github.com/opencontainers/ocitools/pull/49#issuecomment-220511921

validate.go
liangchenye: how about use checkHooks(spec.Hooks, hookscheck).
Since absolute path should be checked anyway.

Mashimiao:
@wking   Ah,  yes.  I did not read @liangchenye 's words carefully. How about now?

wking:
On Fri, May 20, 2016 at 01:30:20AM -0700, Ma Shimiao wrote:
> How about now?

ce49f78 looks great :).

man/ocitools-validate.1.md
```diff
@@ -18,6 +18,9 @@ Validate an OCI bundle
 **--path=PATH
    Path to bundle
 
+**--hooks**
+  Specify check hooks exists and executable on host, default is false.
```
mrunalp: nit: Check specified hooks exist and are executable on the host.

Mashimiao: fixed.

validate.go
```diff
@@ -19,6 +20,7 @@ import (
 
 var bundleValidateFlags = []cli.Flag{
 	cli.StringFlag{Name: "path", Usage: "path to a bundle"},
+	cli.BoolFlag{Name: "hooks", Usage: "Specify check hooks on host, default is false"},
```
mrunalp: Can you change it here as well? Thanks!

Mashimiao: @mrunalp  sorry, miss it. fixed.

mrunalp:
Needs rebase

Mashimiao:
@mrunalp   rebased.

mrunalp:
LGTM

PR #54, opened 2016-05-03
translate: Stub in a translation framework

wking:
Add tooling to translate higher-level configs into the basic OCI config.  On IRC, @julz floated a [`linux.namespaces[].fromContainer`](http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2016-04-27.log.html#t2016-04-27T20:32:09) as a higher-level version of
`linux.namespaces[].path` for emulating exec (opencontainers/runtime-spec#391).  That makes sense to me, with a [UI like](http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2016-04-28.log.html#t2016-04-28T16:09:51):

```
$ ocitools generate --template=high-level-config.json --translate=fromContainer --runtime=runc
```

This commit still needs:
- The state JSON lookup and path logic from  opencontainers/runtime-spec#391.
- A way to convert the `interface{}` to an `rspec.Spec` (the current FIXME raises: `FATA[0000] translated template has an invalid schema`).

I'm comfortable working through the former, but would appreciate pointers on the latter from anyone who is more Go-literate than I.

More generally, I'm curious to know what folks think of this approach.  I'd rather make those before grinding through the implementation details ;).

### DIFF
--- a/generate.go
+++ b/generate.go
@@ -13,6 +13,7 @@ import (
 	"github.com/codegangsta/cli"
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/syndtr/gocapability/capability"
+	"github.com/opencontainers/ocitools/translate"
 )
 
 var generateFlags = []cli.Flag{
@@ -55,7 +56,9 @@ var generateFlags = []cli.Flag{
 	cli.StringSliceFlag{Name: "seccomp-arch", Usage: "specifies Additional architectures permitted to be used for system calls"},
 	cli.StringSliceFlag{Name: "seccomp-syscalls", Usage: "specifies Additional architectures permitted to be used for system calls, e.g Name:Action:Arg1_index/Arg1_value/Arg1_valuetwo/Arg1_op, Arg2_index/Arg2_value/Arg2_valuetwo/Arg2_op "},
 	cli.StringSliceFlag{Name: "seccomp-allow", Usage: "specifies syscalls to be added to allowed"},
+	cli.StringFlag{Name: "runtime", Usage: "select the runtime command (used for some translations)"},
 	cli.StringFlag{Name: "template", Usage: "base template to use for creating the configuration"},
+	cli.StringSliceFlag{Name: "translate", Usage: "translate higher level constructs"},
 	cli.StringSliceFlag{Name: "label", Usage: "add annotations to the configuration e.g. key=value"},
 }
 
@@ -93,12 +96,35 @@ var generateCommand = cli.Command{
 			}
 		}
 
-		err := modify(spec, context)
+		translations := context.StringSlice("translate")
+		for _, translation := range translations {
+			translator, ok := translate.Translators[translation]
+			if !ok {
+				logrus.Fatalf("unrecognized translation: %s", translation)
+			}
+			var err error
+			spec, err = translator(spec, context)
+			if err != nil {
+				logrus.Fatal(err)
+			}
+		}
+
+		buf, err := json.Marshal(spec)
+		if err != nil {
+			logrus.Fatal(err)
+		}
+		var strictSpec rspec.Spec
+		err = json.Unmarshal(buf, &strictSpec)
+		if err != nil {
+			logrus.Fatal(err)
+		}
+
+		err = modify(&strictSpec, context)
 		if err != nil {
 			logrus.Fatal(err)
 		}
 		cName := "config.json"
-		data, err := json.MarshalIndent(&spec, "", "\t")
+		data, err := json.MarshalIndent(&strictSpec, "", "\t")
 		if err != nil {
 			logrus.Fatal(err)
 		}
@@ -108,7 +134,7 @@ var generateCommand = cli.Command{
 	},
 }
 
-func loadTemplate(path string) (spec *rspec.Spec, err error) {
+func loadTemplate(path string) (spec interface{}, err error) {
 	cf, err := os.Open(path)
 	if err != nil {
 		if os.IsNotExist(err) {
@@ -709,7 +735,7 @@ func removeNamespace(namespaces *[]rspec.Namespace, namespaceType rspec.Namespac
 
 func sPtr(s string) *string { return &s }
 
-func getDefaultTemplate() *rspec.Spec {
+func getDefaultTemplate() interface{} {
 	spec := rspec.Spec{
 		Version: rspec.Version,
 		Platform: rspec.Platform{
@@ -824,5 +850,5 @@ func getDefaultTemplate() *rspec.Spec {
 		},
 	}
 
-	return &spec
+	return spec
 }
--- a/man/ocitools-generate.1.md
+++ b/man/ocitools-generate.1.md
@@ -147,6 +147,9 @@ inside of the container.
 **--rootfs**=ROOTFSPATH
   Path to the rootfs
 
+**--runtime**=COMMAND
+  Set the runtime command, which is used for state JSON queries in translations like **--translate=fromContainer**.
+
 **--seccomp-arch**=ARCH
   Specifies Additional architectures permitted to be used for system calls.
   By default if you turn on seccomp, only the host architecture will be allowed.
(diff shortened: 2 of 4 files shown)


liangchenye:
I tried 'reflect' to convert an interface to a struct before. But later I choose json.Marshal/Unmarshal, it feels not good but make the implementation easy and right.
```
buf, _ := json.Marshal(interfaceA)
json.Unmarshal(buf, &structA)
```

liangchenye:
The purpose is converting between different specs, like between rkt and oci, right?
From the commit, seems we can only convert from OCI to another (fromContainer in this) format.

How about maintain a list like this:
```
     [{"from": "Linux:namespace:fromContainer",  "to": "Linux:namespace:path"},
      {"from":  "Platform:FakeOS",  "to": "Platform:OS"},
     ]
```
Otherwise it feels so special to see these certain field names: 'linux', 'namespace', 'path'.

wking:
On Tue, May 03, 2016 at 04:18:46AM -0700, 梁辰晔 (Liang Chenye) wrote:
> … But later I choose json.Marshal/Unmarshal…

Going through JSON to setup strictSpec works for me.  Done in 215cee9
→ 0463c2a.

Now that that's working, testing turned up setupNamespaces clobbering
the template's namespace entries, so I'll need to figure that out too
(likely by making the clobbering opt-in).

wking:
On Tue, May 03, 2016 at 04:48:19AM -0700, 梁辰晔 (Liang Chenye) wrote:
> The purpose is converting between different specs, like between rkt
> and oci, right?

It's currently * → OCI.  The fromNamespace translator is (OCI +
fromNamespace extension [1] → OCI).  But you could make a rkt → OCI
translator using this approach as well.

> Otherwise it feels so special to see these certain field names:
> 'linux', 'namespace', 'path'.

I'm not sure where you're going with your from / to list.  Can you
give more detail?  And docs for the fromContainer translation will be
good.  I'll update ocitools-generate(1) tonight.

[1]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2016-04-27.log.html#t2016-04-27T20:32:09

wking:
I've pushed 0463c2a → c565214 shifting from an init in
translate/from_container.go to a populated Translators definition in
map[string]Translate.

wking:
I just pushed c565214 → dae2098, which:

* Adds the --runtime config option.
* Passes *cli.Context to Translate functions (so fromContainer
  can get --runtime, and future translators can get similar
  configuration).
* Factor out namespaceFromContainer to its own function since
  FromContainer was getting long.
* Call the runtime to get the state, and lookup the
  appropriate (usually, see below) namespace path in /proc.
* Document --translate and --runtime in ocitools-generate(1).

Still todo (marked by FIXMEs):

* Lookup /proc in /proc/self/mounts.  I don't like hard-coding
  /proc, but this is a chicken/egg issue, so we may not be able
  to do anything about that.
* Compare /proc with the state JSON namespace [1].  We currently
  just assume that /proc was mounted by a process in the PID
  namespace in which the state JSON's "pid" entry is valid, but
  that's a brittle assumption.
* Adjust setupNamespaces to avoid clobbering the template.
* Shift user-namespace injection into setupNamespaces to avoid
  duplicates.
* Markup for nested definition lists in the man page.  I expect
  we should drop go-md2man in favor of a more established
  man-page format (e.g. AsciiDoc).

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/ujtABQoCmgk
     Subject: Linux: Adding the PID namespace inode to state JSON
     Date: Wed, 30 Dec 2015 21:34:57 -0800
     Message-ID: <20151231053457.GG3680@odin.tremily.us>

liangchenye:
@wking I was thinking of having a framework for translation, most of your codes could be reuse if someone want to implement another extension, like rkt to oci.  I'll try to find out a case and explain it in code.

wking:
A rkt -> OCI translator should fit cleanly into what I have now.  You'd only need to rethink things if you wanted to cover OCI -> rkt, and other translation chains that don't end in OCI.

wking:
On Tue, May 03, 2016 at 03:47:16PM -0700, W. Trevor King wrote:
> * Adjust setupNamespaces to avoid clobbering the template.
> * Shift user-namespace injection into setupNamespaces to avoid
>   duplicates.

I've pushed fixes for the namespace-setup issues in #60.  I'll rebase
this branch on top of that once it lands.

wking:
On Sat, May 07, 2016 at 12:01:18AM -0700, W. Trevor King wrote:
> I've pushed fixes for the namespace-setup issues in #60.  I'll
> rebase this branch on top of that once it lands.

Rebased with dae2098 → f7f25b3.  Still a few TODOs in the commit
message, but I'm fine punting on everything except docs.  Do we have a
way to handle nested definition lists?  Should I create a separate
section for available translations (instead of listing them under
--translate)?

wking:
On Fri, May 13, 2016 at 04:58:56PM -0700, W. Trevor King wrote:
> Still a few TODOs in the commit message, but I'm fine punting on
> everything except docs.

Moved the “available translations” docs to a subsection in f7f25b3 →
10fe0e9 and adjusted the commit message to explain why I'm ok punting
on the two proc issues.  I think this is ready to land now.

wking:
On Tue, May 03, 2016 at 03:47:16PM -0700, W. Trevor King wrote:
> * Lookup /proc in /proc/self/mounts.  I don't like hard-coding
>   /proc, but this is a chicken/egg issue, so we may not be able
>   to do anything about that.

We *may* be able to do this with ‘mount -l -t proc’, but mount(8) says
that the listing mode is for backwards compatibility only and suggests
findmnt.  findmnt(8) talks explicitly about reading from
/proc/self/mountinfo, and stracing both ‘mount -l -t proc’ and
‘findmnt -t proc’ turns up open("/proc/…") only in the findmnt case.
I haven't dug deeper today to figure out how ‘mount -l …’ is getting
its information, but that may be a possible route if we don't want to
hard-code /proc.

wking:
On Sat, May 14, 2016 at 03:15:43PM -0700, W. Trevor King wrote:
> I haven't dug deeper today to figure out how ‘mount -l …’ is getting
> its information…

I'm not sure why strace only shows entries like:

  open(0x7ff987f091ba, …)

for the ‘mount -l …’ call, but it appears to be using a hard-coded
/proc/self/mountinfo as well:

  $ LIBMOUNT_DEBUG=tab mount -l -t proc
  31039: libmount:     INIT: library debug mask: 0x0022
  31039: libmount:     INIT: library version: 2.26.0
  31039: libmount:     INIT:     feature: assert
  31039: libmount:     INIT:     feature: debug
  31039: libmount:      TAB: [0x1097310]: alloc
  31039: libmount:      TAB: [0x1097310]: mtab parse: #1 read mountinfo
  31039: libmount:      TAB: [0x1097310]: /proc/self/mountinfo: start parsing [entries=0, filter=not]
  …

So I'm currently back to having no ideas about detecting alternative
proc mountpoints.  But I'm still ok with this PR landing with a
hard-coded /proc [1].

[1]: https://github.com/opencontainers/ocitools/pull/54#issuecomment-219253908

mrunalp:
I am okay adding this but will wait for rebase after #131

wking:
On Tue, Jul 12, 2016 at 05:12:08PM -0700, Mrunal Patel wrote:
> I am okay adding this but will wait for rebase after #131 

I think I'm going to need some guidance about how much to change after
#131.  One of the changes I made in the first part of
`generateCommand` was using an interface{} instead of an *rspec.Spec
to make space for the out-of-spec fields like fromContainer [1].
However, post-#131 a lot of the methods are attached to a Generator
type with opaque-data [2].  I can:

a. Update Generator to hold an interface{} but that makes manipulating
   the final spec structure awkward.
b. Update Generator to hold both an interface{} and an *rspec.Spec,
   but that isn't very DRY.
c. I can pull New, NewFromFile, etc. off of Generator, have them
   return interface{}s, and then add a NewFromInterface that returns a
   Generator.

None of those sound perfect to me.  I'm leaning towards (c) at the
moment, but wanted to get feedback from maintainers before wading in.

[1]: https://github.com/opencontainers/ocitools/pull/54/commits/10fe0e95981f13c4d0e27327e34e669f544ab06b#diff-7017bb2aedbbeb1e05b6fbffb0f46f41L111
[2]: https://github.com/opencontainers/ocitools/pull/131/files#diff-ed9e41f69ceb4f9d141e2fe7a9da68a4R22

Mashimiao:
I think currently this doesn't make too much sense. I prefer to reject this one.
How do you think about it, @opencontainers/runtime-tools-maintainers ?

liangchenye:
@Mashimiao I just quickly reviewed this thread. I did not see the user scenario here. +1 for close this.

Mashimiao:
closing it

PR #55, opened 2016-05-03
artifacts must be present in a single directory

liangchenye:
According to https://github.com/opencontainers/runtime-spec/blob/master/bundle.md, 
```
While these artifacts MUST all be present in a single directory on the local filesystem
```
Check if the rootfs and config file were in the same directory.

### DIFF
--- a/validate.go
+++ b/validate.go
@@ -53,7 +53,8 @@ var bundleValidateCommand = cli.Command{
 			logrus.Fatal(err)
 		}
 
-		sf, err := os.Open(path.Join(inputPath, "config.json"))
+		configPath := path.Join(inputPath, "config.json")
+		sf, err := os.Open(configPath)
 		if err != nil {
 			logrus.Fatal(err)
 		}
@@ -75,6 +76,11 @@ var bundleValidateCommand = cli.Command{
 		} else if !fi.IsDir() {
 			logrus.Fatalf("Rootfs: %v is not a directory.", spec.Root.Path)
 		}
+
+		if path.Dir(configPath) != path.Dir(rootfsPath) {
+			logrus.Fatalf("Root filesystem and 'config.json' must be present in a single directory.")
+		}
+
 		bundleValidate(spec, rootfsPath)
 		logrus.Infof("Bundle validation succeeded.")
 	},


liangchenye:
The second commit verifies if os and arch combination is valid.

wking:
On Tue, May 03, 2016 at 01:11:14AM -0700, 梁辰晔 (Liang Chenye) wrote:
> While these artifacts MUST all be present in a single directory on
> the local filesystem…

It looks like that requirement landed in
opencontainers/runtime-spec@7232e4b1 (specs: introduce the concept of
a runtime.json, 2015-07-30, opencontainers/runtime-spec#88), although
I don't see a need for it.  Can I check for (and maybe file) pushback
on that point in runtime-spec before df60d95f lands?

de58be07 looks good to me, although of the non-Linux OSes, only
Solaris has platform-specific configuaration in flight
(opencontainers/runtime-spec#411).

initlove:
@wking sure, if 'rootfs' does not have to be in the same directory with config.json, `df60d95` is not necessary. 
What I'm doing is just to make sure the ocitools follows spec strictly, all the `MUST`, `Required` field should be verified.

liangchenye:
@wking sure, if 'rootfs' does not have to be in the same directory with config.json, df60d95 is not necessary. 
What I'm doing is just to make sure the ocitools follows spec strictly, all the `MUST`, `Required` fields should be verified.

liangchenye:
move the second commit to another PR https://github.com/opencontainers/ocitools/pull/57/files

wking:
On Tue, May 03, 2016 at 05:58:19PM -0700, David Liang wrote:
“What I'm doing is just to make sure the ocitools follows spec
strictly, all the `MUST`, `Required` field should be verified.”

+1 to that.  But I think that's going to bring into focus some things
in the spec that don't really make sense, so I don't want to rush
anything in without thinking it over again ;).  In this case, I ended
up suggesting a larger change than I'd initially expected
(opencontainers/runtime-spec#423).

liangchenye:
Close as invalid according to spec changes: https://github.com/opencontainers/runtime-spec/pull/394

wking:
On Sun, May 15, 2016 at 11:51:42PM -0700, 梁辰晔 (Liang Chenye) wrote:
> Close as invalid according to spec changes:
> https://github.com/opencontainers/runtime-spec/pull/394

I don't think that PR lifted the sibling requirement.  And the most
recent spec release still has [1]:

  This MUST include the following artifacts:

  1. config.json…
  2. A directory representing the root filesystem of the container…

  While these artifacts MUST all be present in a single directory on
  the local filesystem…

I've tried to get this restriction lifted
(opencontainers/runtime-spec#423 and opencontainers/runtime-spec#469),
but opencontainers/runtime-spec#469 was the smallest pivot I can think
of to lift the requirement, and it was rejected [2].  So unless by
some magic the runtime-spec maintainers reverse their current
position, I think we need to re-open and merge this PR.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc3/bundle.md
[2]: https://github.com/opencontainers/runtime-spec/pull/469#issuecomment-273621129

Mashimiao:
> I think we need to re-open and merge this PR.

+1

PR #58, opened 2016-05-07
generate: Set ociVersion if it wasn't set in the template

wking:
The default template has this information already, but user-supplied
templates may not.  Before this commit:

    $ head -n2 config.json
    {
      "ociVersion": "",

And after this commit, the last line is:

      "ociVersion": "0.6.0-dev",

Regardless of this commit, template authors can set whatever version
they like without getting clobbered:

    $ ocitools generate --template <(echo '{"ociVersion": "xyz"}')
    $ head -n2 config.json
    {
      "ociVersion": "xyz",


mrunalp:
LGTM

PR #59, opened 2016-05-07
generate: Set process.args to an empty array if it was nil

wking:
Before this commit:

    $ grep args config.json
        "args": null,

With this commit:

    $ grep args config.json
        "args": [],

That's still not a particularly useful value for the process
arguments, but it complies with the specs [“array of strings”
typing][1] (opencontainers/runtime-spec#416) and is closer to a
the useful entry (a non-empty array).

[1]: https://github.com/opencontainers/runtime-spec/blob/dbce512cec0cd0774b236980ec18a4f6a91bc77e/config.md#process-configuration


mrunalp:
LGTM

PR #9, opened 2016-01-21
Add initial version of runtimetest command:

zenlinTechnofreak:
Add initial version of runtimetest command:
add unit and config for managing cases;
replace runtime_test.sh to golang realization;

Signed-off-by: zen Lin(Lin Zhinan) linzhinan@huawei.com

PR #61, opened 2016-05-09
[WIP] update initruntimetest by linzhinan

liangchenye:
This pr is still under development.
Since we have 'state' and 'lifecircle' defined now (comparing to zhinan's previous PR https://github.com/opencontainers/ocitools/pull/9)
This PR not only test the main configs (verified by cmd/runtimetest), but also test the state and life circle.


liangchenye:
any suggestions on state test and lifecirle test? @zenlinTechnofreak

unit.go
```diff
@@ -0,0 +1,171 @@
...
+import (
+	"errors"
+	"io"
+	"os"
+	"os/exec"
+	"path"
+	"strings"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/opencontainers/ocitools/config"
```
wking: Travis [doesn't like this import][1], and you don't seem to use `config.…` in `unit.go` anyway, so you can probably drop this line.

[1]: https://travis-ci.org/opencontainers/ocitools/builds/128853061#L193

wking: Ah, you dropped the config package in cb986bef and just forgot this line.

wking:
There was some consensus in #47 about making runtime testing a
first-class command ([1] and later).  That would be good, but I'm not
sure how to get there.  One problem with this Go-based approach is
that it's a lot more work to test command-line invocation
(e.g. testState [2]) than it would be in a shell-based test framework
(e.g. [3]).  So as much as I'd like an easily-installable command for
testing, my current preference is still to exercise the runtime's
command-line API through a command-line test framework [4].  RunC's
test suite is likely a good starting point for this, we'd just have to
trim off the out-of-spec components.

[1]: https://github.com/opencontainers/ocitools/pull/47#issuecomment-215328958
[2]: https://github.com/opencontainers/ocitools/commit/cb986bef52d12b6c24d5ce8776d77d8d3e75a32e#diff-b82b80fc8c029043958c2c2747565054R57
[3]: https://github.com/opencontainers/runc/blob/d49ece5a83da3dcb820121d6850e2b61bd0a5fbe/tests/integration/state.bats
[4]: https://github.com/opencontainers/ocitools/pull/15#issuecomment-193902158

unit.go
liangchenye: Thanks for correction, I'll change that add post a working version.

liangchenye:
The purpose of using golang in runtime test (or runtime certification) is we need to test more cases.
For example, the state test and lifecircle test are different with cmd/runtime test (I called it main config test in these commits). So the command-line script will become bigger and more complex if we add these. TestUnit is a wrap of a container, once there are more test cases, it is easier to apply and maintain them.

wking:
On Tue, May 10, 2016 at 07:16:54AM -0700, 梁辰晔 (Liang Chenye) wrote:
> For example, the state test and lifecircle test are different with
> cmd/runtime test (I called it main config test in these commits). So
> the command-line script will become bigger and more complex if we
> add these.

I'm not recommending a single command-line script, I'm recommending
using a script-based framework (like runC already does [1]).  The bats
equivalent to TestUnit is [2,3], although the latter is killing the
container instead of just waiting for it to exit.  In ccon I'm using
sharness, and tests look like [4].  I expect that approach to scale
pretty well, and ‘test_expect_code 1 …’ (e.g. [5]) is a very compact
way to check non-zero exit codes.

[1]: https://github.com/opencontainers/runc/tree/d49ece5a83da3dcb820121d6850e2b61bd0a5fbe/tests/integration
[2]: https://github.com/opencontainers/runc/blob/d49ece5a83da3dcb820121d6850e2b61bd0a5fbe/tests/integration/state.bats#L18-L20
[3]: https://github.com/opencontainers/runc/blob/d49ece5a83da3dcb820121d6850e2b61bd0a5fbe/tests/integration/state.bats#L41-L43
[4]: https://github.com/wking/ccon/blob/c33398f758b4b878891f2270b73199e4c643c1c6/test/t2001-user-namespace.t#L22-L37
[5]: https://github.com/wking/ccon/blob/c33398f758b4b878891f2270b73199e4c643c1c6/test/t2001-user-namespace.t#L73

wking:
It seems like the resistance to using bats, sharness, Python (#98),
etc. is the desire to ship a single binary to cover all ocitools
functionality [1].  I've been looking around for a Go analog to
Python's unittest that can be built into the installed binary, but so
far haven't turned up anything useful.  It's possible that we could
use Go's testing library and create our own runner with testing.Main
[2] (since ‘go test’ expects *_test.go, and those are excluded from
the regular package [3]).  But while it seems like it should be
possible, I haven't been able to figure out how to do it.

I'm not excited about writing and maintaining our own Go test harness,
so I'd appreciate if folks who prefer a single compiled binary to a
scriptier Bash / POSIX shell / Python solution could lay out their
reasoning for wanting the binary.  Do we really expect folks to want
to run ocitools on systems without Python?

[1]: https://github.com/opencontainers/ocitools/issues/98#issuecomment-223864567
[2]: https://golang.org/pkg/testing/#Main
[3]: https://golang.org/pkg/testing/#pkg-overview

Mashimiao:
any updates?

liangchenye:
Close and split it into small issues/PRs.

PR #65, opened 2016-05-11
SELinux: Fixup man page

rhatdan:
Man page does not show correctly.

Signed-off-by: Dan Walsh <dwalsh@redhat.com>


wking:
4fcd94a1 looks good to me, although I'm also fine keeping the
*italics* around placeholders like *MOUNTLABEL*.

wking:
While we're fixing these, --bind also has an ‘=[=’ synopsis.

rhatdan:
Expanded the cleanup to add more consistency.  Most of the time we were not high lighting the option values, so I removed all highlights.

mrunalp:
LGTM

PR #71, opened 2016-05-18
man/ocitools-generate: Reword --poststop (after container-process exit)

### LINKED ISSUE #41: poststop entry in ocitools-generate(1) says "before the container process is destroyed"
See [here][1], with language from #33.  The language looks like it's drifting towards the in-flight opencontainers/runtime-spec#395.  But however that PR goes down, the container *process* is gone by the time the post-stop hooks run, because that process existing is [the stop event][2].  I'd suggest echoing the spec:

> The post-stop hooks are called after the container process is stopped.

although I'd be a bit happier if the spec used “exits” instead of “is stopped” (to avoid confusion with `SIGSTOP`).

[1]: https://github.com/opencontainers/ocitools/blame/b31bc931d1d50fb7d606e10d56065848e8ba4f26/man/ocitools-generate.1.md#L138-L140
[2]: https://github.com/opencontainers/runtime-spec/blob/v0.5.0/config.md#poststop

wking:
“is stopped” echoes [the spec][1].  I'd be a bit happier if the spec
used “exits” instead of “is stopped” (to avoid confusion with
SIGSTOP), but this man page is not the place to be generating new
language.

Fixes #41.

[1]: https://github.com/opencontainers/runtime-spec/blob/v0.5.0/config.md#poststop

### DIFF
--- a/man/ocitools-generate.1.md
+++ b/man/ocitools-generate.1.md
@@ -120,8 +120,8 @@ inside of the container.
   main process has been created.
 
 **--poststop**=CMD
-  Path to command to run in poststop hooks. This command will be run after the
-  container completes but before the container process is destroyed
+  Path to command to run in poststop hooks. The command will be run
+  after the container process is stopped.
 
 **--prestart**=CMD
   Path to command to run in prestart hooks. This command will be run before


mrunalp:
LGTM

Issue #66, opened 2016-05-13
runtimespec VS ocitools


liangchenye:
## [Filesystem bundle](https://github.com/opencontainers/runtime-spec/blob/master/bundle.md)
- [x] [code](https://github.com/opencontainers/runtime-tools/blob/ab1a21ea926b1ae7a2277b267955807ae9f9a2c9/validate/validate.go#L68) This REQUIRED file MUST reside in the root of the bundle directory.
- [x] [code](https://github.com/opencontainers/ocitools/blob/a637b56baa7229f1da25d8f1bf52f1da185135cd/validate.go#L59) config.json : This REQUIRED file, which MUST be named config.json.
- [x] [code](https://github.com/opencontainers/runtime-tools/blob/ab1a21ea926b1ae7a2277b267955807ae9f9a2c9/validate/validate.go#L130) While these artifacts MUST all be present in a single directory on the local filesystem.
- [x] [code](https://github.com/opencontainers/ocitools/blob/a637b56baa7229f1da25d8f1bf52f1da185135cd/validate.go#L72) This directory MUST be referenced from within the config.json file.
## [Runtime and Lifecycle](https://github.com/opencontainers/runtime-spec/blob/master/runtime.md)
- [x] :first_quarter_moon: [code](https://github.com/opencontainers/ocitools/blob/a637b56baa7229f1da25d8f1bf52f1da185135cd/cmd/runtimetest/main.go) The container's runtime environment MUST be created according to the configuration in config.json.
- [x] [code](https://github.com/opencontainers/ocitools/blob/a637b56baa7229f1da25d8f1bf52f1da185135cd/cmd/runtimetest/main.go#L22) Any updates to config.json after container is running MUST not affect the container.
- [x] [WIP](https://github.com/opencontainers/ocitools/pull/61) The user specified process MUST be executed in the container.
- [ ] The container MUST be destroyed by undoing the steps performed during create phase (step 2)
- [ ] [WIP](https://github.com/opencontainers/ocitools/pull/61) Query State: state <container-id>
- [ ] [WIP](https://github.com/opencontainers/ocitools/pull/61) This operation MUST generate an error if it is not provided the ID of a container.
- [ ] [WIP](https://github.com/opencontainers/ocitools/pull/61) This operation MUST return the state of a container as specified in the State section. In particular, the state MUST be serialized as JSON.
### Start :lock:  [discuss](https://github.com/opencontainers/ocitools/issues/66#issuecomment-219103747)
[...]

PR #72, opened 2016-05-19
validation: add hostname validation

Mashimiao:

### DIFF
--- a/validate.go
+++ b/validate.go
@@ -86,7 +86,7 @@ func bundleValidate(spec rspec.Spec, rootfs string) {
 	checkSemVer(spec.Version)
 	checkPlatform(spec.Platform)
 	checkProcess(spec.Process, rootfs)
-	checkLinux(spec.Linux, rootfs)
+	checkLinux(spec.Linux, spec.Hostname, rootfs)
 }
 
 func checkSemVer(version string) {
@@ -154,7 +154,9 @@ func checkProcess(process rspec.Process, rootfs string) {
 }
 
 //Linux only
-func checkLinux(spec rspec.Linux, rootfs string) {
+func checkLinux(spec rspec.Linux, hostname string, rootfs string) {
+	utsExists := false
+
 	if len(spec.UIDMappings) > 5 {
 		logrus.Fatalf("Only 5 UID mappings are allowed (linux kernel restriction).")
 	}
@@ -165,7 +167,13 @@ func checkLinux(spec rspec.Linux, rootfs string) {
 	for index := 0; index < len(spec.Namespaces); index++ {
 		if !namespaceValid(spec.Namespaces[index]) {
 			logrus.Fatalf("namespace %v is invalid.", spec.Namespaces[index])
+		} else if spec.Namespaces[index].Type == rspec.UTSNamespace {
+			utsExists = true
 		}
+	}
+
+	if !utsExists && hostname != "" {
+		logrus.Fatalf("Hostname requires a new UTS namespace to be specified as well")
 	}
 
 	for index := 0; index < len(spec.Devices); index++ {


validate.go
```diff
@@ -163,9 +165,15 @@ func checkLinux(spec rspec.Linux, rootfs string) {
 	for index := 0; index < len(spec.Namespaces); index++ {
 		if !namespaceValid(spec.Namespaces[index]) {
 			logrus.Fatalf("Namespace %s is invalid.", spec.Namespaces[index])
+		} else if spec.Namespces[index].Type == spec.UTSNamespace {
+			utsExist = true
```
liangchenye: s/spec.UTS/rspec.UTS

validate.go
```diff
@@ -163,6 +163,8 @@ func checkLinux(spec rspec.Linux, rootfs string) {
 	for index := 0; index < len(spec.Namespaces); index++ {
 		if !namespaceValid(spec.Namespaces[index]) {
 			logrus.Fatalf("Namespace %s is invalid.", spec.Namespaces[index])
+		} else if spec.Namespaces[index].Type == rspec.UTSNamespace && rspec.Spec.Hostname != "" {
```
wking: This probably needs the whole rspec.Spec, not its current rspec.Linux.  Then you can check `spec.Hostname` and `spec.Linux.Namespaces`.  With cross-platform options like ‘hostname’ depending on platform-specific options like UTS namespaces, this kind of cross-cutting is unavoidable.

liangchenye:
LGTM

liangchenye:
Hi @Mashimiao , the [spec](https://github.com/opencontainers/runtime-spec/blob/master/config.md#hostname) says:
```
 On Linux, you can only set this if your bundle creates a new UTS namespace.
```
So host name could be empty even a config file has a UTS namespace.
We should get an error when  hostname is empty and no UTS namespace is created

validate.go
```diff
@@ -163,6 +163,8 @@ func checkLinux(spec rspec.Linux, rootfs string) {
 	for index := 0; index < len(spec.Namespaces); index++ {
 		if !namespaceValid(spec.Namespaces[index]) {
 			logrus.Fatalf("Namespace %s is invalid.", spec.Namespaces[index])
+		} else if spec.Namespaces[index].Type == rspec.UTSNamespace && hostname != "" {
```
liangchenye:
```
 On Linux, you can only set this if your bundle creates a new UTS namespace.
```
So host name could be empty even a config file has a UTS namespace.
We should get an error when hostname is empty and no UTS namespace is created

liangchenye:
add a WIP link in https://github.com/opencontainers/ocitools/issues/66

validate.go
wking: On Thu, May 19, 2016 at 08:30:00PM -0700, 梁辰晔 (Liang Chenye) wrote:
> So host name could be empty even a config file has a UTS namespace.

Correct.  Maybe the UTS namespace is for a NIS domain name.

> We should get an error when hostname is empty and no UTS namespace
> is created

I think you mean, when the hostname is *not* empty and no UTS
namespace is created.  All other combinations (UTS and no hostname,
neither UTS not hostname, both UTS and hostname) are valid.

liangchenye: Yes, `when hostname is not empty`.
@Mashimiao 's first commit did the right thing and I mislead him to squash code lines.

Mashimiao:
@liangchenye  @wking   fixed.

wking:
b62e1cc5 looks good to me.

liangchenye:
LGTM

validate.go
```diff
@@ -152,7 +152,9 @@ func checkProcess(process rspec.Process, rootfs string) {
 }
 
 //Linux only
-func checkLinux(spec rspec.Linux, rootfs string) {
+func checkLinux(spec rspec.Linux, hostname string, rootfs string) {
+	utsExist := false
```
mrunalp: nit: `utsExist` --> `utsExists`

Mashimiao: fixed

validate.go
```diff
@@ -177,9 +179,15 @@ func checkLinux(spec rspec.Linux, rootfs string) {
 	for index := 0; index < len(spec.Namespaces); index++ {
 		if !namespaceValid(spec.Namespaces[index]) {
 			logrus.Fatalf("namespace %v is invalid.", spec.Namespaces[index])
+		} else if spec.Namespaces[index].Type == rspec.UTSNamespace {
+			utsExists = true
 		}
 	}
 
+	if !utsExists && hostname != "" {
+		logrus.Fatalf("New UTS namespace should to be created.")
```
mrunalp: "hostname requires a new UTS namespace to be specified as well"

Mashimiao: @mrunalp  fixed.

mrunalp:
LGTM

PR #20, opened 2016-02-05
add bundle validate as sub-command

liangchenye:
Signed-off-by: liangchenye <liangchenye@huawei.com>

Clean the worthless commits in (https://github.com/opencontainers/ocitools/pull/4/) and re-submit it.

PR #73, opened 2016-05-20
validate: Drop mount checks

wking:
These landed as CheckMounts in 647e3552 (bundle validate update to
0.3.0, 2016-02-23, #20), but both checks are too strict.

The first (destination exists in the rootfs) errors on valid cases
like:

    "mounts": [
      {
        "source": "users",
        "destination": "/home",
        "type": "bind"
      },
      {
        "source": "none",
        "destination": "/home/wking",
        "type": "tmpfs"
      }
    ]

Where the source `users` directory already contained a `wking`
subdirectory.  So by the time the tmpfs was setup, the destination
directory would exist, but at validation time (without having run the
bind mount) the tmpfs destination directory would not exist.

The second (destination is a directory) errors on valid cases like:

    "mounts": [
      {
        "source": "/etc/resolv.conf",
        "destination": "/etc/resolv.conf",
        "type": "bind"
      }
    ]

because binding files to files works.  In a shell:

    # touch test
    # mount --bind /etc/resolv.conf test
    # umount test

However binding directories to files does not work:

    # mount --bind /etc test
    mount: mount point /tmp/test is not a directory

Figuring out which mount configurations are valid and which aren't may
be possible, but I'm pretty sure it's more trouble than we want to get
into.  There may be room for other mount tests (e.g. comparing `type`
against `/proc/filesystems` as a host-specific test), but I'm leaving
those to subsequent pull requests.

### DIFF
--- a/validate.go
+++ b/validate.go
@@ -86,7 +86,6 @@ func bundleValidate(spec rspec.Spec, rootfs string) {
 	checkSemVer(spec.Version)
 	checkPlatform(spec.Platform)
 	checkProcess(spec.Process, rootfs)
-	checkMounts(spec.Mounts, rootfs)
 	checkLinux(spec.Linux, rootfs)
 }
 
@@ -94,17 +93,6 @@ func checkSemVer(version string) {
 	re, _ := regexp.Compile("^(\\d+)?\\.(\\d+)?\\.(\\d+)?$")
 	if ok := re.Match([]byte(version)); !ok {
 		logrus.Fatalf("%q is not a valid version format, please read 'SemVer v2.0.0'", version)
-	}
-}
-
-func checkMounts(mounts []rspec.Mount, rootfs string) {
-	for _, mount := range mounts {
-		destPath := path.Join(rootfs, mount.Destination)
-		if fi, err := os.Stat(destPath); err != nil {
-			logrus.Fatalf("Cannot find the mount destination %q", destPath)
-		} else if !fi.IsDir() {
-			logrus.Fatalf("Mount destination %q is not a directory.", destPath)
-		}
 	}
 }
 


wking:
And somewhat related to “more trouble than we want to get into” is opencontainers/runtime-spec#418, which allows the runtime to not validate every possible thing before asking the kernel to do it ;).  And in `ocitools validate` we're tasked with validating the bundle config against the runtime spec, but we don't have to get down into validating the bundle config against kernel constraints unless we want to.

mrunalp:
I am okay removing this check but would like to hear @liangchenye opinion as well.

liangchenye:
Looks good to  me too. @mrunalp @wking

mrunalp:
Needs rebase.

wking:
On Mon, May 23, 2016 at 09:26:19AM -0700, Mrunal Patel wrote:
> Needs rebase.

Rebased with 2eaec20 → 3b9d616.

mrunalp:
github is still complaining about conflicts.

wking:
On Mon, May 23, 2016 at 09:31:19AM -0700, Mrunal Patel wrote:
> github is still complaining about conflicts. 

Odd, I'd pushed the commit [1].  I just bumped the commit date to get
a new hash and pushed db4632b, and GitHub doesn't seem to have noticed
that either.  Maybe just merge this one from the command line?  Or I
can open a new PR?

[1]: https://github.com/wking/ocitools-v2/commit/3b9d616e8e1b58c4e5ceb29acf1f3623e52055d5

mrunalp:
@wking Can you open a new PR?

wking:
On Mon, May 23, 2016 at 09:57:12AM -0700, Mrunal Patel wrote:
> @wking Can you open a new PR?

Filed in #74.

mrunalp:
LGTM. Looks like github was probably slow in updating merge status.

PR #70, opened 2016-05-18
verify: config is encoded in utf8; cwd is absolute path; env is valid

liangchenye:
fix https://github.com/opencontainers/ocitools/issues/67
Signed-off-by: liang chenye <liangchenye@huawei.com>

PR #76, opened 2016-05-25
runtimetest: add readonly path validation

Mashimiao:

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -233,6 +233,17 @@ func validateMaskedPaths(spec *rspec.Spec) error {
 	return nil
 }
 
+func validateROPaths(spec *rspec.Spec) error {
+	fmt.Println("validating readonlyPaths")
+	for _, v := range spec.Linux.ReadonlyPaths {
+		err := testWriteAccess(v)
+		if err == nil {
+			return fmt.Errorf("%v should be readonly", v)
+		}
+	}
+	return nil
+}
+
 func main() {
 	spec, err := loadSpecConfig()
 	if err != nil {
@@ -247,6 +258,7 @@ func main() {
 		validateRlimits,
 		validateSysctls,
 		validateMaskedPaths,
+		validateROPaths,
 	}
 
 	for _, v := range validations {


wking:
On Tue, May 24, 2016 at 08:52:20PM -0700, Ma Shimiao wrote:
>   * runtimetest: add readonly path validation

Travis is missing ‘spec’ in main.go around line 194 [1].

[1]: https://travis-ci.org/opencontainers/ocitools/builds/132739528#L186

cmd/runtimetest/main.go
```diff
@@ -191,6 +191,20 @@ func validateSysctls(spec *rspec.Spec) error {
 	return nil
 }
 
+func validateROPaths(spec, *rspec.Spec) error {
```
wking: Drop the comma?

cmd/runtimetest/main.go
```diff
@@ -191,6 +191,20 @@ func validateSysctls(spec *rspec.Spec) error {
...
+	fmt.Println("validating Readonly Paths")
+	for _, v := range spec.Linux.ReadonlyPaths {
+		fi, err := os.Stat(v)
```
wking: This looks like host-mount-namespace access, but the read-only paths may only exist in the container mount namespace.  Is there code somewhere to enter the container mount namespace for these checks?

Mashimiao: I got from spec `readonlyPaths will set the provided paths as readonly inside the container.`

wking: On Tue, May 24, 2016 at 09:35:14PM -0700, Ma Shimiao wrote:
> > @@ -191,6 +191,20 @@ func validateSysctls(spec *rspec.Spec) error {
> >  	return nil
> >  }
> >  
> > +func validateROPaths(spec, *rspec.Spec) error {
> > +	fmt.Println("validating Readonly Paths")
> > +	for _, v := range spec.Linux.ReadonlyPaths {
> > +		fi, err := os.Stat(v)
>
> I got from spec `readonlyPaths will set the provided paths as
> readonly inside the container.`

Right, but unless you're joining the container's mount namespace, you
may not be looking inside the container.  Consider the case where the
container creates a new mount namespace, mounts a new /proc, and lists
/proc/cpuinfo (or whatever) in readonlyPaths (which the runtime
decides to implement by bind-mounting with a ro option).  That's going
to still be read/write in the *runtime* mount namespace [1]; it's only
read-only in the *container* mount namespace [2].

[1]: https://github.com/opencontainers/runtime-spec/blob/v0.5.0/glossary.md#runtime-namespace
[2]: https://github.com/opencontainers/runtime-spec/blob/v0.5.0/glossary.md#container-namespace

Mashimiao:
@wking  Yes, but `runtimetest` as configured process in config.json, it should be executed in container namespace. So, I think there is no problem.

cmd/runtimetest/main.go
Mashimiao: @wking  Yes, but `runtimetest` as configured process in config.json, it should be executed in container namespace. So, I think there is no problem.

wking:
On Tue, May 24, 2016 at 11:11:04PM -0700, Ma Shimiao wrote:
> Yes, but `runtimetest` as configured process in config.json, it
> should be executed in container namespace.

Ah, that's the piece I was forgetting.  This looks good to me then.

cmd/runtimetest/main.go
```diff
@@ -191,6 +191,20 @@ func validateSysctls(spec *rspec.Spec) error {
 	return nil
 }
 
+func validateROPaths(spec *rspec.Spec) error {
+	fmt.Println("validating Readonly Paths")
```
liangchenye: using 'validating readonlyPaths' as Mrunal suggested [match what is expected to be the property name]( https://github.com/opencontainers/ocitools/pull/70).

Mashimiao:
@liangchenye  fixed.

liangchenye:
LGTM

cmd/runtimetest/main.go
```diff
@@ -191,6 +191,20 @@ func validateSysctls(spec *rspec.Spec) error {
 	return nil
 }
 
+func validateROPaths(spec *rspec.Spec) error {
+	fmt.Println("validating readonlyPaths")
+	for _, v := range spec.Linux.ReadonlyPaths {
+		fi, err := os.Stat(v)
```
wking: This stat-based approach should be changed to an [attempted-write approach][1].

[1]: https://github.com/opencontainers/ocitools/pull/78#issuecomment-221837970

wking: And it's probably worth making the “is this path readonly?” a separate function so it can be shared with the check in #81.

Mashimiao: @wking  The spec says "`readonlyPaths will set the provided paths as readonly inside the container.`"
It seems not to remount readonlyPaths and just change their permisson. If this is true, I think current implementation is OK.

cyphar:
This has the same problem as #78 -- mounting something readonly doesn't change the mode. We should check that it is actually readonly by attempting a write and making sure it fails with `EROFS`.

cmd/runtimetest/main.go
wking: On Thu, May 26, 2016 at 07:30:15PM -0700, Ma Shimiao wrote:
> It seems not to remount readonlyPaths and just change their
> permisson.

I think the spec wording leaves the implemenation ambiguous.  And
attempting a read clearly covers all possible implementations, while
checking the mode only covers a single implementation.

liangchenye:
3c250a9 LGTM

cyphar:
IANAM, but LGTM 3c250a9.

mrunalp:
Looks good but needs rebase.

Mashimiao:
@mrunalp   rebased.

mrunalp:
LGTM

PR #78, opened 2016-05-26
runtimetest: add root filesystem validation

Mashimiao:

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -191,6 +191,30 @@ func validateSysctls(spec *rspec.Spec) error {
 	return nil
 }
 
+func testWriteAccess(path string) error {
+	tmpfile, err := ioutil.TempFile(path, "Test")
+	if err != nil {
+		return err
+	}
+
+	tmpfile.Close()
+	os.RemoveAll(filepath.Join(path, tmpfile.Name()))
+
+	return nil
+}
+
+func validateRootFS(spec *rspec.Spec) error {
+	fmt.Println("validating root")
+	if spec.Root.Readonly {
+		err := testWriteAccess("/")
+		if err == nil {
+			return fmt.Errorf("Rootfs should be readonly")
+		}
+	}
+
+	return nil
+}
+
 func main() {
 	spec, err := loadSpecConfig()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


liangchenye:
Hi @Mashimiao I just submitted a PR https://github.com/opencontainers/ocitools/pull/81 which does the same thing, but by using mount information.

I think it might be wrong to verify the `readonly property` of the root file system by using os.stat.
if you `mount yourPartition -o /mnt/readonly-point` and use os.stat, it might still have write permission. But when you touch `a` file in /tmp/readonly-point, it will tell:
```
touch: cannot touch 'a': Read-only file system
```

wking:
On Thu, May 26, 2016 at 03:43:54AM -0700, 梁辰晔 (Liang Chenye) wrote:
> if you `mount yourPartition -o /mnt/readonly-point` and use os.stat,
> it might still have write permission. But when you touch `a` file in
> /tmp/readonly-point…

Ah, good point.  We should definitely be testing read-only by
attempting writes, because the per-file POSIX-mode permissions are
distinct from the per-mount filesystem options.

Mashimiao:
@liangchenye  @wking   re-implemented with pre-write

wking:
On Thu, May 26, 2016 at 09:47:37PM -0700, Ma Shimiao wrote:
> @liangchenye  @wking   re-implemented with pre-write

I'll leave it to you and @liangchenye to decide whether this rerolled
version takes precedence over #81 ;).

liangchenye:
@wking  Ah! I'll rollback my second commit in #81.

cmd/runtimetest/main.go
```diff
@@ -191,13 +191,38 @@ func validateSysctls(spec *rspec.Spec) error {
...
 }
 
+func testWriteAccess(path string) error {
+	tmpfile, err := ioutil.TempFile(path, "Test")
+	if err != nil {
+		return err
+	}
+
+	tmpfile.Close()
+	os.RemoveAll(filepath.Join(path,tmpfile.Name()))
```
mrunalp: Probably needs gofmt run.

Mashimiao: @mrunalp    Ah, sorry. fixed.

mrunalp:
LGTM

PR #81, opened 2016-05-26
Mount test 

liangchenye:
validate the runtime mount information by using docker/pkg/mount
and validate the rootfs readonly setting

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -12,6 +12,7 @@ import (
 	"syscall"
 
 	"github.com/Sirupsen/logrus"
+	"github.com/opencontainers/ocitools/cmd/runtimetest/mount"
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/syndtr/gocapability/capability"
 )
@@ -237,6 +238,56 @@ func validateROPaths(spec *rspec.Spec) error {
 			return fmt.Errorf("%v should be readonly", v)
 		}
 	}
+
+	return nil
+}
+
+func mountMatch(specMount rspec.Mount, sysMount rspec.Mount) error {
+	if specMount.Destination != sysMount.Destination {
+		return fmt.Errorf("mount destination expected: %v, actual: %v", specMount.Destination, sysMount.Destination)
+	}
+
+	if specMount.Type != sysMount.Type {
+		return fmt.Errorf("mount %v type expected: %v, actual: %v", specMount.Destination, specMount.Type, sysMount.Type)
+	}
+
+	if specMount.Source != sysMount.Source {
+		return fmt.Errorf("mount %v source expected: %v, actual: %v", specMount.Destination, specMount.Source, sysMount.Source)
+	}
+
+	return nil
+}
+
+func validateMountsExist(spec *rspec.Spec) error {
+	fmt.Println("validating mounts exist")
+	mountInfos, err := mount.GetMounts()
+	if err != nil {
+		return err
+	}
+
+	mountsMap := make(map[string][]rspec.Mount)
+	for _, mountInfo := range mountInfos {
+		m := rspec.Mount{
+			Destination: mountInfo.Mountpoint,
+			Type:        mountInfo.Fstype,
+			Source:      mountInfo.Source,
+		}
+		mountsMap[mountInfo.Mountpoint] = append(mountsMap[mountInfo.Mountpoint], m)
+	}
+
+	for _, specMount := range spec.Mounts {
+		found := false
+		for _, sysMount := range mountsMap[specMount.Destination] {
+			if err := mountMatch(specMount, sysMount); err == nil {
+				found = true
+				break
+			}
+		}
+		if !found {
+			return fmt.Errorf("Expected mount %v does not exist", specMount)
+		}
+	}
+
 	return nil
 }
 
@@ -255,6 +306,7 @@ func main() {
 		validateSysctls,
 		validateMaskedPaths,
 		validateROPaths,
+		validateMountsExist,
 	}
 
 	for _, v := range validations {
--- a/cmd/runtimetest/mount/README.md
+++ b/cmd/runtimetest/mount/README.md
@@ -1,0 +1,6 @@
+## Original code
+`github.com/docker/docker/pkg/mount`
+commit 089166ebe29c1801483965b0e30b41fda53b1ec2
+
+## Changelog
+Keep the implementation of `mountinfo*`, remove `flags*` `mounter*` `shared*` files.
--- a/cmd/runtimetest/mount/LICENSE
+++ b/cmd/runtimetest/mount/LICENSE
@@ -1,0 +1,191 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        https://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
+
+      "Contribution" shall mean any work of authorship, including
+      the original version of the Work and any modifications or additions
+      to that Work or Derivative Works thereof, that is intentionally
+      submitted to Licensor for inclusion in the Work by the copyright owner
+      or by an individual or Legal Entity authorized to submit on behalf of
+      the copyright owner. For the purposes of this definition, "submitted"
+      means any form of electronic, verbal, or written communication sent
+      to the Licensor or its representatives, including but not limited to
+      communication on electronic mailing lists, source code control systems,
+      and issue tracking systems that are managed by, or on behalf of, the
+      Licensor for the purpose of discussing and improving the Work, but
+      excluding communication that is conspicuously marked or otherwise
+      designated in writing by the copyright owner as "Not a Contribution."
+
+      "Contributor" shall mean Licensor and any individual or Legal Entity
+      on behalf of whom a Contribution has been received by Licensor and
+      subsequently incorporated within the Work.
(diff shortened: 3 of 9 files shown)


cmd/runtimetest/main.go
```diff
@@ -191,6 +192,81 @@ func validateSysctls(spec *rspec.Spec) error {
...
+		if opt == str {
+			return true
+		}
+	}
+
+	return false
+}
+
+func validateMounts(spec *rspec.Spec) error {
+	fmt.Println("validating mounts")
```
wking: We should probably be using a logging framework for these so folks can adjust the verbosity.

cmd/runtimetest/main.go
```diff
@@ -191,6 +192,81 @@ func validateSysctls(spec *rspec.Spec) error {
...
+	}
+
+	mountsMap := make(map[string]rspec.Mount)
+	for _, info := range infos {
+		opts := strings.Split(info.Opts, ",")
+
+		if info.Mountpoint == "/" {
+			if spec.Root.Readonly && !containString(opts, "ro") ||
+				!spec.Root.Readonly && !containString(opts, "rw") {
+				return fmt.Errorf("root readonly expected: %v, actual: %v", spec.Root.Readonly, !spec.Root.Readonly)
```
wking: Hmm.  If someone has an entry in their `mounts` array that clobbers `/`, I think that should be considered independent of their `root` config.  So you can set it's read-only-ness however you like, and we don't have to compare it with `root.readonly` here.

cmd/runtimetest/main.go
```diff
@@ -11,6 +11,7 @@ import (
 	"syscall"
 
 	"github.com/Sirupsen/logrus"
+	"github.com/docker/docker/pkg/mount"
```
wking: I'm not familiar with Docker's internal package organization, but I worry about cyclic dependencies when a low level tool like ocitools pulls in code from a high-level spec consumer like docker/docker.  Can generic mount parsing/handling get pulled out into an independent repository somewhere so that it's clear that it's also low-level (if it is, I haven't looked)?

cmd/runtimetest/main.go
```diff
@@ -191,6 +192,81 @@ func validateSysctls(spec *rspec.Spec) error {
...
+
+		if info.Mountpoint == "/" {
+			if spec.Root.Readonly && !containString(opts, "ro") ||
+				!spec.Root.Readonly && !containString(opts, "rw") {
+				return fmt.Errorf("root readonly expected: %v, actual: %v", spec.Root.Readonly, !spec.Root.Readonly)
+			}
+		}
+
+		mountsMap[info.Mountpoint] = rspec.Mount{
+			Destination: info.Mountpoint,
```
wking: This “destination → config” mapping isn't going to work if folks define several mounts over the same location.  I'm not sure why folks would want to do that (although it's close to [this workflow][1]), but it is certainly legal with the current spec wording.

[1]: http://man7.org/linux/man-pages/man8/switch_root.8.html#NOTES

cmd/runtimetest/main.go
```diff
@@ -191,6 +192,81 @@ func validateSysctls(spec *rspec.Spec) error {
...
+
+	}
+
+	for _, specMount := range spec.Mounts {
+		if sysMount, ok := mountsMap[specMount.Destination]; ok {
+			if err := mountMatch(specMount, sysMount); err != nil {
+				return err
+			}
+		} else {
+			return fmt.Errorf("Expected mount %v does not exist", specMount.Destination)
```
wking: I think we should log a warning in this case, but don't think it should be an error ([reasoning][1]).

[1]: https://github.com/opencontainers/ocitools/pull/81/files#r64780518

cmd/runtimetest/main.go
mrunalp: I think we should probably pull out mount parsing into a single function. Pulling this dependency is adding a lot of code we don't need.

cmd/runtimetest/main.go
mrunalp: We can use logrus which is already imported.

cmd/runtimetest/main.go
liangchenye: `docker/pkg/mount` supports different platform that is why I pull this dependency.
If we just need to support Linux now, it could be a single function.

liangchenye: > but I worry about cyclic dependencies 
Make sense.

> Can generic mount parsing/handling get pulled out into an independent repository somewhere so that it's clear that it's also low-level (if it is, I haven't looked)?
I'll search that today. Or I'll pull the mount function to ocitools.

cmd/runtimetest/main.go
liangchenye: why not an error? 
I think mounts in config.json must be setup up by a runtime. 
` The runtime MUST mount entries in the listed order `

cmd/runtimetest/main.go
wking: On Thu, May 26, 2016 at 06:24:30PM -0700, 梁辰晔 (Liang Chenye) wrote:
> > +	"github.com/docker/docker/pkg/mount"
> 
> `docker/pkg/mount` supports different platform that is why I pull this dependency.
> If we just need to support Linux now, it could be a single function.

Hmm.  I think we *do* want to support multiple platforms (and stay as
platform-agnostic as possible here).  Anything that's just
conformance-testing for Linux namespaces should probably go in the
runC test-suite ;).  That probably means either pushing to get
docker/pkg/mount split out into it's own repo, or auditing it to see
how much unrelated stuff we're pulling in.

cmd/runtimetest/main.go
wking: On Thu, May 26, 2016 at 07:33:47PM -0700, 梁辰晔 (Liang Chenye) wrote:
> I think mounts in config.json must be setup up by a runtime. 

I agree, I just don't think you will be able to tell if they have by
reading /proc/self/mounts.  For example:

  $ cd /tmp
  $ touch a
  $ echo hello >b
  $ echo world >c
  $ sudo mount --bind b a 
  $ sudo mount -o remount,ro a 
  $ sudo mount --bind c a 
  $ sudo mount -o remount,rw a 
  $ cat a
  world
  $ grep /tmp/a /proc/self/mounts
  none /tmp/a tmpfs ro,relatime 0 0
  none /tmp/a tmpfs rw,relatime 0 0
  $ echo goodbye >a
  $ head a b c
  ==> a <==
  goodbye

  ==> b <==
  hello

  ==> c <==
  goodbye

Your /proc/self/mounts reader stores the final read/write entry (say)
in your mapping.  And then when you iterate through and compare, that
read/write entry doesn't match your earlier read-only entry (but it
does match your later read/write entry).

liangchenye: Thanks for the examples, these fields could be same in different mounts, so using map is not a correct way.
I found problems with 'docker/pkg/mount/mountinfo_linux.go', it parses `/proc/self/mount` but the data of this file is not exactly same with what I input in 'mount' operation.
It becomes:
```
/dev/disk/by-uuid/55a**** /tmp/a ext4 rw ***
...
```
But I could get the same result by using `mount`. 
```
/tmp/b on /tmp/a type none (rw, bind)
...
```

wking: On Wed, Jun 01, 2016 at 03:12:20AM -0700, 梁辰晔 (Liang Chenye) wrote:
> I found problems with 'docker/pkg/mount/mountinfo_linux.go', it
> parses `/proc/self/mount` but the data of this file is not exactly
> same with what I input in 'mount' operation.

I think all of this is sufficiently complicated that it's better to
give up on generically validating all possible mount configurations.
Instead, we should have a handful of testcases that exercise some
mount configurations and test them directly.  For example, a container
setup with:

  {
    …
    "mounts": [
      {
        "destination": "/root",
        "type": "tmpfs",
        "source": "none",
        "options": ["nosuid"]
      },
      …
    ],
    "process": {
      "args": ["cat", "/proc/self/mounts"]
    }
  }

should spit a bunch of stuff to stdout including a line matching:

  none /root tmpfs …nosuid… 0 0

cmd/runtimetest/mount/README.md
```diff
@@ -0,0 +1 @@
+[code from](github.com/docker/docker/pkg/mount)
```
cyphar: Minor thing: If we're going to start pulling code from other repositories we should mention the license (I know it's the same as this project but it's a different project).

liangchenye: Thanks, LICENCE.md and commit ID are added.

liangchenye:
Updated, PTAL @mrunalp @wking

wking:
On Thu, Jun 02, 2016 at 02:13:36AM -0700, 梁辰晔 (Liang Chenye) wrote:
> Updated, PTAL @mrunalp @wking 

3f334dcb → 49ecdec0 is forking and vendoring docker/pkg/mount?  Can't
we get the Docker maintainers to agree on a more neutral place for
shared development?  It seems inefficient to fork.

liangchenye:
>> 3f334dc → 49ecdec is forking and vendoring docker/pkg/mount?  Can't
we get the Docker maintainers to agree on a more neutral place for
shared development?  It seems inefficient to fork.

Yes, it is forking and vendoring docker/pkg/mount and does not take much time actually.
A neutral place is an elegant solution, the only disadvantage is it still brings unnecessary code.

cmd/runtimetest/main.go
```diff
@@ -241,6 +242,67 @@ func validateROPaths(spec *rspec.Spec) error {
...
+			return fmt.Errorf("mount %v option %v does not exist", specMount.Destination, option)
+		}
+	}
+
+	return nil
+}
+
+func validateMountsExistence(spec *rspec.Spec) error {
+	fmt.Println("validating mounts existence")
+	infos, err := mount.GetMounts()
```
mrunalp: nit: infos -> mounts ?

cmd/runtimetest/main.go
```diff
@@ -241,6 +242,67 @@ func validateROPaths(spec *rspec.Spec) error {
...
+		if _, ok := optMap[option]; !ok {
+			return fmt.Errorf("mount %v option %v does not exist", specMount.Destination, option)
+		}
+	}
+
+	return nil
+}
+
+func validateMountsExistence(spec *rspec.Spec) error {
+	fmt.Println("validating mounts existence")
```
mrunalp: s/existence/exist

cmd/runtimetest/main.go
```diff
@@ -241,6 +242,67 @@ func validateROPaths(spec *rspec.Spec) error {
...
+			Type:        info.Fstype,
+			Source:      info.Source,
+			Options:     append(strings.Split(info.Opts, ","), strings.Split(info.VfsOpts, ",")...),
+		}
+		mountsMap[info.Mountpoint] = append(mountsMap[info.Mountpoint], m)
+	}
+
+	for _, specMount := range spec.Mounts {
+		found := false
+		for _, sysMount := range mountsMap[specMount.Destination] {
```
mrunalp: why a loop here?

liangchenye: Different mounts may have the same 'destination', but using 'destination' as a key in mountsMap is still an efficient way. So I use an array (`[]rspec.Mount`) as the value of mountsMap.

cmd/runtimetest/main.go
liangchenye: thanks, updated.

cmd/runtimetest/main.go
liangchenye: The returned struct of mount.GetMounts() is named of 'Info", so I use `infos` previously.
Now it changes to mountInfos.

mrunalp:
The check for mount options doesn't work for `strictatime`. It is passed as MS_STRICTATIME to the /dev mount syscall as it errors out when passed as an option. We should probably skip checking mount options or only check for ones that actually are visible in the mount output.

wking:
On Mon, Jun 06, 2016 at 02:38:39PM -0700, Mrunal Patel wrote:
> It is passed as MS_STRICTATIME to the /dev mount syscall as it
> errors out when passed as an option…

Just to clarify, that distinction is ‘mountflags’ vs. ‘data’ in
mount(2).  Does the “doesn't work in ‘data’” restriction apply to only
a subset of filesystem-independent mount options [1]?

[1]: http://man7.org/linux/man-pages/man8/mount.8.html#FILESYSTEM-INDEPENDENT%C2%A0MOUNT%C2%A0OPTIONS

mrunalp:
@wking Yes, it mountflags vs data. And yes I would think that the filesystem-independent options may or may not be visible in the output.

liangchenye:
Remove 'option' checking for now for three reasons:
1. some flags like 'strictatime' are invisible but not all flags are not invisible. 
    I did not see any rule about this in the mount page or document.
2. options like 'newinstance' is invisible too
    Also I did not see any explanation about it.
3. data like 'a=b' is could be detected, but the output format is a little different (mode=666 vs mode=0666). It will be complicated and may not cover all the differences.

mrunalp:
LGTM. We need to remove cgroups mount from runc config which is non-standard. I'll create a follow-on PR for that.

PR #82, opened 2016-05-27
runtimetest: add masked paths validation

Mashimiao:

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -215,6 +216,23 @@ func validateRootFS(spec *rspec.Spec) error {
 	return nil
 }
 
+func validateMaskedPaths(spec *rspec.Spec) error {
+	fmt.Println("validating maskedPaths")
+	for _, maskedPath := range spec.Linux.MaskedPaths {
+		f, err := os.Open(maskedPath)
+		if err != nil {
+			return err
+		}
+		defer f.Close()
+		b := make([]byte, 1)
+		_, err = f.Read(b)
+		if err != io.EOF {
+			return fmt.Errorf("%v should not be readable", maskedPath)
+		}
+	}
+	return nil
+}
+
 func main() {
 	spec, err := loadSpecConfig()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/runtimetest/main.go
```diff
@@ -191,6 +191,20 @@ func validateSysctls(spec *rspec.Spec) error {
...
 
+func validateMaskedPaths(spec *rspec.Spec) error {
+	fmt.Println("validating maskedPaths")
+	for _, v := range spec.Linux.MaskedPaths {
+		fi, err := os.Stat(v)
+		if err != nil {
+			return err
+		}
+		if fi.Mode()&0444 != 0 {
+			return fmt.Errorf("%v should be readonly", v)
```
wking: The error message shouldn't be about “readonly”.  And this is probably another place where we want to use [“just try it”][1] (in this case, read and write) instead of a mode check.

[1]: https://github.com/opencontainers/ocitools/pull/78#issuecomment-221837970

Mashimiao: @wking  spec says "mask over the provided paths inside the container so that they cannot be read."  it definitely says change paths permission to make them  can't be read.
So,  I think there is no need to "just try it".

wking: On Fri, May 27, 2016 at 12:30:03AM -0700, Ma Shimiao wrote:
> spec says "mask over the provided paths inside the container so that
> they cannot be read."  it definitely says change paths permission to
> make them can't be read.

I don't read “mask over” as clearly “change the permission bits”.  I
haven't been able to turn up runC's implementation with grep, but
@mrunalp's original runtime-spec PR was clearly recommending
bind-mounts [1].

[1]: https://github.com/opencontainers/runtime-spec/pull/186/files

Mashimiao: Ah, maybe we need "just try it". I get from runc code, it implement maskedpaths as bing mount them to /dev/null

mrunalp:
We can add a test that tries to read from a path and gets a EOF.
Something like this
```go
	f, err := os.Open(maskedPath)
        if err != nil {
		return err
	}
	defer f.Close()
	b := make([]byte, 1)
	_, err = f.Read(b)
	if err == io.EOF {
		fmt.Println("Test succeeded")
	} else {
		fmt.Println("Test failed")
	}
```

Mashimiao:
@mrunalp   I combined your suggested code with mine.  If a maskedPath can not be read, an error "permission denied" will occur, I think that is not what we want.
How do you think about it?

Mashimiao:
@mrunalp  It seems I think too much :-).  runc should be run by root.

mrunalp:
LGTM

PR #84, opened 2016-06-01
validation: add env validation for hooks

Mashimiao:

### DIFF
--- a/validate.go
+++ b/validate.go
@@ -125,12 +125,12 @@ func checkPlatform(platform rspec.Platform) {
 }
 
 func checkHooks(hooks rspec.Hooks, hooksCheck bool) {
-	checkEventHookPaths("pre-start", hooks.Prestart, hooksCheck)
-	checkEventHookPaths("post-start", hooks.Poststart, hooksCheck)
-	checkEventHookPaths("post-stop", hooks.Poststop, hooksCheck)
+	checkEventHooks("pre-start", hooks.Prestart, hooksCheck)
+	checkEventHooks("post-start", hooks.Poststart, hooksCheck)
+	checkEventHooks("post-stop", hooks.Poststop, hooksCheck)
 }
 
-func checkEventHookPaths(hookType string, hooks []rspec.Hook, hooksCheck bool) {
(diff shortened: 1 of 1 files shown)


validate.go
```diff
@@ -125,12 +125,12 @@ func checkPlatform(platform rspec.Platform) {
...
-	checkEventHookPaths("pre-start", hooks.Prestart, hooksCheck)
-	checkEventHookPaths("post-start", hooks.Poststart, hooksCheck)
-	checkEventHookPaths("post-stop", hooks.Poststop, hooksCheck)
+	hookValid("pre-start", hooks.Prestart, hooksCheck)
+	hookValid("post-start", hooks.Poststart, hooksCheck)
+	hookValid("post-stop", hooks.Poststop, hooksCheck)
 }
 
-func checkEventHookPaths(hookType string, hooks []rspec.Hook, hooksCheck bool) {
+func hookValid(hookType string, hooks []rspec.Hook, hooksCheck bool) {
```
wking: `checkEventHooks`?

validate.go
```diff
@@ -144,6 +144,11 @@ func checkEventHookPaths(hookType string, hooks []rspec.Hook, hooksCheck bool) {
 			if fi.Mode()&0111 == 0 {
 				logrus.Fatalf("The %s hook %v: is not executable", hookType, hook.Path)
 			}
+			for _, env := range hook.Env {
+				if !envValid(env) {
+					logrus.Fatalf("Env %q for hook %v is in the invalid form.", env, hook.Path)
+				}
+			}
```
wking: This check is host-independent, so it should happen regardless of whether `--hooks` is set (and that portability was the main reason for the flag in #49.  For example, see [here][1]).

wking:
Two minor nits (I left inline comments), but otherwise 62ab8b7f looks
reasonable to me.

validate.go
Mashimiao: checkEventHook?

Mashimiao: I think `checkEventHooks` is OK.

validate.go

mrunalp:
LGTM

PR #86, opened 2016-06-01
validate: small fix for root path error

Mashimiao:

I think absolute path is clearer for users.


wking:
98ef3581 looks good to me, although I'd have gone with “root.path %q
is not a directory”.

mrunalp:
LGTM

PR #88, opened 2016-06-02
runtimetest: add validation of mount order

cyphar:
In order to make sure that runtimes correctly implement the ordering of
[]spec.Mount, we need to check /proc/1/mountinfo (keeping in mind that
Unix allows a user to mount over the top of an existing mountpoint --
thus masking it). We only run this test if there happen to be two
mountpoints in the list where one is a parent of the other.

An extension of this might be to check all of the nested mounts
specified in the spec.

Signed-off-by: Aleksa Sarai asarai@suse.de

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
 	"syscall"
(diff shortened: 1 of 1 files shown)


wking:
This seems like it overlaps with the in-flight #81 (but that may be a
false-positive based on my quick skim); ping @liangchenye.

cyphar:
I just realised this implementation is way too crazy. We should just iterate through `/proc/1/mountinfo` and then make sure that (from the bottom to make sure we don't get false positives from the parent processes' mount namespace) the order is right.

liangchenye:
A little overlap :) @wking   But luckily I'm working on mount existence validation and Aleksa is working on mount order. 
@cyphar I think we can share and use a same mount lib.
In #81 I used docker/pkg/mount at the beginning, now I pull part of that into ocitools/cmd/runtimetest/mount.
The advantage of adding `mount` sub directory is we could support multiple platforms.
Would you mind to rebase your code once #81 is done?

cyphar:
@liangchenye Sure, but I can't remember if `docker/pkg/mount` uses `mountinfo` or `mount`. We need to use `mountinfo` if we want to get the right list of mountpoints (even masked ones).

liangchenye:
@cyphar it uses `mountinfo`, the mountinfo struct also has 'parentID' field.
https://github.com/liangchenye/ocitools/blob/mount/cmd/runtimetest/mount/mountinfo_linux.go

cyphar:
Okay cool, we can use that then. Less work for me. :P

mrunalp:
#81 has been merged so this can be updated. Thanks!

Mashimiao:
ping @cyphar   need rebase and update.

cyphar:
I haven't looked at the code in a few months, but I think I've updated this properly. PTAL.

cmd/runtimetest/main.go
```diff
@@ -459,6 +459,122 @@ func validateMountsExist(spec *rspec.Spec) error {
 	return nil
 }
 
+func isParent(parent, child string) bool {
+	if parent == child {
+		return false
+	}
+
+	cparts := filepath.SplitList(child)
+	for i, part := range filepath.SplitList(parent) {
```
Mashimiao: I think here may be a little problem.  
I'm not sure what SplitList() here works for. Is a mount destination can be "pathA:pathB"?
But path in Windows generally looks like "C:\test". And mount path is not allowed nested in Windows.

cyphar: > I'm not sure what SplitList() here works for. Is a mount destination can be "pathA:pathB"?

`SplitList` splits a path into each component. So `"A/B"` becomes `[]string{"A", "B"}`. This code is checking if one path is lexically the parent of another. While this isn't bulletproof there isn't a nice way to deal with other cases like symlinks (though I could try using `ResolveSymlinks`).


> But path in Windows generally looks like "C:\test". And mount path is not allowed nested in Windows.

I don't use Windows. If Windows doesn't support nested mounts, then this code is not useful for Windows.

Mashimiao:
On 10/07/2016 05:29 PM, Aleksa Sarai wrote:
>     I'm not sure what SplitList() here works for. Is a mount destination can be "pathA:pathB"?
> 
> |SplitList| splits a path into each component. So |"A/B"| becomes |[]string{"A", "B"}|. This code is checking if one path is lexically the parent of another. While this isn't bulletproof there isn't a nice way to deal with other cases like symlinks (though I could try using |ResolveSymlinks|).
> 

I think SplitList is not used for that.
In go document's example, it works like |"/a/b/c:/usr/bin"| becomes |[]string{"/a/b/c", "/usr/bin"}|.
And I tested, it really works like this.
Did I miss something?

>     But path in Windows generally looks like "C:\test". And mount path is not allowed nested in Windows.
> 
> I don't use Windows. If Windows doesn't support nested mounts, then this code is not useful for Windows.
> 

In config.md of runtime-spec, "For the Windows operating system, one mount destination MUST NOT be nested within another mount. (Ex: c:\foo and c:\foo\bar)."
So, I think we should make exception for Windows.

cmd/runtimetest/main.go
Mashimiao: >     I'm not sure what SplitList() here works for. Is a mount destination can be "pathA:pathB"?
> 
> |SplitList| splits a path into each component. So |"A/B"| becomes |[]string{"A", "B"}|. This code is checking if one path is lexically the parent of another. While this isn't bulletproof there isn't a nice way to deal with other cases like symlinks (though I could try using |ResolveSymlinks|).
> 

I think SplitList is not used for that.
In go document's example, it works like |"/a/b/c:/usr/bin"| becomes |[]string{"/a/b/c", "/usr/bin"}|.
And I tested, it really works like this.
Did I miss something?

>     But path in Windows generally looks like "C:\test". And mount path is not allowed nested in Windows.
> 
> I don't use Windows. If Windows doesn't support nested mounts, then this code is not useful for Windows.
> 

In config.md of runtime-spec, "For the Windows operating system, one mount destination MUST NOT be nested within another mount. (Ex: c:\foo and c:\foo\bar)."
So, I think we should make exception for Windows.

cyphar: > I think SplitList is not used for that.

You're quite right. I've fixed it to use `filepath.ToSlash` and then `strings.Split`.

> So, I think we should make exception for Windows.

Sure. I've added it using `runtime.GOOS` -- is that fine?

cmd/runtimetest/main.go
```diff
@@ -459,6 +460,131 @@ func validateMountsExist(spec *rspec.Spec) error {
...
+	}
+
+	// A must *always* be a mountpoint.
+	if ok, err := isMountPoint(A, mountinfos); err != nil {
+		return fmt.Errorf("failed to get whether %q is a mountpoint: %q", A, err)
+	} else if !ok {
+		return fmt.Errorf("expected %q to be a mountpoint", A)
+	}
+
+	// B must be a mountpoint iff A was first.
```
Mashimiao: iff -> if

cmd/runtimetest/main.go
```diff
@@ -459,6 +460,131 @@ func validateMountsExist(spec *rspec.Spec) error {
...
+		return false
+	}
+
+	parent = filepath.ToSlash(parent)
+	child = filepath.ToSlash(child)
+
+	cparts := strings.Split(child, "/")
+	for i, part := range strings.Split(parent, "/") {
+		if cparts[i] != part {
+			return false
```
Mashimiao: maybe we use filepath.Rel() can reduce the loop

cmd/runtimetest/main.go
Mashimiao: I think it's fine

Mashimiao:
It seems currently we can just check one pair of parent and child mount points.
For the whole case, how about:
first, find all pairs of parent, child mount points
second , check whether all pairs are mounted in valid order

@cyphar  how do you think about it?

cmd/runtimetest/main.go
cyphar: iff == if and only if. It's a mathetmical statement.

Mashimiao: Oh, got it. 
But I think using general and easy understood expression may be better.

cyphar: I'll write out the words "if and only if" if it'll make you feel better. The reason I wrote it like that is because the test is testing that *precise* statement, namely that B will only be a mountpoint if A was first (and that if A is first then B will be a mountpoint).

cmd/runtimetest/main.go
cyphar: Sure. I remember that I didn't use it for /some/ reason but I can't remember what it was -- so we'll just switch it.

cyphar:
We could do that. It's just probably going to be a bit tricky to track which orderings matter (if you have mountpoints for `/a`, `/a/b`, `/a/b/c` in arbitrary orders what is an efficient way to check it correctly). I'm also quite busy with other patches at the moment, so if you want to carry it yourself feel free.

Mashimiao:
@cyphar  OK, got it.

Mashimiao:
close in favour of #444 and #456

Issue #91, opened 2016-06-03
Should the `--path` option of `ocitools validate` have a default value?


hmeng-19:
Currently, calling `ocitools validate` without specifying `--path` option would fail:

    [hmeng@localhost test]$ ocitools validate 
    FATA[0000] Bundle path shouldn't be empty  

I am thinking whether we should have a default value for the `--path` option of `ocitools validate`, the default value can be `.` (the current working dir).
This makes calling `ocitools validate` and `ocitools generate` consistent, also saves some typing.

wking:
On Thu, Jun 02, 2016 at 05:17:01PM -0700, hmeng-19 wrote:
> I am thinking whether we should have a default value for the
> `--path` option of `ocitools validate`, the default value can be `.`
> (the current working dir).

Sounds good to me.

rhatdan:
SGTM

mrunalp:
:+1:

liangchenye:
PR checked in. This issue could be close now :)   @mrunalp  @hmeng-19

hmeng-19:
Fixed by #92 .

PR #92, opened 2016-06-03
Add default value for `--path` opt of `validate`

hmeng-19:
This PR is for #91 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,11 +19,12 @@ $ cat config.json
 ## Validating an OCI bundle
 
 [`ocitools validate`][validate.1] validates an OCI bundle.
+The error message will be printed if the OCI bundle failed the validation procedure.
 
 ```sh
 $ ocitools generate
 $ ocitools validate
-FATA[0000] Bundle path shouldn't be empty
+INFO[0000] Bundle validation succeeded.
 ```
 
 ## Testing OCI runtimes
--- a/validate.go
+++ b/validate.go
@@ -19,7 +19,7 @@ import (
 )
 
 var bundleValidateFlags = []cli.Flag{
-	cli.StringFlag{Name: "path", Usage: "path to a bundle"},
+	cli.StringFlag{Name: "path", Value: ".", Usage: "path to a bundle"},
 	cli.BoolFlag{Name: "hooks", Usage: "Check specified hooks exist and are executable on the host."},
 }
 


mrunalp:
@liangchenye WDYT?

liangchenye:
LGTM

liangchenye:
@mrunalp  remove my previous LGTM

@hmeng-19 would you like to change README.md by adding 'Bundle validation succeed.'

hmeng-19:
@liangchenye , do you mean the successful output of `ocitools validate` should be `Bundle validation succeed`, right? 
If that is the case, we may also need to modify `validate.go` reflect the change. right?

liangchenye:
Yes, I just notice the process of bundle validation is invisible to users. The previous README.md does not provide useful information. Since you've changed REAMDE.me in this commit, it will be a good improvement if you add logs like runtimetest(https://github.com/opencontainers/ocitools/blob/master/cmd/runtimetest/main.go#L38).

wking:
On Thu, Jun 02, 2016 at 07:01:09PM -0700, hmeng-19 wrote:
> @liangchenye , do you mean the successful output of `ocitools
> validate` should be `Bundle validation succeed`, right?

I like my command line tools to exit zero and print nothing on success
(if they have nothing to say beyond “everything is fine”).  If we're
adding noise to the output (stderr?), can we at least put it behind a
--log-level=[error|info|debug] flag or similar?

wking:
On Thu, Jun 02, 2016 at 07:29:23PM -0700, 梁辰晔 (Liang Chenye) wrote:
> Since you've changed REAMDE.me in this commit…

I like 16d4ed2b as it stands and am not wild about the logging change.
Can we bump the logging discussion to a follow-up PR?

liangchenye:
@wking yes, that is a follow-up PR. (with --log-level)
@hmeng-19  without the big log changes, you don't need to change validation.go, it already exists.
If a config.json passes the validation test, 'Bundle validation succeeded' is a nature output.
I prefer to have it as it reflecting the running status.

wking:
On Thu, Jun 02, 2016 at 11:20:31PM -0700, 梁辰晔 (Liang Chenye) wrote:
> @hmeng-19 without the big log changes, you don't need to change
> validation.go, it already exists.  If a config.json passes the
> validation test, 'Bundle validation succeeded' is a nature output.
> I prefer to have it as it reflecting the running status.

I don't think this PR or #91 have anything to do with a
printed-on-success message.  This is just about “don't require folks
to set --path”.  And the error message being removed from the README
suggests that current-directory default does not already exist.

liangchenye:
README.md tells how ocitools works.
It is right to remove the error message, but I think we should add other message back.
Running "./ocitools validate" with this commit merged, it will either tells "Bundle valiation succeeded" or something is wrong, there is no blank result.

wking:
On Fri, Jun 03, 2016 at 12:10:59AM -0700, 梁辰晔 (Liang Chenye) wrote:
> It is right to remove the error message, but I think we should add
> other message back.  Running "./ocitools validate" with this commit
> merged, it will either tells "Bundle valiation succeeded" or
> something is wrong, there is no blank result.

Ah.  I'd read your previous comments as:

  We should update validate.go to print “Bundle validation succeeded.”
  on success, and then reflect that change in the README.

What you're actually saying is:

  validate.go already prints “Bundle validation succeeded.”
  on success.  We should reflect that change in the README.

That sounds good to me.

hmeng-19:
@liangchenye , please give it a new look.

liangchenye:
LGTM
@mrunalp

mrunalp:
LGTM

PR #93, opened 2016-06-03
runtimetest: add validation of cgroups

Mashimiao:

### DIFF
--- a/cgroups/cgroups.go
+++ b/cgroups/cgroups.go
@@ -1,0 +1,72 @@
+package cgroups
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	rspec "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+// Cgroup represents interfaces for cgroup validation
+type Cgroup interface {
+	GetBlockIOData(cgPath string) (*rspec.LinuxBlockIO, error)
+	GetCPUData(cgPath string) (*rspec.LinuxCPU, error)
+	GetDevicesData(cgPath string) ([]rspec.LinuxDeviceCgroup, error)
+	GetHugepageLimitData(cgPath string) ([]rspec.LinuxHugepageLimit, error)
+	GetMemoryData(cgPath string) (*rspec.LinuxMemory, error)
+	GetNetworkData(cgPath string) (*rspec.LinuxNetwork, error)
+	GetPidsData(cgPath string) (*rspec.LinuxPids, error)
+}
+
+// FindCgroup gets cgroup root mountpoint
+func FindCgroup() (Cgroup, error) {
+	f, err := os.Open("/proc/self/mountinfo")
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+	cgroupv2 := false
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		text := scanner.Text()
+		fields := strings.Split(text, " ")
+		// Safe as mountinfo encodes mountpoints with spaces as \040.
+		index := strings.Index(text, " - ")
+		postSeparatorFields := strings.Fields(text[index+3:])
+		numPostFields := len(postSeparatorFields)
+
+		// This is an error as we can't detect if the mount is for "cgroup"
+		if numPostFields == 0 {
+			return nil, fmt.Errorf("Found no fields post '-' in %q", text)
+		}
+
+		if postSeparatorFields[0] == "cgroup" {
+			// Check that the mount is properly formated.
+			if numPostFields < 3 {
+				return nil, fmt.Errorf("Error found less than 3 fields post '-' in %q", text)
+			}
+
+			cg := &CgroupV1{
+				MountPath: filepath.Dir(fields[4]),
+			}
+			return cg, nil
+		} else if postSeparatorFields[0] == "cgroup2" {
+			cgroupv2 = true
+			continue
+			//TODO cgroupv2 unimplemented
+		}
+	}
+
+	if err := scanner.Err(); err != nil {
+		return nil, err
+	}
+
+	if cgroupv2 {
+		return nil, fmt.Errorf("cgroupv2 is not supported yet")
+	}
+	return nil, fmt.Errorf("cgroup is not found")
+}
--- a/cgroups/cgroups.go
+++ b/cgroups/cgroups.go
@@ -3,6 +3,7 @@ package cgroups
 import (
 	"bufio"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
@@ -10,15 +11,22 @@ import (
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
 )
 
+var (
+	// AbsCgroupPath is absolute path for container's cgroup mount
+	AbsCgroupPath = "/cgrouptest"
+	// RelCgroupPath is relative path for container's cgroup mount
+	RelCgroupPath = "testdir/cgrouptest/container"
+)
+
 // Cgroup represents interfaces for cgroup validation
 type Cgroup interface {
-	GetBlockIOData(cgPath string) (*rspec.LinuxBlockIO, error)
-	GetCPUData(cgPath string) (*rspec.LinuxCPU, error)
-	GetDevicesData(cgPath string) ([]rspec.LinuxDeviceCgroup, error)
-	GetHugepageLimitData(cgPath string) ([]rspec.LinuxHugepageLimit, error)
-	GetMemoryData(cgPath string) (*rspec.LinuxMemory, error)
-	GetNetworkData(cgPath string) (*rspec.LinuxNetwork, error)
-	GetPidsData(cgPath string) (*rspec.LinuxPids, error)
+	GetBlockIOData(pid int, cgPath string) (*rspec.LinuxBlockIO, error)
+	GetCPUData(pid int, cgPath string) (*rspec.LinuxCPU, error)
+	GetDevicesData(pid int, cgPath string) ([]rspec.LinuxDeviceCgroup, error)
+	GetHugepageLimitData(pid int, cgPath string) ([]rspec.LinuxHugepageLimit, error)
+	GetMemoryData(pid int, cgPath string) (*rspec.LinuxMemory, error)
+	GetNetworkData(pid int, cgPath string) (*rspec.LinuxNetwork, error)
+	GetPidsData(pid int, cgPath string) (*rspec.LinuxPids, error)
 }
 
 // FindCgroup gets cgroup root mountpoint
@@ -69,4 +77,28 @@ func FindCgroup() (Cgroup, error) {
 		return nil, fmt.Errorf("cgroupv2 is not supported yet")
 	}
 	return nil, fmt.Errorf("cgroup is not found")
+}
+
+// GetSubsystemPath gets path of subsystem
+func GetSubsystemPath(pid int, subsystem string) (string, error) {
+	contents, err := ioutil.ReadFile(fmt.Sprintf("/proc/%d/cgroup", pid))
+	if err != nil {
+		return "", err
+	}
+
+	parts := strings.Split(strings.TrimSpace(string(contents)), "\n")
+	for _, part := range parts {
+		elem := strings.SplitN(part, ":", 3)
+		if len(elem) < 3 {
+			continue
+		}
+		subelems := strings.Split(elem[1], ",")
+		for _, subelem := range subelems {
+			if subelem == subsystem {
+				return elem[2], nil
+			}
+		}
+	}
+
+	return "", fmt.Errorf("subsystem %s not found", subsystem)
 }
--- a/validation/linux_cgroups_blkio.go
+++ b/validation/linux_cgroups_blkio.go
@@ -13,7 +13,7 @@ func main() {
 	var major, minor int64 = 8, 0
 	var rate uint64 = 102400
 	g := util.GetDefaultGenerator()
-	g.SetLinuxCgroupsPath("/test")
+	g.SetLinuxCgroupsPath(cgroups.AbsCgroupPath)
 	g.SetLinuxResourcesBlockIOWeight(weight)
 	g.SetLinuxResourcesBlockIOLeafWeight(leafWeight)
 	g.AddLinuxResourcesBlockIOWeightDevice(major, minor, weight)
@@ -22,12 +22,12 @@ func main() {
 	g.AddLinuxResourcesBlockIOThrottleWriteBpsDevice(major, minor, rate)
 	g.AddLinuxResourcesBlockIOThrottleReadIOPSDevice(major, minor, rate)
 	g.AddLinuxResourcesBlockIOThrottleWriteIOPSDevice(major, minor, rate)
-	err := util.RuntimeOutsideValidate(g, func(path string) error {
+	err := util.RuntimeOutsideValidate(g, cgroups.AbsCgroupPath, func(pid int, path string) error {
 		cg, err := cgroups.FindCgroup()
 		if err != nil {
 			return err
 		}
-		lbd, err := cg.GetBlockIOData(path)
+		lbd, err := cg.GetBlockIOData(pid, path)
 		if err != nil {
 			return err
 		}
--- a/validation/linux_cgroups_blkio.go
+++ b/validation/linux_cgroups_blkio.go
@@ -1,0 +1,115 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/opencontainers/runtime-tools/cgroups"
+	"github.com/opencontainers/runtime-tools/validation/util"
+)
+
+func main() {
+	var weight uint16 = 500
+	var leafWeight uint16 = 300
+	var major, minor int64 = 8, 0
+	var rate uint64 = 102400
+	g := util.GetDefaultGenerator()
+	g.SetLinuxCgroupsPath("/test")
+	g.SetLinuxResourcesBlockIOWeight(weight)
+	g.SetLinuxResourcesBlockIOLeafWeight(leafWeight)
+	g.AddLinuxResourcesBlockIOWeightDevice(major, minor, weight)
+	g.AddLinuxResourcesBlockIOLeafWeightDevice(major, minor, leafWeight)
+	g.AddLinuxResourcesBlockIOThrottleReadBpsDevice(major, minor, rate)
+	g.AddLinuxResourcesBlockIOThrottleWriteBpsDevice(major, minor, rate)
+	g.AddLinuxResourcesBlockIOThrottleReadIOPSDevice(major, minor, rate)
+	g.AddLinuxResourcesBlockIOThrottleWriteIOPSDevice(major, minor, rate)
+	err := util.RuntimeOutsideValidate(g, func(path string) error {
+		cg, err := cgroups.FindCgroup()
+		if err != nil {
+			return err
+		}
+		lbd, err := cg.GetBlockIOData(path)
+		if err != nil {
+			return err
+		}
+		if *lbd.Weight != weight {
+			return fmt.Errorf("blkio weight is not set correctly, expect: %d, actual: %d", weight, lbd.Weight)
+		}
+		if *lbd.LeafWeight != leafWeight {
+			return fmt.Errorf("blkio leafWeight is not set correctly, expect: %d, actual: %d", weight, lbd.LeafWeight)
+		}
+
+		found := false
+		for _, wd := range lbd.WeightDevice {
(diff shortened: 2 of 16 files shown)


cmd/runtimetest/main.go
```diff
@@ -244,6 +245,45 @@ func validateROPaths(spec *rspec.Spec) error {
...
+			}
+		}
+	}
+	return "", fmt.Errorf("cgroup %v not found", subsystem)
+}
+
+func validateCgroupsPath(spec *rspec.Spec) error {
+	fmt.Println("validating cgroupsPath")
+	expectedPath := spec.Linux.CgroupsPath
+	actualPath, err := getThisCgroupPath("devices")
```
wking: Why are we blessing `devices`?  What if the runtime is using [unified cgroups][1]?

[1]: https://www.kernel.org/doc/Documentation/cgroup-v2.txt

wking:
On Thu, Jun 02, 2016 at 07:39:53PM -0700, Ma Shimiao wrote:
>   * runtimetest: add validation of cgrup path

226e9a3 has the “cgrup” → “cgroup” typo in the commit-message still.

cmd/runtimetest/main.go
Mashimiao: Well, We need many default devices to create container. So cgroup device should be hard requirement.  
The spec says "The Spec does not support split hierarchy[https://www.kernel.org/doc/Documentation/cgroup-v2.txt]. The cgroups will be created if they don't exist."
unified cgroups may not to be supported?

wking: I think this needs some spec work before we can implement something reliable.  I expect we'll want to make this comparison for unified cgroups if the container process belongs to the unified hierarchy, and with ([some][1]?) v1 cgroups if the container does not belong to the unified hierarchy.

[1]: https://github.com/opencontainers/runc/issues/861

wking: On Thu, Jun 02, 2016 at 10:31:32PM -0700, Ma Shimiao wrote:
> > +	actualPath, err := getThisCgroupPath("devices")
> 
> Well, We need many default devices to create container. So cgroup
> device should be hard requirement.

But just because you create devices doesn't mean the user is setting
anything in linux.resources.devices [1].

> The spec says "The Spec does not support split
> hierarchy[https://www.kernel.org/doc/Documentation/cgroup-v2.txt]. The
> cgroups will be created if they don't exist."  unified cgroups may
> not to be supported?

No, that wording is “we don't support per-controller paths” for the
reasons discussed in cgroup-v2.txt [2].  It's not about limiting the
cgroup version.

[1]: https://github.com/opencontainers/runtime-spec/blob/v0.5.0/config-linux.md#device-whitelist
[2]: https://github.com/opencontainers/runtime-spec/pull/137#discussion_r38601118

Mashimiao: Thanks for your information. I misunderstood the referred link.
I will make more investigation and update the PR later.

cyphar:
I agree with @wking that the runtime-spec still needs some more work around `cgroupPath` as well as general cgroup handling. I'll work on a PR for this soon.

Mashimiao:
@cyphar  OK, I will update the PR until we have a clear status of cgroupPath

Mashimiao:
ping @wking  @cyphar  @mrunalp

cmd/runtimetest/main.go
```diff
@@ -387,14 +434,15 @@ func validate(context *cli.Context) error {
...
+		validateCgroupsPath,
 		validateHostname,
 		validateRlimits,
 		validateMountsExist,
 	}
 
 	linuxValidations := []validation{
 		validateDefaultFS,
-		validateDefaultDevices,
+//		validateDefaultDevices,
```
wking: Why comment this out?

cmd/runtimetest/main.go
```diff
@@ -370,6 +371,52 @@ func validateMountsExist(spec *rspec.Spec) error {
...
+
+	s := bufio.NewScanner(f)
+	for s.Scan() {
+		if err := s.Err(); err != nil {
+			return "", err
+		}
+
+		text := s.Text()
+		parts := strings.Split(text, ":")
+		return parts[2], nil
```
wking: With 946b75e → 8d6fcb6 it looks like you are no longer blessing `devices`, but are instead returning the first entry you find.  That's fine as long as it happens to be one of the controllers managed by the runtime, but opencontainers/runtime-spec#493 only requires the runtime to manage controllers needed to enforce `linux.resources`.  If the first entry happens to *not* be one of those controllers, the path you're extracting here may not be the unified path used by controllers that the runtime is managing.  I think the logic should be:

> For all controllers needed for `linux.resources` (may be an empty set), ensure that they have the same path.  Return that path for `validateCgroupsPath`.

cmd/runtimetest/main.go
```diff
@@ -370,6 +371,52 @@ func validateMountsExist(spec *rspec.Spec) error {
...
+	return "", fmt.Errorf("can not get cgroup path")
+}
+
+func validateCgroupsPath(spec *rspec.Spec) error {
+	logrus.Debugf("validating cgroupsPath")
+	expectedPath := spec.Linux.CgroupsPath
+	if expectedPath == nil {
+		return nil
+	}
+	*expectedPath = strings.Replace(*expectedPath, ":", "/", -1)
```
wking: What's going on here?  I suspect something about systemd slicing, but there's no mention of colon→slash changes in opencontainers/runtime-spec#493, so I think we need to leave it out here.

cmd/runtimetest/main.go
```diff
@@ -370,6 +371,52 @@ func validateMountsExist(spec *rspec.Spec) error {
...
+	if err != nil {
+		return err
+	}
+
+	if filepath.IsAbs(*expectedPath) {
+		if *expectedPath != actualPath {
+			return fmt.Errorf("Cgroup path expected: %v, actual: %v", *expectedPath, actualPath)
+		}
+	} else {
+		*expectedPath = strings.Replace(*expectedPath, "../", "", -1)
```
wking: Also not sure what's going on here.  Why would you have `../` in `expectedPath`?  That sounds like an invalid `cgroupsPath` to me.

cmd/runtimetest/main.go
```diff
@@ -370,6 +371,52 @@ func validateMountsExist(spec *rspec.Spec) error {
...
+		if !strings.Contains(actualPath, *expectedPath) {
```
wking: `strings.Contains` is a difficult way to do this (e.g. what if `actualPath` is `ab/cd` but `*expectedPath` is `b/cd`?).  Can we use [`filepath.Rel`][1] instead?

[1]: https://golang.org/pkg/path/filepath/#Rel

cmd/runtimetest/main.go
```diff
@@ -370,6 +371,52 @@ func validateMountsExist(spec *rspec.Spec) error {
...
+			return fmt.Errorf("Cgroup path expected: %v, actual: %v", *expectedPath, actualPath)
```
wking: opencontainers/runtime-spec#493 only has MAY wording for relative `cgroupsPath`, so this should be a warning and not an error.

cmd/runtimetest/main.go
Mashimiao: Sorry, this is a mistake

cmd/runtimetest/main.go
cyphar: In fact, in runC we clean the `cgroupsPath` quite strongly to avoid path traversal vulnerabilities. So I have the feeling that we would fail this test case from the runC side.

wking: On Mon, Aug 01, 2016 at 01:24:04AM -0700, Aleksa Sarai wrote:
> > +	if filepath.IsAbs(*expectedPath) {
> > +		if *expectedPath != actualPath {
> > +			return fmt.Errorf("Cgroup path expected: %v, actual: %v", *expectedPath, actualPath)
> > +		}
> > +	} else {
> > +		*expectedPath = strings.Replace(*expectedPath, "../", "", -1)
> 
> In fact, in runC we clean the `cgroupsPath` quite strongly to avoid
> path traversal vulnerabilities. So I have the feeling that we would
> fail this test case from the runC side.

The not-absolute spec is an unfailable MAY, so no worries ;).

cmd/runtimetest/main.go
wking: On Sun, Jul 31, 2016 at 09:36:26PM -0700, W. Trevor King wrote:
> > For all controllers needed for `linux.resources` (may be an empty
> > set), ensure that they have the same path.  Return that path for
> > `validateCgroupsPath`.

This would miss runtimes that for whatever reason attached an unneeded
controller with an inconsistent path.  Better logic:

* Pass the host controller paths in to the container (like we already
  pass config.json in).  This could be via a new file, or via
  config.json's annotations.
* For each entry in /proc/self/cgroup, assert that the path is the
  same as the host path for that controller, or some new path.  If
  there are multiple controllers with new paths, ensure they all
  match.

In pseudocode:

  var containerPath *string
  for s.Scan() {
    …
    parts := strings.Split(text, ":")
    // should probably check to make sure len(parts) == 3
    hierarchyId := parts[0]
    newPath := parts[2]
    oldPath := oldPaths[hierarchyId]
    if newPath == oldPath {
      controllers := strings.Split(parts[1], ",")
      for _, controller := range(controllers) {
        err = assertCgroupUnchanged(controller, oldControllerConfigs[controller])
        // log the error
      }
    } else if containerPath == nil {
      containerPath = newPath
    } else if newPath != containerPath {
      // log the error
    }
  }
  return containerPath, nil

cyphar:
I don't understand any of the uses of `strings.Replace` with paths (nor `strings.Contains`). If you're trying to convert a path so that it no longer contains leading `../`s then you can use `libcontainer/utils.CleanPath` for that purpose. It cleans both absolute and relative paths in a sane way.

As for replacing `:` with `/`, I don't get what you're trying to do (are you trying to generate a systemd slice path from a name?).

wking:
On Mon, Aug 01, 2016 at 01:52:15PM -0700, Aleksa Sarai wrote:
> If you're trying to convert a path so that it no longer contains
> leading `../`s then you can use `libcontainer/utils.CleanPath` for
> that purpose.

I'd rather not introduce a libcontainer dependency, because ocitools
is intended to test runC/libcontainer (among other runtimes).  But in
this case, I don't see a need to clean ../ at all, since they are not
mentioned in the current runtime-spec.

I'd be ok landing language in runtime-spec about ‘..’ being an illegal
path component in cgroupsPath, but we don't have that language now, so
ocitools shouldn't do anything special for such paths.

cmd/runtimetest/main.go
Mashimiao: Fine, fixed.

cmd/runtimetest/main.go
Mashimiao: OK

cmd/runtimetest/main.go
Mashimiao: There is actually not mention of colon->slash. But runc do like this. I we leave this out, I guess most of runtime may fail.

cmd/runtimetest/main.go
Mashimiao: @wking  The Spec says `If a cgroupsPath value is specified, the runtime MUST consistently attach to the same place in the cgroup hierarchy given the same value of cgroupsPath.`
So, I think all controllers'path in /proc/self/cgroup must be same. If not, the validation should be fail. If it is, return the path. There is no need to be so complicated.

cmd/runtimetest/main.go
cyphar: The problem is that we're applying stronger limits on the cgroupsPath logic than actually exist in the spec. The spec doesn't say how relative paths must be handled -- the only constraint is that it join the same place in the hierarchy consistently. So we can test that if you want, but testing any case where you're checking the actual value of the cgroup paths against some predefined value and `!filepath.IsAbs` is going beyond the specification -- even if runC will pass the tests no other runtime is obligated to.

cmd/runtimetest/main.go
```diff
@@ -370,6 +371,62 @@ func validateMountsExist(spec *rspec.Spec) error {
...
+	if err != nil {
+		return err
+	}
+
+	if filepath.IsAbs(*expectedPath) {
+		if *expectedPath != actualPath {
+			return fmt.Errorf("Cgroup path expected: %v, actual: %v", *expectedPath, actualPath)
+		}
+	} else {
+		if _, err := filepath.Rel(*expectedPath, actualPath); err != nil {
```
cyphar: Basically, this branch should either be a noop or should create a new container and compare the value. Those are the only two things we can safely do here.

cmd/runtimetest/main.go
wking: On Tue, Aug 02, 2016 at 02:18:35AM -0700, Ma Shimiao wrote:
> @wking The Spec says `If a cgroupsPath value is specified, the
> runtime MUST consistently attach to the same place in the cgroup
> hierarchy given the same value of cgroupsPath.`

This is “for multiple runtime invocations”, which is something you can
test via a test harness, but not something you can test by
runtimetest, which only knows about the container it is inside.

> So, I think all controllers'path in /proc/self/cgroup must be
> same. If not, the validation should be fail. If it is, return the
> path. There is no need to be so complicated.

The reason controllers touched by the runtime need to use the same
path is [1]:

  The Spec does not support per-controller paths for the reasons
  discussed in the [cgroupv2 documentation][cgroup-v2].

The reason the host paths are also an option is [2]:

  A runtime MUST at least use the minimum set of cgroup controllers
  required to fulfill the `resources` settings.  However, a runtime
  MAY attach the container process to additional cgroup controllers
  supported by the system.

which means the runtime is free to ignore controllers that aren't
related to the configured ‘linux.resources’.  Those ignored
controllers will inherit whatever paths the host process had.

[1]: https://github.com/opencontainers/runtime-spec/pull/493/files#diff-428eec4013a655816cdefafd5d3505f1R171
[2]: https://github.com/opencontainers/runtime-spec/pull/493/files#diff-428eec4013a655816cdefafd5d3505f1R178

cmd/runtimetest/main.go
```diff
@@ -370,6 +371,62 @@ func validateMountsExist(spec *rspec.Spec) error {
...
+	for s.Scan() {
+		if err := s.Err(); err != nil {
+			return "", err
+		}
+
+		text := s.Text()
+		parts := strings.Split(text, ":")
+		if cgroupPath == parts[2] && cgroupPath != "" {
+			continue
+		} else if cgroupPath == "" && parts[2] != "" {
```
wking: is `parts[2] == ""` possible?  If we want to guard against it, I'd rather have a check like:

    if len(parts) != 3 || parts[2] == "" {
        return "", fmt.Errorf("invalid line in /proc/self/cgroup: %q", text)
    }

right after the `Split`.

cmd/runtimetest/main.go
```diff
@@ -370,6 +371,62 @@ func validateMountsExist(spec *rspec.Spec) error {
...
+			cgroupPath = parts[2]
+		} else {
+			return "", fmt.Errorf("error path with cgroup controllers")
```
wking: This will break for runtimes that ignore unneeded cgroups (which opencontainers/runtime-spec#493 allows).  See [here][1] and [here][2] for more.

[1]: https://github.com/opencontainers/ocitools/pull/93#discussion_r73002225
[2]: https://github.com/opencontainers/ocitools/pull/93#discussion_r73188918

cmd/runtimetest/main.go
wking: It currently [warns][1] about the MAY violation, which I think is fine.

[1]: https://github.com/opencontainers/ocitools/pull/93#discussion_r72923398

wking: Although I think `Rel` *erroring* isn't the right condition.  We want to warn if the path returned by `Rel` starts with `..`.

wking: And “starts with `..`” should be “the first path segment (split on the path separator) is `..`” not “the path string starts with the two characters `..`”.

cmd/runtimetest/main.go
Mashimiao: I will remove it.

cmd/runtimetest/main.go
Mashimiao: I think `parts[2] == ""` is not possible.  If controller was not mounted, there would not be item in /proc/self/cgroup

Mashimiao:
Test all cgroups except devices subsystem for croupv1

Mashimiao:

validation/util/test.go
```diff
@@ -138,3 +138,41 @@ func RuntimeInsideValidate(g *generate.Generator, f PreFunc) (err error) {
...
+	if err != nil {
+		return err
+	}
+	err = fileutils.CopyFile("runtimetest", filepath.Join(r.BundleDir, "runtimetest"))
+	if err != nil {
+		return err
+	}
+
+	r.SetID(uuid.NewV4().String())
+	stderr, err := r.Create()
```
liangchenye: no 'r.start'?

wking: > no 'r.start'?

What would you start?  The [bulk of the container is setup by the time `create` returns][1], and the runtime's dummy process is perfect for holding the container open while it is inspected from outside.  So I don't think we want a `start` call in `RuntimeOutsideValidate`.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/runtime.md#L103

validation/util/test.go
```diff
@@ -138,3 +138,41 @@ func RuntimeInsideValidate(g *generate.Generator, f PreFunc) (err error) {
...
+	if err != nil {
+		os.Stderr.WriteString("failed to start the container\n")
+		os.Stderr.Write(stderr)
+		return err
+	}
+
+	if f != nil {
+		if err := f("test"); err != nil {
```
wking: I don't understand this hard-coded `test`.  Don't we want to pass the bundle path, like [we do for `RuntimeInsideValidate`][1]?

[1]: https://github.com/opencontainers/runtime-tools/blob/acbe085c54698d0ccb98892efbc233f61d2faed5/validation/util/test.go#L85

validation/linux_cgroups_blkio.go
```diff
@@ -0,0 +1,115 @@
...
+	g.AddLinuxResourcesBlockIOThrottleReadBpsDevice(major, minor, rate)
+	g.AddLinuxResourcesBlockIOThrottleWriteBpsDevice(major, minor, rate)
+	g.AddLinuxResourcesBlockIOThrottleReadIOPSDevice(major, minor, rate)
+	g.AddLinuxResourcesBlockIOThrottleWriteIOPSDevice(major, minor, rate)
+	err := util.RuntimeOutsideValidate(g, func(path string) error {
+		cg, err := cgroups.FindCgroup()
+		if err != nil {
+			return err
+		}
+		lbd, err := cg.GetBlockIOData(path)
```
wking: I think `path` should point to the bundle directory, which means we'll need something else for finding the cgroup path.  I think we want two tests for absolute `cgroupsPath`:

* Get the container process PID using `state`, and read in `/proc/{pid}/cgroup` to get the associated paths.  Compare those paths with `cgroupsPath` to test “[the runtime MUST take the path to be relative to the cgroups mount point][1]”, and raise `CgroupsAbsPathRelToMount` on mismatch.
* Then read the cgroup limits from whichever group you read from `/proc/{pid}/cgroup` (which may or may not be at the correct path), and compare the value with the configured value.  Failures here should probably be non-spec warnings, because [the spec has no RFC 2119 requirements for cgroup values][2].

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/config-linux.md#L181
[2]: https://github.com/opencontainers/runtime-spec/issues/746#issuecomment-297482403

validation/util/test.go
Mashimiao: No, here just to pass absolute cgroup path.
I will update PRs to support absolute and relative cgroup path validation

Mashimiao:
Add support for relative cgrouppath test and add cgroup tests related with it

cgroups/cgroups.go
```diff
@@ -0,0 +1,104 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+
+	rspec "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+var (
+	// AbsCgroupPath is absolute path for container's cgroup mount
+	AbsCgroupPath = "/cgrouptest"
```
liangchenye: Do we need to put these two variables into the cgroups pkg?
Beside these, the PR looks good to me.

liangchenye:
LGTM

cgroups/cgroups.go
liangchenye: We can change it later if there is a better place to leave these.

q384566678:
LGTM

PR #95, opened 2016-06-03
validate: Improved version handling

wking:
Use an off-the-shelf SemVer parser and error out if the config targets
an unsupported spec version.  Slightly more detail in the commit
messages ;).

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -8,6 +8,11 @@
 			"Rev": "07d998d174c4e2dc90e2f1989a20724220bca1ff"
 		},
 		{
+			"ImportPath": "github.com/blang/semver",
+			"Comment": "v3.1.0",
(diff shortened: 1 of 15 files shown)


wking:
It looks like I'm missing some Godeps magic.  Can anyone point me in the right direction?

mrunalp:
@wking If you have `"github.com/blang/semver"` in your $GOPATH, then just run godep save after changing to ocitools directory. Your git diff should now show a change in Godeps.json to add this dependency and copy the files under Godeps/_workspace.

wking:
On Fri, Jun 03, 2016 at 04:35:01PM -0700, Mrunal Patel wrote:
> … just run godep save …

Thanks.  Added with 629e326 → be8c9ff.

mrunalp:
LGTM. @liangchenye PTAL.

liangchenye:
LGTM

Issue #98, opened 2016-06-05
Use Python's unittest for validating bundles and configurations?


wking:
Currently we're using Go, which means it's hard to build a single validator that can process multiple versions of the spec, because you'd have to use tedious [`interface{}` manipulation][1] or load separate packages with structures for each supported version.

We're also not using a test suite, which means it's hard to test multiple aspects of the configuration and report on several errors (the current code just [dies on the first error][2]).

By using Python's [unittest][3], we get more convenient handling of generic JSON and an established test framework that doesn't need a lot of boilerplate.  I've started working up this approach [here][4] if folks want to kick the tires.  Examples of potentially tricky things and how this approach lets us handle them easily:

* Skipping particular tests if the configuration has an unrecognized version (wking/oci-runtime-config-validator@c607380f).
* Applying the Windows-specific mount-nesting restriction ([here][5]).
* Applying the 0.5.0-specific (and previous, but I haven't bothered supporting them) relative `root.path` requirement ([here][6]).

Example compact output with the current tip (wking/oci-runtime-config-validator@cd0facf0a.  I still haven't finished `process`, later entries in `config.md` or anything from the platform-specific files):

    $ BUNDLE=~/src/opencontainers/my-app python3 -m unittest
    ...s.............
    ----------------------------------------------------------------------
    Ran 17 tests in 0.006s

    OK (skipped=1)

and verbose output:

    $ BUNDLE=~/src/opencontainers/my-app python3 -m unittest -v
    test_configuration (test.test_bundle.TestBundle)
    config.json MUST reside in the root of the bundle directory. ... ok
    test_root (test.test_bundle.TestBundle)
    The bundle directory MUST contain the root filesystem. ... ok
    test_destination (test.test_mounts.TestMounts)
    destination (string, required). ... ok
    test_destination_nesting (test.test_mounts.TestMounts)
    Mount destinations MUST not be nested within another mount. ... skipped 'the destination-nesting restriction only applies to Windows for specification version 1.0.0-rc1.'
    test_options (test.test_mounts.TestMounts)
    options (list of strings, optional). ... ok
    test_source (test.test_mounts.TestMounts)
    source (string, required). ... ok
    test_type (test.test_mounts.TestMounts)
    type (string, required). ... ok
    test_args (test.test_process.TestProcess)
    args (array of strings, required). ... ok
    test_cwd (test.test_process.TestProcess)
    cwd (string, required). ... ok
    test_env (test.test_process.TestProcess)
[...]

PR #97, opened 2016-06-04
Add --log-level flags to ocitools and runtimetest

wking:
Details in the commit messages.

PR #99, opened 2016-06-06
Change Fatal to Error

liangchenye:
Fix `dies on the first error` issue mentioned in https://github.com/opencontainers/ocitools/issues/98.
Check/validate all the tests, output all the error messages.

It works better with https://github.com/opencontainers/ocitools/pull/97.

### DIFF
--- a/validate.go
+++ b/validate.go
@@ -3,6 +3,7 @@ package main
 import (
 	"bufio"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -18,6 +19,8 @@ import (
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/urfave/cli"
 )
+
+type configCheck func(rspec.Spec, string, bool) []string
 
 var bundleValidateFlags = []cli.Flag{
 	cli.StringFlag{Name: "path", Value: ".", Usage: "path to a bundle"},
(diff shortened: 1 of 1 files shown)


cmd/runtimetest/main.go
```diff
@@ -257,9 +257,12 @@ func main() {
 		validateROPaths,
 	}
 
+	ret := 0
+	defer os.Exit(ret)
```
wking: Why defer here, instead of calling `os.Exit` after the loop completes?  The only case I can see is if validation panics, and presumably that is already leading to a non-zero exit status.

validate.go
```diff
@@ -78,29 +81,48 @@ var bundleValidateCommand = cli.Command{
...
+			checkMandatoryFields,
+			checkSemVer,
+			checkPlatform,
+			checkProcess,
+			checkLinux,
+			checkHooks,
+		}
+
+		ret := 0
+		defer os.Exit(ret)
```
wking: My [“why `defer`?” question][1] also applies here.

[1]: https://github.com/opencontainers/ocitools/pull/99/files#r65918706

wking:
On Mon, Jun 06, 2016 at 05:09:33AM -0700, 梁辰晔 (Liang Chenye) wrote:
> Fix `dies on the first error` issue mentioned in
> https://github.com/opencontainers/ocitools/issues/98.

I think this approach is an improvement, but it seems like more work
than using an off-the-shelf test framework (which is what I'm
recommending in #98).

validate.go
liangchenye: updated.

mrunalp:
:+1: to the PR. Needs rebase.

liangchenye:
@mrunalp Better output in this PR and is rebased. PTAL

```
FATA[0000] 5 Errors detected:
  1. 'Platform.OS' should not be empty.
  2. "0.x.0-invalid" is not valid SemVer: Invalid character(s) found in minor number "x"
  3. internal error: validate currently only handles version 1.0.0-rc1-dev, but the supplied configuration targets 0.x.0-invalid
  4. Operation system "" of the bundle is not supported yet.
  5. capability "CAP_CHOWN_ERROR" is not valid, man capabilities(7)
```

mrunalp:
LGTM

PR #104, opened 2016-06-07
Update Action Function signature for validate

mrunalp:
The signature now requires returning an error.


wking:
On Tue, Jun 07, 2016 at 02:18:46PM -0700, Mrunal Patel wrote:
> The signature now requires returning an error.

Ah, the Go ecosystem :p.  I guess backwards compatibility breakage
should not be surprising ;).  generate.go is using Action too.

wking:
0d613b7b looks good to me.

liangchenye:
LGTM

PR #105, opened 2016-06-07
Remove non-standard cgroups mount from generated config

mrunalp:
Needed for runc runtime validation to pass after #81 


wking:
On Tue, Jun 07, 2016 at 02:29:35PM -0700, Mrunal Patel wrote:
>   * Remove non-standard cgroups mount from generated config

This is the default since #48, no?

mrunalp:
@wking Hmm, yes, let me test again.

mrunalp:
Confirmed that this ins't required. Thanks!

PR #107, opened 2016-06-08
generate: Add support for --seccomp-errno

mrunalp:


wking:
64f8060 is missing a man-page entry.  And it seems like it's probably
worth adding flags for all the seccomp stuff at once [1], although we
can nibble the actions off one at a time if you like.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/config-linux.md#seccomp

wking:
Wait, can't you use the existing --seccomp-syscalls (from
mrunalp/ocitools#13) for this?

  --seccomp-syscalls getcwd:SCMP_ACT_ERRNO:

mrunalp:
man page updated.

@wking These are meant to be more convenient to use.

wking:
On Tue, Jun 14, 2016 at 11:04:59AM -0700, Mrunal Patel wrote:
> @wking These are meant to be more convenient to use. 

Probably mention that with cross-references from the man-page entries
then.  E.g. from --seccomp-syscalls mention the more-convenient
--seccomp-allow and --seccomp-errno, and from the latter mention the
more-generic --seccomp-syscalls.

mrunalp:
Updated man page with cross-references.

liangchenye:
LGTM

Issue #108, opened 2016-06-13
Bind this repository (or a branch) to runtime-spec releases


wking:
This is somewhat related to #16 (where [a comment][1] sounded like “we'll get serious once we get to 1.0”), and #69 (where I argue that [we should not float with runC][2]).  But concretely, I am unable to test ccon with the current ocitools.  For one, my command-line API is different (I expect `RUNTIME start --id $CONTAINER_ID`, while runC seems to prefer `RUNTIME start $CONTAINER_ID`).  But once I've patched around that:

    ocitools$ git diff -U0
    diff --git a/test_runtime.sh b/test_runtime.sh
    index da3f026..d498ae8 100755
    --- a/test_runtime.sh
    +++ b/test_runtime.sh
    @@ -84 +84 @@ popd > /dev/null
    -TESTCMD="${RUNTIME} start $(uuidgen)"
    +TESTCMD="${RUNTIME} start --id $(uuidgen)"

I still cannot run the tests, because they're based on an unreleased 0.6.0-dev version of the spec:

    ocitools$ ./test_runtime.sh -l debug -r ccon-oci
    -----------------------------------------------------------------------------------
    VALIDATING RUNTIME: ccon-oci
    -----------------------------------------------------------------------------------
    Traceback (most recent call last):
      File "/home/wking/bin/ccon-oci", line 578, in <module>
        main()
      File "/home/wking/bin/ccon-oci", line 574, in main
        args.func(**kwargs)
      File "/home/wking/bin/ccon-oci", line 442, in start
        config_state = _load_config(bundle=bundle, id=id)
      File "/home/wking/bin/ccon-oci", line 421, in _load_config
        version = _get_version(config=config)
      File "/home/wking/bin/ccon-oci", line 64, in _get_version
        'unsupported OCI-spec version: {}'.format(version))
    NotImplementedError: unsupported OCI-spec version: 0.6.0-dev
    Runtime ccon-oci failed validation

Is there a branching plan for supporting per-version testing?  For example, we might want to start a v1.0 branch to track the v1.0 release series while `master` continues on with any changes destined for v1.1 / v2.0, etc.  Then the v1.0 branch could generate (at the moment) 1.0.0-rc1 specs that ccon-oci understands.

[1]: https://github.com/opencontainers/ocitools/issues/16#issuecomment-174609225
[2]: https://github.com/opencontainers/ocitools/issues/69#issuecomment-223844232

wking:
[...]

PR #112, opened 2016-06-18
generate: Adjust to Spec.Linux being a pointer

wking:
Catch up with the in-flight opencontainers/runtime-spec@6323157 (specs-go/config: Make Linux and Solaris omitempty (again), 2016-06-17, opencontainers/runtime-spec#502).

The “does the marshaled JSON look like '{}'?” check is a pretty cheap trick, but it was the easiest way I could think of for “is there anything useful in here?”.  An alternative approach that conditionally added an `&rspec.Linux{}` only if needed seemed too tedious.

This should not land before opencontainers/runtime-spec#502, and it may not even be worth reviewing until then.  But I'd worked up the changes while testing the runtime-spec PR, so I thought I might as well push them in case other folks wanted to see how I'd tested it.

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -196,6 +196,16 @@ func (g *Generator) Spec() *rspec.Spec {
 func (g *Generator) Save(w io.Writer, exportOpts ExportOptions) (err error) {
 	var data []byte
 
+	if g.spec.Linux != nil {
+		buf, err := json.Marshal(g.spec.Linux)
+		if err != nil {
+			return err
+		}
+		if string(buf) == "{}" {
+			g.spec.Linux = nil
+		}
+	}
+
 	if exportOpts.Seccomp {
 		data, err = json.MarshalIndent(g.spec.Linux.Seccomp, "", "\t")
 	} else {


liangchenye:
Your commit to runtime spec is checked in.

I don't quite like the '{}' way, how about using 'reflect.DeepEqual', like this:
```
if reflect.DeepEqual(Spec.Linux, &rspec.Linux{}) {
     spec.Linux = nil
}
```

wking:
On Tue, Jun 28, 2016 at 02:57:30AM -0700, 梁辰晔 (Liang Chenye) wrote:
> Your commit to runtime spec is checked in.

I'd rather hold off on landing this PR until runtime-spec cuts another
tag (#108).

> I don't quite like the '{}' way, how about using
> 'reflect.DeepEqual', like this:
> ```
> if reflect.DeepEqual(Spec.Linux, &rspec.Linux{}) {
>      spec.Linux = nil
> }
> ```

A agree that the JSON marshaling feels awkward, but it is checking
what we want (“will this serialize to really boring JSON?”).  I tried
the DeepEqual way and still had a Linux entry with:

  $ ./ocitools generate -template <(echo '{}')

My guess is that DeepEquals doesn't work because spec.Annotations is
an empty map [1] (instead of nil), or some such.

[1]: https://github.com/opencontainers/ocitools/blob/aa9cba2bdf52cddd060bf9fbe4bee47e5bfc802a/generate.go#L194

wking:
Rebased around #131 with 965e585 → 1cd5062.  Now that there are more
potential entry points, I needed more “Is Linux nil?” guards, but I
think I added them in the right places ;).

wking:
Rebased onto master with 1cd5062 → f2107e5.  Between #140 and #144,
the only thing left here is the “don't serialize Spec.Linux as an
empty JSON object” check.  Although to notice now, you have to do
something like:

  $ ocitools generate --template <(echo '{"linux": {"namespaces": [{"type": "user"}]}}') --user=host

which generates a ‘"linux": {}’ entry on master but not with this
branch.

generate/generate.go
```diff
@@ -959,6 +969,9 @@ func (g *Generator) SetupPrivileged(privileged bool) {
 		g.spec.Process.Capabilities = finalCapList
 		g.spec.Process.SelinuxLabel = ""
 		g.spec.Process.ApparmorProfile = ""
+		if g.spec.Linux == nil {
+			g.spec.Linux = &rspec.Linux{}
+		}
 		g.spec.Linux.Seccomp = nil
```
Mashimiao: I think this is not needed any more. Because we have g.initSpecLinux()

wking:
On Fri, Sep 23, 2016 at 01:31:59AM -0700, Ma Shimiao wrote:
> > @@ -959,6 +969,9 @@ func (g *Generator) SetupPrivileged(privileged bool) {
>  		g.spec.Process.Capabilities = finalCapList
>  		g.spec.Process.SelinuxLabel = ""
>  		g.spec.Process.ApparmorProfile = ""
> +		if g.spec.Linux == nil {
> +			g.spec.Linux = &rspec.Linux{}
> +		}
>  		g.spec.Linux.Seccomp = nil
> 
> I think this is not needed any more. Because we have g.initSpecLinux() 

Dropped and rebased onto master with f2107e5 → 7fcc366.

Mashimiao:
need rebase @wking

wking:
On Sun, Dec 18, 2016 at 05:57:26PM -0800, Ma Shimiao wrote:
> need rebase @wking 

Rebased with 7fcc366 → 9004f6d.

Mashimiao:
@liangchenye  any comments?

liangchenye:
https://golang.org/src/encoding/json/encode.go?s=5895:5938#L585
I think this solution is the simplest way (to my knowledge) and won't bring any side-effect.

Mashimiao:
LGTM

PR #114, opened 2016-06-18
test_runtime.sh: Add a user namespace

wking:
You shouldn't need to be root to test a runtime.  The `id` calls use the
[POSIX command](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/id.html) to find the current user's user and group IDs.

Builds on #113, so review that first.

### DIFF
--- a/test_runtime.sh
+++ b/test_runtime.sh
@@ -78,7 +78,7 @@ tar -xf  rootfs.tar.gz -C ${TESTDIR}
 cp runtimetest ${TESTDIR}
 
 pushd $TESTDIR > /dev/null
-ocitools generate "${TEST_ARGS[@]}" --rootfs '.'
+ocitools generate "${TEST_ARGS[@]}" --rootfs '.' --user '' --uidmappings "$(id -u):0:1" --gidmappings "$(id -g):0:1"
 popd > /dev/null
 
 TESTCMD="${RUNTIME} start $(uuidgen)"


wking:
An alternative to picking one way or the other is to use a test framework with both and skip the entries which require root when the caller isn't.  For examples in the ccon test suite, see:

* The [definition of the ROOT flag][1].
* A [test that runs regardless of the caller's root-ness][2].
* A [test that runs only for root users][3].
* A [test that runs only for non-root users][4].

[1]: https://github.com/wking/ccon/blob/v0.4.0/test/sharness.d/root.sh#L17
[2]: https://github.com/wking/ccon/blob/v0.4.0/test/t1001-process-user.t#L22
[3]: https://github.com/wking/ccon/blob/v0.4.0/test/t1001-process-user.t#L55
[4]: https://github.com/wking/ccon/blob/v0.4.0/test/t1001-process-user.t#L97

wking:
Rebased onto master with 01c92d6 → fdbcc47 now that #113 has landed (so GitHub will recognize this as a single-commit PR).

liangchenye:
Hi @wking , how does it work? It just adds maps to config.json, but a runtime (runc) still need 'root' to create/start the container.

wking:
On Mon, Jul 04, 2016 at 03:56:22AM -0700, 梁辰晔 (Liang Chenye) wrote:
> Hi @wking , how does it work? It just adds maps to config.json, but
> a runtime (runc) still need 'root' to create/start the container.

If runC still needs root, you can use ‘-r 'sudo runc'’ or call the
script with ‘test_runtime.sh’.  But ccon-oci [1] does not need root,
so it's nicer for me to not elevate privileges before running a test
suite that I haven't audited.

And my preferred approach is to have the runtime tests cover both
privileged and unprivileged callers [2], which would give me the
unprivileged tests I'm looking for (and skips for the privileged
tests) while keeping privileged tests for runC (with skips for the
unprivileged tests).  Folks that want to test for complete coverage
would want to run both the privileged and unprivileged tests.  When
testing ccon, I do this with two passes [3]:

  $ make
  …
  ok 4 # skip Test privileged process.user.uid (missing ROOT of ECHO,ID,ROOT)
  …
  fixed   0
  success 49
  failed  0
  broken  0
  total   58
  …
  $ sudo make
  …
  ok 7 # skip Test unprivileged process.user.uid (missing !ROOT of ECHO,GREP,ID,!ROOT)
  …
  fixed   0
  success 53
  failed  0
  broken  0
  total   58
  …

[1]: https://github.com/wking/ccon/tree/ccon-oci/examples/good/oci
[2]: https://github.com/opencontainers/ocitools/pull/114#issuecomment-226965729
[3]: https://github.com/wking/ccon/tree/v0.4.0/test

mrunalp:
This did not work with runc last time I tried. I'll try again tomorrow (still on PTO).

wking:
On Tue, Jul 05, 2016 at 08:29:08AM -0700, Mrunal Patel wrote:
> This did not work with runc last time I tried. I'll try again
> tomorrow (still on PTO).

Searching runC for open user namespace issues turned up
opencontainers/runc#252 and the setgroups call in
opencontainers/runc#774 and opencontainers/runc#814.  There's also the
gid=5 option in the default template's /dev/pts mount.

But stepping back, this is a good reason to have a test suite that
iterates over a number of configurations [1], so we can fail runC when
it doesn't support a particular use-case and pass it when it does, but
don't have to tailor the test suite to target what runC currently
supports.

[1]: https://github.com/opencontainers/ocitools/pull/114#issuecomment-226965729

wking:
On Tue, Jul 05, 2016 at 09:20:01AM -0700, W. Trevor King wrote:
> Tue, Jul 05, 2016 at 08:29:08AM -0700, Mrunal Patel:
> > This did not work with runc last time I tried. I'll try again
> > tomorrow (still on PTO).
> 
> Searching runC for open user namespace issues…

Also opencontainers/runc#799.

cyphar:
Yes, there are a couple of user namespace issues I've found in runC as a result of opencontainers/runc#774. They include:

* Everything to do with consoles.
* Namespace joining and creation issues.
* The default config (which is understandable) has `gid=5`.

I mentioned in the rootless containers PR that we really need to make our test suite do checks for user namespaces because quite a few things break that shouldn't. Problem is that you can't even run a single integration test case with the current console setup code (which is something that I mentioned in the last weekly call is something I'm working on fixing).

wking:
On Fri, Jul 22, 2016 at 11:30:56AM -0700, Aleksa Sarai wrote:
> Yes, there are a couple of user namespace issues I've found in runC…

And while fixing runC to work with user namespaces would be nice, I
think this PR is really a band-aid for testing without root.  I'd much
rather have a validation test suite that covered *both* root and
non-root invocations and user-namespace and non configs [1].  Do we
expect #61 to get picked up again any time soon?  Can I PR some bats
stuff cribbed from runC?  Or some sharness stuff cribbed from ccon?

[1]: https://github.com/opencontainers/ocitools/pull/114#issuecomment-230324293

mrunalp:
Yeah, I think we should make progress on #61. We can review it and request a rebase.

Mashimiao:
@opencontainers/runtime-tools-maintainers   should we close this PR?

wking:
On Wed, Nov 16, 2016 at 01:13:39AM -0800, Ma Shimiao wrote:
> should we close this PR?

I'm fine closing this PR, but would like more clarity on the path
forward.  #61 has not seen progress since May, and I haven't seen a
reason to expect development on it to pick back up in the near future.
I've offered to submit a bats or sharness runtime test suite [1] and
asked for clarification on resitance to those approaches [2].  But I
think we should have progress in one of those directions (#61 or a
bats/sharness runtime test suite) before we close this pain-point PR.

[1]: https://github.com/opencontainers/runtime-tools/pull/114#issuecomment-234622773
[2]: https://github.com/opencontainers/runtime-tools/pull/61#issuecomment-235343447

cyphar:
For an update, since the last time I posted here my rootless containers PR (opencontainers/runc#774) now passes all of the runC tests. So runC no longer has the same issue it did before (though you need to create a different config using `runc spec --rootless` for the same reasons described above -- configurations become invalid very easily with user namespaced containers where you only map a single user).

The only real issue I have with this PR is that it uses user namespaces for all tests. While this sounds like a no-op in the root-to-root case it actually isn't (you are messing with a lot of different aspects of the final environment). So there should be a flag to enable user namespace testing.

Also, as an aside runC's test suite also now works with regular user namespaces (which wasn't possible before without `--console-socket`) because of opencontainers/runc#1018 and opencontainers/runc#774.

wking:
On Wed, Nov 16, 2016 at 09:48:55AM -0800, Aleksa Sarai wrote:
> The only real issue I have with this PR is that it uses user
> namespaces for all tests…

There's no way to get this right when you only run one test container
;), which is why I'd like an actual test harness running multiple
tests (#61, or sharness like ccon, or bats like runC, …).

Mashimiao:
As test_runtime.sh is removing as planed, this PR will not be needed.

PR #116, opened 2016-06-20
validation: add args validation for process

Mashimiao:

### DIFF
--- a/validate.go
+++ b/validate.go
@@ -45,9 +45,9 @@ var (
 )
 
 var bundleValidateCommand = cli.Command{
-	Name:  "validate",
-	Usage: "validate a OCI bundle",
-	Flags: bundleValidateFlags,
+	Name:   "validate",
+	Usage:  "validate a OCI bundle",
+	Flags:  bundleValidateFlags,
 	Before: before,
 	Action: func(context *cli.Context) error {
 		inputPath := context.String("path")
@@ -168,6 +168,10 @@ func checkProcess(process rspec.Process, rootfs string) {
 		if !envValid(env) {
 			logrus.Fatalf("env %q should be in the form of 'key=value'. The left hand side must consist solely of letters, digits, and underscores '_'.", env)
 		}
+	}
+
+	if len(process.Args) < 1 {
+		logrus.Fatalf("args should not be empty")
 	}
 
 	for index := 0; index < len(process.Capabilities); index++ {


validate.go
```diff
@@ -170,6 +170,17 @@ func checkProcess(process rspec.Process, rootfs string) {
...
 	}
 
+	if len(process.Args) < 1 {
+		logrus.Fatalf("args should not be empty")
+	}
+	command := process.Args[0]
+	if filepath.IsAbs(command) {
+		cmdPath := path.Join(rootfs, command)
+		if _, err := os.Stat(cmdPath); err != nil {
+			logrus.Fatalf("Cannot find the command path %q", cmdPath)
```
wking: This check is too strict.  Perhaps the user intends to copy the executable into the right location before running `start` (see earlier discussion [here][1]).  Or maybe the user will bind-mount the command in from somewhere else.  Anyhow, I think it's to complicated to try and make this a validation-time check, and we should just leave it to the runtime/kernel to error out if the command isn't there at execve-time (or whatever syscall the runtime is triggering with ‘start’).

[1]: https://github.com/opencontainers/runc/pull/827#discussion_r63464201

wking: See also my notes [here][1] on the current spec phrasing being inconsistent.  The in-flight opencontainers/runtime-spec#427 should help make the specification more self-consistent.

[1]: https://github.com/wking/oci-runtime-config-validator/blob/cd0facf0ae1a6e6be7d8a20cac74b1299716b6b7/test/test_process.py#L134-L157

Mashimiao: @wking  Thanks for you information. I think you are right. I will remove code which checks whether command exists or not
From runtime-sepc, it seems args is required not optional. So I think checking whether args is empty is necessary.

Mashimiao: sorry, It seems no need to check args is empty or not. I will close this PR.

wking:
On Tue, Jun 21, 2016 at 06:47:39PM -0700, Ma Shimiao wrote:
> Closed #116.

Wait, what?  I was pushing back on the executable lookup [1], but I
think the current spec wording unambiguously requires at least one
entry in the process.args array [2], and after my suggested spec
changes that will still be the case [3].

[1]: https://github.com/opencontainers/ocitools/pull/116#discussion_r67768060
[2]: https://github.com/wking/oci-runtime-config-validator/blob/cd0facf0ae1a6e6be7d8a20cac74b1299716b6b7/test/test_process.py#L149-L150
[3]: https://github.com/opencontainers/runtime-spec/pull/427/files#diff-c9c91c29b41257aea3a3403cc606ad99R93

Mashimiao:
@wking  there is a function `checkMandatoryUnit` in validation.go can check args wheter should be empty or not. So, I think there is no need to check it again.
Am I right?

wking:
On Tue, Jun 21, 2016 at 09:58:49PM -0700, Ma Shimiao wrote:
> there is a function `checkMandatoryUnit` in validation.go can check
> args whether should be empty or not. So, I think there is no need to
> check it again.

Ah, you're right:

  $ ./ocitools generate
  $ CONFIG=$(jq 'del(.process.args)' config.json)
  $ echo "${CONFIG}" >config.json 
  $ ./ocitools validate
  FATA[0000] Mandatory information missing: ['Process.Args' should not be empty.].

So I agree that this PR wasn't needed.

PR #119, opened 2016-06-23
validate: add mount type check

Mashimiao:

### DIFF
--- a/man/ocitools-validate.1.md
+++ b/man/ocitools-validate.1.md
@@ -18,8 +18,15 @@ Validate an OCI bundle
 **--path=PATH
   Path to bundle
 
-**--hooks**
-  Check specified hooks exist and are executable on the host.
+**--host-specific**
+  Check host specific configs.
+  By default, validation only tests for compatibility with a hypothetical host.
+  With this flag, validation will also run more specific tests to see whether
+  the current host is capable of launching a container from the configuration.
+  For example, validating a compliant Windows configuration on a Linux machine
+  will pass without this flag ("there may be a Windows host capable of
+  launching this container"), but will fail with it ("this host is not capable
+  of launching this container").
 
 # SEE ALSO
 **ocitools**(1)
--- a/validate.go
+++ b/validate.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bufio"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -20,7 +21,7 @@ import (
 
 var bundleValidateFlags = []cli.Flag{
 	cli.StringFlag{Name: "path", Value: ".", Usage: "path to a bundle"},
-	cli.BoolFlag{Name: "hooks", Usage: "Check specified hooks exist and are executable on the host."},
+	cli.BoolFlag{Name: "host-specific", Usage: "Check host specific configs."},
 }
 
 var (
@@ -79,20 +80,21 @@ var bundleValidateCommand = cli.Command{
 			return fmt.Errorf("The root path %q is not a directory.", rootfsPath)
 		}
 
-		hooksCheck := context.Bool("hooks")
-		bundleValidate(spec, rootfsPath, hooksCheck)
+		hostCheck := context.Bool("host-specific")
+		bundleValidate(spec, rootfsPath, hostCheck)
 		logrus.Infof("Bundle validation succeeded.")
 		return nil
 	},
 }
 
-func bundleValidate(spec rspec.Spec, rootfs string, hooksCheck bool) {
+func bundleValidate(spec rspec.Spec, rootfs string, hostCheck bool) {
 	checkMandatoryField(spec)
 	checkSemVer(spec.Version)
 	checkPlatform(spec.Platform)
 	checkProcess(spec.Process, rootfs)
+	checkMounts(spec, hostCheck)
 	checkLinux(spec)
-	checkHooks(spec.Hooks, hooksCheck)
+	checkHooks(spec.Hooks, hostCheck)
 }
 
 func checkSemVer(version string) {
@@ -129,19 +131,19 @@ func checkPlatform(platform rspec.Platform) {
 	logrus.Fatalf("Operation system %q of the bundle is not supported yet.", platform.OS)
 }
 
-func checkHooks(hooks rspec.Hooks, hooksCheck bool) {
-	checkEventHooks("pre-start", hooks.Prestart, hooksCheck)
-	checkEventHooks("post-start", hooks.Poststart, hooksCheck)
-	checkEventHooks("post-stop", hooks.Poststop, hooksCheck)
+func checkHooks(hooks rspec.Hooks, hostCheck bool) {
+	checkEventHooks("pre-start", hooks.Prestart, hostCheck)
+	checkEventHooks("post-start", hooks.Poststart, hostCheck)
+	checkEventHooks("post-stop", hooks.Poststop, hostCheck)
 }
 
-func checkEventHooks(hookType string, hooks []rspec.Hook, hooksCheck bool) {
+func checkEventHooks(hookType string, hooks []rspec.Hook, hostCheck bool) {
 	for _, hook := range hooks {
 		if !filepath.IsAbs(hook.Path) {
 			logrus.Fatalf("The %s hook %v: is not absolute path", hookType, hook.Path)
 		}
 
-		if hooksCheck {
+		if hostCheck {
 			fi, err := os.Stat(hook.Path)
 			if err != nil {
 				logrus.Fatalf("Cannot find %s hook: %v", hookType, hook.Path)
@@ -188,6 +190,63 @@ func checkProcess(process rspec.Process, rootfs string) {
 		_, err := os.Stat(profilePath)
 		if err != nil {
 			logrus.Fatal(err)
+		}
+	}
+}
+
+func supportedMountTypes(OS string, hostCheck bool) (map[string]bool, error) {
+	supportedTypes := make(map[string]bool)
+
+	if OS != "linux" && OS != "windows" {
+		logrus.Warnf("%v is not supported to check mount type", OS)
+		return nil, nil
+	} else if OS == "windows" {
+		supportedTypes["ntfs"] = true
+		return supportedTypes, nil
+	}
+
+	if hostCheck {
+		f, err := os.Open("/proc/filesystems")
+		if err != nil {
+			return nil, err
+		}
+		defer f.Close()
+
+		s := bufio.NewScanner(f)
+		for s.Scan() {
+			if err := s.Err(); err != nil {
+				return supportedTypes, err
+			}
+
+			text := s.Text()
+			parts := strings.Split(text, "\t")
+			if len(parts) > 1 {
+				supportedTypes[parts[1]] = true
+			} else {
+				supportedTypes[parts[0]] = true
+			}
+		}
+
+		supportedTypes["bind"] = true
+
+		return supportedTypes, nil
+	} else {
+		logrus.Warn("Checking linux mount types without --host-specific is not supported yet")
+		return nil, nil
+	}
+}
+
+func checkMounts(spec rspec.Spec, hostCheck bool) {
+	supportedTypes, err := supportedMountTypes(spec.Platform.OS, hostCheck)
+	if err != nil {
+		logrus.Fatal(err)
+	}
+
+	if supportedTypes != nil {
+		for _, mount := range spec.Mounts {
+			if !supportedTypes[mount.Type] {
+				logrus.Fatalf("Unsupported mount type %q", mount.Type)
+			}
 		}
 	}
 }
--- a/completions/bash/ocitools
+++ b/completions/bash/ocitools
@@ -253,7 +253,7 @@ _ocitools_validate() {
 
 	case "$cur" in
 		-*)
-			COMPREPLY=( $( compgen -W "--hooks --path --help" -- "$cur" ) )
+			COMPREPLY=( $( compgen -W "--host-specific --path --help" -- "$cur" ) )
 			;;
 	esac
 


validate.go
```diff
@@ -192,6 +194,48 @@ func checkProcess(process rspec.Process, rootfs string) {
...
+		text := s.Text()
+		parts := strings.Split(text, "\t")
+		if len(parts) > 1 {
+			supportedTypes[parts[1]] = true
+		} else {
+			supportedTypes[parts[0]] = true
+		}
+	}
+
+	supportedTypes["nfs"] = true
```
wking: I see an `nfs` entry in my `/proc/filesystems` (and an `nfs4` entry).  Why are you adding it by hand here?

validate.go
```diff
@@ -192,6 +194,48 @@ func checkProcess(process rspec.Process, rootfs string) {
...
+	supportedTypes["bind"] = true
```
wking: While I agree that bind mounts are important, I don't see any wording in [the current spec][1] about them.  We probably want to file a PR adding them as an additional “type” (and maybe add other wording about bind(8) options, e.g. the shared/slave/private/unbindable stuff?) before allowing them here.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/config.md#mounts

validate.go
```diff
@@ -192,6 +194,48 @@ func checkProcess(process rspec.Process, rootfs string) {
 	}
 }
 
+func supportedMountTypes() (map[string]bool, error) {
+	supportedTypes := make(map[string]bool)
+	f, err := os.Open("/proc/filesystems")
```
wking: This probably needs to go behind a [`--host-checks`][1] (or [`--host-specific`][2] or some such) flag, since reading the local system's `/proc/filesystems` [is not portable][3].

[2]: https://github.com/opencontainers/ocitools/pull/49#issuecomment-220534023
[3]: https://github.com/opencontainers/ocitools/pull/49#issuecomment-217793768

validate.go
Mashimiao: Ah, it seems it's really no need to add.

validate.go
Mashimiao: In runtime-spec's mount example, I see `{
        "destination": "/data",
        "type": "bind",
        "source": "/volumes/testing",
        "options": ["rbind","rw"]
    }`, so I think bind may must be supported.

wking: On Thu, Jun 23, 2016 at 06:10:01PM -0700, Ma Shimiao wrote:
> In runtime-spec's mount example, I see `{
>         "destination": "/data",
>         "type": "bind",
>         "source": "/volumes/testing",
>         "options": ["rbind","rw"]
>     }`, so I think bind may must be supported.

I'd rather have it spelled out in non-example words before we land
tests for it here, although I agree that landing non-example words
shouldn't be too contentious.

validate.go
Mashimiao: I think this is more related to container's config, So I added a new commit to split linux-sepcific validations.

Mashimiao: @wking  please ignore the above comment.

Mashimiao: On 06/24/2016 01:41 AM, W. Trevor King wrote:
>> +	f, err := os.Open("/proc/filesystems")
> 
> This probably needs to go behind a |--host-checks| <https://github.com/opencontainers/ocitools/pull/49#issuecomment-216585596> (or |--host-specific| <https://github.com/opencontainers/ocitools/pull/49#issuecomment-220534023> or some such) flag, since reading the local system's |/proc/filesystems| is not portable <https://github.com/opencontainers/ocitools/pull/49#issuecomment-217793768>.
> 
Thanks for your suggestion.
I think using additional options like --host-checks or --host-specific is an option.

But currently supportedMountTypes() just supports Linux and I'm not sure does other systems has file like /proc/filesystems to
list system supported file systems? 
If user validated with --host-checks and used host with OS which is not Linux, the validation would still failed.
User may not want to get this result.

How about check host's OS first before executing checkMounts()?
If the host is Linux, then execute checkMounts(). If not, skip checkMounts().


-- 
Ma Shimiao
Development Dept.I

wking: On Thu, Jun 23, 2016 at 07:55:48PM -0700, Ma Shimiao wrote:
> But currently supportedMountTypes() just supports Linux and I'm not
> sure does other systems has file like /proc/filesystems to list
> system supported file systems?

The problem with /proc/filesystems is partly that only a Linux host
will have it, but also that the contents depend on the particular
kernel.  For example, I have ‘nfs4’ in my /proc/filesystems because my
kernel was compiled with CONFIG_NFS_V4=y.  My kernel was not compiled
with CONFIG_SQUASHFS, so I don't have ‘squashfs’ in my
/proc/filesystems.  So this, like whether or not a hook command is
present [1], is a check that may pass on some hosts but not others.
It's important if you're about to run that config on that host, but
less useful if you're running a validation service or some such.

So I think we want:

* A --host-specific flag that skips this test (at least on Linux,
  where the test is based on a per-kernel /proc/filesystems).
* A platform.os check, because we don't want to be comparing mounts
  against /proc/filesystems if the container targets Windows.
* A skip if platform.os is ‘linux’, but we can't open
  /proc/filesystems (e.g. because we're running the validator on a
  Windows host).

wking:
Also related to this and the ‘bind’ type is that mount.type might
become optional (opencontainers/runtime-spec#470).  In ccon, bind
mounts are handled with explict MS_BIND (and MS_REC) flags [1].

[1]: https://github.com/wking/ccon/tree/v0.4.0#mount-namespace

validate.go
Mashimiao: On 06/24/2016 12:12 PM, W. Trevor King wrote:
> So I think we want: * A --host-specific flag that skips this test (at least on Linux, where the test is based on a per-kernel /proc/filesystems). * A platform.os check, because we don't want to be comparing mounts against /proc/filesystems if the container targets Windows. * A skip if platform.os is ‘linux’, but we can't open /proc/filesystems (e.g. because we're running the validator on a Windows host). [1]: #49 (comment) <https://github.com/opencontainers/ocitools/pull/49#issuecomment-216585596>
fixed as your suggestions.

-- 
Ma Shimiao
Development Dept.I

man/ocitools-validate.1.md
```diff
@@ -21,6 +21,9 @@ Validate an OCI bundle
 **--hooks**
   Check specified hooks exist and are executable on the host.
 
+**--host-specific**
+  Check host specified configs.
```
wking: “specified” → “specific”.  And I'd add more context like:

> By default, validation only tests for compatibility with a hypothetical host.  With this flag, validation will also run more specific tests to see whether the current host is capable of launching a container from the configuration.  For example, validating a compliant Windows configuration on a Linux machine will pass without this flag ("there may be a Windows host capable of launching this container"), but will fail with it ("this host is not capable of launching this container").

validate.go
```diff
@@ -21,6 +23,7 @@ import (
 var bundleValidateFlags = []cli.Flag{
 	cli.StringFlag{Name: "path", Value: ".", Usage: "path to a bundle"},
 	cli.BoolFlag{Name: "hooks", Usage: "Check specified hooks exist and are executable on the host."},
+	cli.BoolFlag{Name: "hook-specific", Usage: "Check host specified configs."},
```
wking: “hook-specific” → “host-specific”.  And I think the new flag should *replace* `--hooks`, not be an addition to it (because the reason we disable the “do you have that hook on *this* host?” check is [because it is host-specific][1]).  Is there a reason to keep the hook flag as an independent knob?

wking: Also “specified” → “specific”.

validate.go
```diff
@@ -80,17 +83,21 @@ var bundleValidateCommand = cli.Command{
...
 		logrus.Infof("Bundle validation succeeded.")
 		return nil
 	},
 }
 
-func bundleValidate(spec rspec.Spec, rootfs string, hooksCheck bool) {
+func bundleValidate(spec rspec.Spec, rootfs string, hooksCheck bool, hostCheck bool) {
 	checkMandatoryField(spec)
 	checkSemVer(spec.Version)
 	checkPlatform(spec.Platform)
```
wking: This can probably also get a `hostCheck` that compares `spec.Platform` against the current host (`runtime.GOOS`?).

validate.go
```diff
@@ -80,17 +83,21 @@ var bundleValidateCommand = cli.Command{
...
 	checkProcess(spec.Process, rootfs)
+	if hostCheck && spec.Platform.OS == "linux" && runtime.GOOS == "linux" {
+		checkMounts(spec.Mounts)
```
wking: Mounts are a thing regardless of the Linux-ness of the config/host.  I'd rather have this always execute, pass the full `spec` and `hostChec` through to `supportedTypes`, and then have `supportedMountTypes` raise a skip/warning if it was asked for something it couldn't provide yet:

* `spec.Platform.OS == "windows"` → `["ntfs"], nil`
* `spec.Platform.OS == "linux" && hostCheck` → `[…stuff from /proc/filesystems…], nil`
* `spec.Platform.OS == "linux" && !hostCheck` → `nil, Skip(I don't have a generic list of Linux filesystem types)`
* `spec.Platform.OS != "linux" && spec.Platform.OS != "windows"` → `nil, Skip(I don't have a list of {OS} filesystem types)`

See opencontainers/runtime-spec#470 about `ntfs` being underspecified for Windows, but that's what the spec currently has.

validate.go
Mashimiao: fixed. Thx.

validate.go
Mashimiao: I think Paltform check has nothing to do with hostCheck

validate.go
Mashimiao: fixed mostly as you suggested. Thanks.

validate.go
wking: On Mon, Jun 27, 2016 at 12:00:12AM -0700, Ma Shimiao wrote:
> >  	checkPlatform(spec.Platform)
> 
> I think Paltform check has nothing to do with hostCheck

The spec has [1]:

  The runtime MUST generate an error if it does not support the
  configured os.

I guess a VM-based runtime could support launching a Windows bundle
from a Linux host (etc.), but your /proc/filesystems approach here is
assuming runtimes are using a container approach, and not a VM
approach.  With a container approach, a runtime based on the host's
Linux kernel is not going to be able to support a Windows container.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/config.md#platform

Mashimiao: I'm not sure what spec means. I think that's runtime's business. checkPlatform seems can't judge you mentioned above.
Or could you give me more detailed information about how to modify checkPlatform?
Thanks a lot.

man/ocitools-validate.1.md
```diff
@@ -18,8 +18,15 @@ Validate an OCI bundle
...
-  Check specified hooks exist and are executable on the host.
+**--host-specific**
+  Check host specific configs.
+  By default, validation only tests for compatibility with a hypothetical host. 
+  With this flag, validation will also run more specific tests to see whether 
+  the current host is capable of launching a container from the configuration. 
+  For example, validating a compliant Windows configuration on a Linux machine 
+  will pass without this flag ("there may be a Windows host capable of 
+  launching this container"), but will fail with it ("this host is not capable 
+  of launching this container").
```
wking: nit: there's trailing whitespace here (in case we care in this repository; I don't if we do).

validate.go
```diff
@@ -192,6 +194,62 @@ func checkProcess(process rspec.Process, rootfs string) {
...
+			} else {
+				supportedTypes[parts[0]] = true
+			}
+		}
+
+		supportedTypes["bind"] = true
+
+		return supportedTypes, nil
+	} else {
+		return nil, nil
```
wking: nit: you don't need an else.  If you get this far, you didn't match the `if` condition.

You probably want a warn here, with something like “Linux mount checks without --host-specific are not implemented yet”.

validate.go
wking: On Mon, Jun 27, 2016 at 07:27:17AM -0700, Ma Shimiao wrote:
> >  	checkPlatform(spec.Platform)
> 
> I'm not sure what spec means. I think that's runtime's
> business. checkPlatform seems can't judge you mentioned above.  Or
> could you give me more detailed information about how to modify
> checkPlatform?

Let's just punt this to a follow-up PR.

wking:
32803b5c looks good to me, excepting a possible whitespace nit [1] and
a warning for Linux configs when --host-specific is not set [2].  I'm
fine punting checkPlatform changes to a follow-up PR [3].

[1]: https://github.com/opencontainers/ocitools/pull/119#discussion_r68609432
[2]: https://github.com/opencontainers/ocitools/pull/119#discussion_r68609880
[3]: https://github.com/opencontainers/ocitools/pull/119#discussion_r68610091

validate.go
Mashimiao: fixed. Thanks.

man/ocitools-validate.1.md
Mashimiao: Fixed. Thanks.

Mashimiao:
ping @liangchenye , @mrunalp

liangchenye:
LGTM
the  `--host-specific` looks great to me :+1:

validate.go
```diff
@@ -192,6 +194,63 @@ func checkProcess(process rspec.Process, rootfs string) {
...
+	} else if OS == "windows" {
+		supportedTypes["ntfs"] = true
+		return supportedTypes, nil
+	}
+
+	if hostCheck {
+		f, err := os.Open("/proc/filesystems")
+		if err != nil {
+			return supportedTypes, err
```
wking: This should probably return `nil, err`.  There's not much point in returning an empty map, and the caller will bail when it sees the error anyway.

Mashimiao: On 06/28/2016 11:21 AM, W. Trevor King wrote:
> This should probably return |nil, err|. There's not much point in returning an empty map, and the caller will bail when it sees the error anyway.
fixed. Thanks.

-- 
Ma Shimiao
Development Dept.I

wking:
Even with my pedantic hat on, I don't see anything else to polish with
f1ff4d09 ;).  Looks good to me.

Mashimiao:
ping @mrunalp

Mashimiao:
ping @mrunalp

mrunalp:
LGTM

PR #120, opened 2016-06-23
checkMandatory: fix slice check

Mashimiao:

When json sets null to slice type and does not have omitempty tag, the check should failed.

### DIFF
--- a/validate.go
+++ b/validate.go
@@ -383,7 +383,7 @@ func checkMandatoryUnit(field reflect.Value, tagField reflect.StructField, paren
 	mandatory := !strings.Contains(tagField.Tag.Get("json"), "omitempty")
 	switch field.Kind() {
 	case reflect.Ptr:
-		if mandatory && field.IsNil() == true {
+		if mandatory && field.IsNil() {
 			msgs = append(msgs, fmt.Sprintf("'%s.%s' should not be empty.", parent, tagField.Name))
 			return msgs, false
 		}
@@ -393,7 +393,7 @@ func checkMandatoryUnit(field reflect.Value, tagField reflect.StructField, paren
 			return msgs, false
 		}
 	case reflect.Slice:
-		if mandatory && (field.Len() == 0) {
+		if mandatory && (field.IsNil() || field.Len() == 0) {
 			msgs = append(msgs, fmt.Sprintf("'%s.%s' should not be empty.", parent, tagField.Name))
 			return msgs, false
 		}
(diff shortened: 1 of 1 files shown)


validate.go
```diff
@@ -393,7 +393,7 @@ func checkMandatoryUnit(field reflect.Value, tagField reflect.StructField, paren
 			return msgs, false
 		}
 	case reflect.Slice:
-		if mandatory && (field.Len() == 0) {
+		if mandatory && (field.IsNil() || (field.Len() == 0)) {
```
wking: You [don't need those parens][1] around the `Len == 0` bit.

The `IsNil` addition looks ok, but the docs say that it differs from `== nil` by [being more panicky][2], so I think I'd prefer `== nil`.

[1]: https://golang.org/ref/spec#Operator_precedence
[2]: https://golang.org/pkg/reflect/#Value.IsNil

Mashimiao: On 06/24/2016 01:26 AM, W. Trevor King wrote:
> In validate.go <https://github.com/opencontainers/ocitools/pull/120#discussion_r68277602>:
> 
>> @@ -393,7 +393,7 @@ func checkMandatoryUnit(field reflect.Value, tagField reflect.StructField, paren
>>  			return msgs, false
>>  		}
>>  	case reflect.Slice:
>> -		if mandatory && (field.Len() == 0) {
>> +		if mandatory && (field.IsNil() || (field.Len() == 0)) {
> 
> You don't need those parens <https://golang.org/ref/spec#Operator_precedence> around the |Len == 0| bit.
Fine.
> 
> The |IsNil| addition looks ok, but the docs say that it differs from |== nil| by being more panicky <https://golang.org/pkg/reflect/#Value.IsNil>, so I think I'd prefer |== nil|.
> 
Yes, IsNil is more panicky, but I think used here is ok.
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub <https://github.com/opencontainers/ocitools/pull/120/files/318320bd055e8b6a9617236de88947460ab7f1fa#r68277602>, or mute the thread <https://github.com/notifications/unsubscribe/AEoKADtngjQ22kjTgg3ouQVmzUrlJ3lMks5qOsHTgaJpZM4I85sf>.Web Bug from https://github.com/notifications/beacon/AEoKAIXccgBR3PAwUVX-0j3jaWkOZkZHks5qOsHTgaJpZM4I85sf.gif
> 


-- 
Ma Shimiao
Development Dept.I

mrunalp:
LGTM. @liangchenye PTAL.

liangchenye:
LGTM

PR #124, opened 2016-06-27
validate: add sysctl check

Mashimiao:

### DIFF
--- a/validate.go
+++ b/validate.go
@@ -195,6 +195,9 @@ func checkProcess(process rspec.Process, rootfs string) {
 //Linux only
 func checkLinux(spec rspec.Spec) {
 	utsExists := false
+	ipcExists := false
+	mountExists := false
+	netExists := false
 
 	if len(spec.Linux.UIDMappings) > 5 {
 		logrus.Fatalf("Only 5 UID mappings are allowed (linux kernel restriction).")
@@ -206,8 +209,27 @@ func checkLinux(spec rspec.Spec) {
 	for index := 0; index < len(spec.Linux.Namespaces); index++ {
 		if !namespaceValid(spec.Linux.Namespaces[index]) {
 			logrus.Fatalf("namespace %v is invalid.", spec.Linux.Namespaces[index])
-		} else if spec.Linux.Namespaces[index].Type == rspec.UTSNamespace {
-			utsExists = true
+		} else if len(spec.Linux.Namespaces[index].Path) == 0 {
+			if spec.Linux.Namespaces[index].Type == rspec.UTSNamespace {
+				utsExists = true
+			} else if spec.Linux.Namespaces[index].Type == rspec.IPCNamespace {
+				ipcExists = true
+			} else if spec.Linux.Namespaces[index].Type == rspec.NetworkNamespace {
+				netExists = true
+			} else if spec.Linux.Namespaces[index].Type == rspec.MountNamespace {
+				mountExists = true
+			}
+		}
+	}
+
+	for k := range spec.Linux.Sysctl {
+		if strings.HasPrefix(k, "net.") && !netExists {
+			logrus.Fatalf("Sysctl %v requires a new Network namespace to be specified as well", k)
+		}
+		if strings.HasPrefix(k, "fs.mqueue.") {
+			if !mountExists || !ipcExists {
+				logrus.Fatalf("Sysctl %v requires a new IPC namespace and Mount namespace to be specified as well", k)
+			}
 		}
 	}
 


validate.go
```diff
@@ -208,6 +210,19 @@ func checkLinux(spec rspec.Spec) {
...
+			netExists = true
+		}
+	}
+
+	for k, _ := range spec.Linux.Sysctl {
+		if strings.HasPrefix(k, "net.") && !netExists {
+			logrus.Fatalf("Sysctl %v requires a new Network namespace to be specified as well", k)
+		}
+		if strings.HasPrefix(k, "fs.mqueue.") && !ipcExists {
+			logrus.Fatalf("Sysctl %v requires a new IPC namespace to be specified as well", k)
```
wking: Does this also depend on having a `/dev/mqueue` that is specific to your IPC namespace?  That would require a new mount namespace as well (see testing recommendations [here][1]).  And `/dev/mqueue` is not one of the [required filesystems][2], but testing for “will this config supply a new `/dev/mqueue`?” is getting a bit dicey (e.g. maybe the user intends to supply it on their own after `create` and before `start`).  So I think we may want to require a new mount namespace here, but not bother about checking for a new `/dev/mqueue`.

[1]: http://thread.gmane.org/gmane.linux.kernel/782342
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/config-linux.md#default-file-systems

Mashimiao: add new mount namespace detection is OK. will add it.

Mashimiao: add new mount namespace detection is OK. will adding it.

validate.go
```diff
@@ -208,6 +211,23 @@ func checkLinux(spec rspec.Spec) {
 			logrus.Fatalf("namespace %v is invalid.", spec.Linux.Namespaces[index])
 		} else if spec.Linux.Namespaces[index].Type == rspec.UTSNamespace {
 			utsExists = true
+		} else if spec.Linux.Namespaces[index].Type == rspec.IPCNamespace {
+			ipcExists = true
```
wking: These checks should probably skip namespaces where `path` is set (so they only pick up *new* namespaces).  If tweaking existing namespaces is fine, there's no need for the “does the config declare a namespace” checks.  And the current spec [explicitly forbids tweaking non-new namespaces][1]:

> Also, when a path is specified, a runtime MUST assume that the setup for that particular namespace has already been done and error out if the config specifies anything else related to that namespace.

Although [I'm happy to revisit that restriction][2].

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/config-linux.md#namespaces
[2]: https://github.com/opencontainers/runtime-spec/pull/158#issuecomment-138997548

Mashimiao: I don't think we should skip namespace where `path` is set.
That's true spec says `runtime must assume path specified namespace has already been done`.
But the target of `validate` is to help user to write a complete config file and all items can be applied correctly.
Some items in config file need to work with some namespace.
If some namespaces did not exist, those items would not make sense.
So check path specified namespace declared or not is also the business of `validate`, we should not skip it.

wking: On Mon, Jun 27, 2016 at 11:58:28PM -0700, Ma Shimiao wrote:
> Some items in config file need to work with some namespace.  If some
> namespaces did not exist, those items would not make sense.

The container process will always be in *some* namespace (for each
type).  Without a namespace entry, it will be in the runtime namespace
[1].  With a namespace entry that sets ‘path’, it will join some
existing namespace.  With a namespace entry that does not set ‘path’,
it will create a new namespace.  The current spec wording requires a
new namespace “if the config specifies anything else related to that
namespace”.  I think that means you need an IPC namespace without
‘path’ and a mount namespace without ‘path’ before you can set an
‘fs.mqueue.…’ sysctl.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/glossary.md#runtime-namespace

Mashimiao: Got it. Thanks, fixed.

Mashimiao:
ping @mrunalp  @liangchenye

validate.go
```diff
@@ -208,6 +211,23 @@ func checkLinux(spec rspec.Spec) {
 			logrus.Fatalf("namespace %v is invalid.", spec.Linux.Namespaces[index])
 		} else if spec.Linux.Namespaces[index].Type == rspec.UTSNamespace {
 			utsExists = true
+		} else if spec.Linux.Namespaces[index].Type == rspec.IPCNamespace && len(spec.Linux.Namespaces[index].Path) <= 0 {
```
initlove: I suggest to use `len(***Path)  == 0` in all these empty string checks.

liangchenye: I suggest to use len(***Path) == 0 in all these empty string checks.

liangchenye:
2c32a42 LGTM

validate.go
```diff
@@ -208,6 +211,23 @@ func checkLinux(spec rspec.Spec) {
 			logrus.Fatalf("namespace %v is invalid.", spec.Linux.Namespaces[index])
 		} else if spec.Linux.Namespaces[index].Type == rspec.UTSNamespace {
 			utsExists = true
```
wking: I think we want:

    else if len(spec.Linux.Namespaces[index].Path) == 0 {
        continue
    }

as the first block after the validity check.  That way [the logic][1] also applies to this UTS case, and we don't have to duplicate `Path` checks for each namespace below.

[1]: https://github.com/opencontainers/ocitools/pull/124#discussion_r68693222

Mashimiao: On 07/06/2016 11:39 AM, W. Trevor King wrote:
> I think we want:
> 
> |else if len(spec.Linux.Namespaces[index].Path) == 0 { continue } |
> 
> as the first block after the validity check. That way the logic <https://github.com/opencontainers/ocitools/pull/124#discussion_r68693222> also applies to this UTS case, and we don't have to duplicate |Path| checks for each namespace below.
> 

Yes, it really is. will fix.

-- 
Ma Shimiao
Development Dept.I

wking:
a9ae71a3 looks good to me :).

Mashimiao:
ping @mrunalp

mrunalp:
LGTM

PR #127, opened 2016-07-01
validation: add linux default devices validation

Mashimiao:

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -29,6 +29,17 @@ var (
 		"/dev/pts": "devpts",
 		"/dev/shm": "tmpfs",
 	}
+
+	defaultDevices = []string{
+		"/dev/null",
+		"/dev/zero",
+		"/dev/full",
+		"/dev/random",
+		"/dev/urandom",
+		"/dev/tty",
+		"/dev/console",
+		"/dev/ptmx",
+	}
 )
 
 type validation func(*rspec.Spec) error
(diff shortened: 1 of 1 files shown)


cmd/runtimetest/main.go
```diff
@@ -260,6 +271,27 @@ func validateDefaultFS(spec *rspec.Spec) error {
...
+		fi, err := os.Stat(device)
+		if err != nil {
+			if os.IsNotExist(err) {
+				return fmt.Errorf("Device %v must exist but not", device)
+			} else {
+				return err
+			}
+		} else {
+			if fi.Mode() != os.ModeDevice {
+				return fmt.Errorf("file %v expected is a device but not", device)
```
wking: This looks like a reasonable first level of check.  It would be nice to confirm that the device major/minors also match the [spec requirements], but I'm ok punting that to a follow-up PR.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/config-linux.md#default-devices

wking:
I'd like to test the major/minor eventually, but eb147670 looks like a
step in the right direction to me.

cmd/runtimetest/main.go
Mashimiao: On 07/01/2016 01:45 PM, W. Trevor King wrote:
> This looks like a reasonable first level of check. It would be nice to confirm that the device major/minors also match the [spec requirements], but I'm ok punting that to a follow-up PR.
I'm also OK to punting that to a follow-up PR. :-)

-- 
Ma Shimiao
Development Dept.I

mrunalp: file %v is not a device node as expected

cmd/runtimetest/main.go
```diff
@@ -260,6 +271,27 @@ func validateDefaultFS(spec *rspec.Spec) error {
...
 }
 
+func validateDefaultDevices(spec *rspec.Spec) error {
+	logrus.Debugf("validating linux default devices")
+
+	for _, device := range defaultDevices {
+		fi, err := os.Stat(device)
+		if err != nil {
+			if os.IsNotExist(err) {
+				return fmt.Errorf("Device %v must exist but not", device)
```
mrunalp: device node %v not found

cmd/runtimetest/main.go
Mashimiao: fixed. Thanks.

cmd/runtimetest/main.go
Mashimiao: fixed, thanks.

Mashimiao:
ping @mrunalp

mrunalp:
LGTM

PR #122, opened 2016-06-24
generate: add new cgroup namespace option

Mashimiao:
Signed-off-by: Ma Shimiao <mashimiao.fnst@cn.fujitsu.com>

PR #128, opened 2016-07-01
Use oci-seccomp-gen library for adding seccomp configurations

GrantSeltzer:
I added functionality to specify syscalls and their arguments for all actions. Also changed default action and archs flags to ociseccompgen library. 

Now we can do things like:

`ocitools generate --seccomp-errno clone,write --seccomp-trace getcwd --seccomp-trap umount:0:1:2:NE`

The behavior mimics that of: https://github.com/GrantSeltzer/Manhattan

Also updated man page to reflect this change.

### DIFF
--- a/cmd/ocitools/generate.go
+++ b/cmd/ocitools/generate.go
@@ -3,6 +3,7 @@ package main
 import (
 	"os"
 	"runtime"
+	"strings"
 
 	"github.com/opencontainers/ocitools/generate"
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
@@ -48,11 +49,14 @@ var generateFlags = []cli.Flag{
 	cli.StringSliceFlag{Name: "gidmappings", Usage: "add GIDMappings e.g HostID:ContainerID:Size"},
 	cli.StringSliceFlag{Name: "sysctl", Usage: "add sysctl settings e.g net.ipv4.forward=1"},
 	cli.StringFlag{Name: "apparmor", Usage: "specifies the the apparmor profile for the container"},
-	cli.StringFlag{Name: "seccomp-default", Usage: "specifies the the defaultaction of Seccomp syscall restrictions"},
-	cli.StringSliceFlag{Name: "seccomp-arch", Usage: "specifies Additional architectures permitted to be used for system calls"},
-	cli.StringSliceFlag{Name: "seccomp-syscalls", Usage: "specifies Additional architectures permitted to be used for system calls, e.g Name:Action:Arg1_index/Arg1_value/Arg1_valuetwo/Arg1_op, Arg2_index/Arg2_value/Arg2_valuetwo/Arg2_op "},
-	cli.StringSliceFlag{Name: "seccomp-allow", Usage: "specifies syscalls to be added to allowed"},
-	cli.StringSliceFlag{Name: "seccomp-errno", Usage: "specifies syscalls to be added to list that returns an error"},
+	cli.BoolFlag{Name: "seccomp-only", Usage: "specifies to export just a seccomp configuration file"},
+	cli.StringFlag{Name: "seccomp-arch", Usage: "specifies additional architectures permitted to be used for system calls"},
+	cli.StringFlag{Name: "seccomp-default", Usage: "specifies default action to be used for system calls"},
+	cli.StringFlag{Name: "seccomp-allow", Usage: "specifies syscalls to respond with allow"},
+	cli.StringFlag{Name: "seccomp-trap", Usage: "specifies syscalls to respond with trap"},
+	cli.StringFlag{Name: "seccomp-errno", Usage: "specifies syscalls to respond with errno"},
+	cli.StringFlag{Name: "seccomp-trace", Usage: "specifies syscalls to respond with trace"},
+	cli.StringFlag{Name: "seccomp-kill", Usage: "specifies syscalls to respond with kill"},
 	cli.StringFlag{Name: "template", Usage: "base template to use for creating the configuration"},
 	cli.StringSliceFlag{Name: "label", Usage: "add annotations to the configuration e.g. key=value"},
 }
@@ -82,13 +86,15 @@ var generateCommand = cli.Command{
 		if err != nil {
 			return err
 		}
+		var exportOpts generate.ExportOptions
+		exportOpts.Seccomp = context.Bool("seccomp-only")
 
 		if context.IsSet("output") {
-			output := context.String("output")
-			err = specgen.SaveToFile(output)
+			err = specgen.SaveToFile(context.String("output"), exportOpts)
 		} else {
-			err = specgen.Save(os.Stdout)
+			err = specgen.Save(os.Stdout, exportOpts)
 		}
+
 		if err != nil {
 			return err
 		}
@@ -301,65 +307,85 @@ func setupSpec(g generate.Generator, context *cli.Context) error {
 		}
 	}
 
-	var sd string
-	var sa, ss []string
-
-	if context.IsSet("seccomp-default") {
-		sd = context.String("seccomp-default")
-	}
-
-	if context.IsSet("seccomp-arch") {
-		sa = context.StringSlice("seccomp-arch")
-	}
-
-	if context.IsSet("seccomp-syscalls") {
-		ss = context.StringSlice("seccomp-syscalls")
-	}
-
-	if sd == "" && len(sa) == 0 && len(ss) == 0 {
-		return nil
-	}
-
-	// Set the DefaultAction of seccomp
-	if context.IsSet("seccomp-default") {
-		if err := g.SetLinuxSeccompDefault(sd); err != nil {
-			return err
-		}
-	}
-
-	// Add the additional architectures permitted to be used for system calls
-	if context.IsSet("seccomp-arch") {
-		for _, arch := range sa {
-			if err := g.AddLinuxSeccompArch(arch); err != nil {
-				return err
-			}
-		}
-	}
-
-	// Set syscall restrict in Seccomp
-	if context.IsSet("seccomp-syscalls") {
-		for _, syscall := range ss {
-			if err := g.AddLinuxSeccompSyscall(syscall); err != nil {
-				return err
-			}
-		}
-	}
-
-	if context.IsSet("seccomp-allow") {
-		seccompAllows := context.StringSlice("seccomp-allow")
-		for _, s := range seccompAllows {
-			g.AddLinuxSeccompSyscallAllow(s)
-		}
-	}
-
-	if context.IsSet("seccomp-errno") {
-		seccompErrnos := context.StringSlice("seccomp-errno")
-		for _, s := range seccompErrnos {
-			g.AddLinuxSeccompSyscallErrno(s)
-		}
+	g, err := addSeccomp(g, context)
+	if err != nil {
+		return err
 	}
 
 	return nil
+}
+
+func addSeccomp(g generate.Generator, context *cli.Context) (generate.Generator, error) {
+	seccompDefault := context.String("seccomp-default")
+	seccompArch := context.String("seccomp-arch")
+	seccompKill := context.String("seccomp-kill")
+	seccompTrace := context.String("seccomp-trace")
+	seccompErrno := context.String("seccomp-errno")
+	seccompTrap := context.String("seccomp-trap")
+	seccompAllow := context.String("seccomp-allow")
+
+	// Set the DefaultAction of seccomp
+	if seccompDefault == "" {
+		seccompDefault = "errno"
+	}
+
+	err := g.SetDefaultSeccompAction(seccompDefault)
+	if err != nil {
+		return g, err
+	}
+
+	// Add the additional architectures permitted to be used for system calls
+	if seccompArch == "" {
+		seccompArch = "amd64,x86,x32" // Default Architectures
+	}
+
+	architectureArgs := strings.Split(seccompArch, ",")
+	err = g.SetSeccompArchitectures(architectureArgs)
+	if err != nil {
+		return g, err
+	}
+
+	if seccompKill != "" {
+		killArgs := strings.Split(seccompKill, ",")
+		err = g.SetSyscallActions("kill", killArgs)
+		if err != nil {
+			return g, err
+		}
+	}
+
+	if seccompTrace != "" {
+		traceArgs := strings.Split(seccompTrace, ",")
+		err = g.SetSyscallActions("trace", traceArgs)
+		if err != nil {
+			return g, err
+		}
+	}
+
+	if seccompErrno != "" {
+		errnoArgs := strings.Split(seccompErrno, ",")
+		err = g.SetSyscallActions("errno", errnoArgs)
+		if err != nil {
+			return g, err
+		}
+	}
+
+	if seccompTrap != "" {
+		trapArgs := strings.Split(seccompTrap, ",")
+		err = g.SetSyscallActions("trap", trapArgs)
+		if err != nil {
+			return g, err
+		}
+	}
+
+	if seccompAllow != "" {
+		allowArgs := strings.Split(seccompAllow, ",")
+		err = g.SetSyscallActions("allow", allowArgs)
+		if err != nil {
+			return g, err
+		}
+	}
+
+	return g, nil
 }
 
 func checkNs(nsMaps map[string]string, nsName string) bool {
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -10,6 +10,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/opencontainers/ocitools/generate/seccomp"
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/syndtr/gocapability/capability"
 )
@@ -22,6 +23,11 @@ var (
 // Generator represents a generator for a container spec.
 type Generator struct {
 	spec *rspec.Spec
+}
+
+// ExportOptions has toggles for exporting only certain parts of the specification
+type ExportOptions struct {
+	Seccomp bool // seccomp toggles if only seccomp should be exported
 }
 
 // New creates a spec Generator with the default spec.
@@ -136,6 +142,7 @@ func New() Generator {
 					Type: "mount",
 				},
 			},
+			Seccomp: &rspec.Seccomp{},
 			Devices: []rspec.Device{},
 		},
 	}
@@ -180,8 +187,14 @@ func (g *Generator) GetSpec() *rspec.Spec {
 }
 
 // Save writes the spec into w.
-func (g *Generator) Save(w io.Writer) error {
-	data, err := json.MarshalIndent(g.spec, "", "\t")
+func (g Generator) Save(w io.Writer, exportOpts ExportOptions) (err error) {
+	var data []byte
+
+	if exportOpts.Seccomp {
+		data, err = json.MarshalIndent(g.spec.Linux.Seccomp, "", "\t")
+	} else {
+		data, err = json.MarshalIndent(g.spec, "", "\t")
+	}
 	if err != nil {
 		return err
 	}
@@ -195,13 +208,13 @@ func (g *Generator) Save(w io.Writer) error {
 }
 
 // SaveToFile writes the spec into a file.
-func (g *Generator) SaveToFile(path string) error {
+func (g Generator) SaveToFile(path string, exportOpts ExportOptions) error {
 	f, err := os.Create(path)
 	if err != nil {
 		return err
 	}
 	defer f.Close()
-	return g.Save(f)
+	return g.Save(f, exportOpts)
 }
 
 // SetVersion sets g.spec.Version.
@@ -465,278 +478,6 @@ func (g *Generator) RemoveLinuxSysctl(key string) {
 		return
 	}
 	delete(g.spec.Linux.Sysctl, key)
-}
-
-// SetLinuxSeccompDefault sets g.spec.Linux.Seccomp.DefaultAction.
-func (g *Generator) SetLinuxSeccompDefault(sdefault string) error {
-	switch sdefault {
-	case "":
-	case "SCMP_ACT_KILL":
-	case "SCMP_ACT_TRAP":
-	case "SCMP_ACT_ERRNO":
-	case "SCMP_ACT_TRACE":
-	case "SCMP_ACT_ALLOW":
-	default:
-		return fmt.Errorf("seccomp-default must be empty or one of " +
-			"SCMP_ACT_KILL|SCMP_ACT_TRAP|SCMP_ACT_ERRNO|SCMP_ACT_TRACE|" +
-			"SCMP_ACT_ALLOW")
-	}
-
-	g.initSpecLinuxSeccomp()
-	g.spec.Linux.Seccomp.DefaultAction = rspec.Action(sdefault)
-	return nil
-}
-
-func checkSeccompArch(arch string) error {
-	switch arch {
-	case "":
-	case "SCMP_ARCH_X86":
-	case "SCMP_ARCH_X86_64":
-	case "SCMP_ARCH_X32":
-	case "SCMP_ARCH_ARM":
-	case "SCMP_ARCH_AARCH64":
-	case "SCMP_ARCH_MIPS":
-	case "SCMP_ARCH_MIPS64":
-	case "SCMP_ARCH_MIPS64N32":
-	case "SCMP_ARCH_MIPSEL":
-	case "SCMP_ARCH_MIPSEL64":
-	case "SCMP_ARCH_MIPSEL64N32":
-	default:
-		return fmt.Errorf("seccomp-arch must be empty or one of " +
-			"SCMP_ARCH_X86|SCMP_ARCH_X86_64|SCMP_ARCH_X32|SCMP_ARCH_ARM|" +
-			"SCMP_ARCH_AARCH64SCMP_ARCH_MIPS|SCMP_ARCH_MIPS64|" +
-			"SCMP_ARCH_MIPS64N32|SCMP_ARCH_MIPSEL|SCMP_ARCH_MIPSEL64|" +
-			"SCMP_ARCH_MIPSEL64N32")
-	}
-	return nil
-}
-
-// ClearLinuxSeccompArch clears g.spec.Linux.Seccomp.Architectures.
-func (g *Generator) ClearLinuxSeccompArch() {
-	if g.spec == nil || g.spec.Linux == nil || g.spec.Linux.Seccomp == nil {
-		return
-	}
-
-	g.spec.Linux.Seccomp.Architectures = []rspec.Arch{}
-}
-
-// AddLinuxSeccompArch adds sArch into g.spec.Linux.Seccomp.Architectures.
-func (g *Generator) AddLinuxSeccompArch(sArch string) error {
-	if err := checkSeccompArch(sArch); err != nil {
-		return err
-	}
-
-	g.initSpecLinuxSeccomp()
-	g.spec.Linux.Seccomp.Architectures = append(g.spec.Linux.Seccomp.Architectures, rspec.Arch(sArch))
-
-	return nil
-}
-
-// RemoveSeccompArch removes sArch from g.spec.Linux.Seccomp.Architectures.
-func (g *Generator) RemoveSeccompArch(sArch string) error {
-	if err := checkSeccompArch(sArch); err != nil {
-		return err
-	}
-
-	if g.spec == nil || g.spec.Linux == nil || g.spec.Linux.Seccomp == nil {
-		return nil
-	}
-
-	for i, arch := range g.spec.Linux.Seccomp.Architectures {
-		if string(arch) == sArch {
-			g.spec.Linux.Seccomp.Architectures = append(g.spec.Linux.Seccomp.Architectures[:i], g.spec.Linux.Seccomp.Architectures[i+1:]...)
-			return nil
-		}
-	}
-
-	return nil
-}
-
-func checkSeccompSyscallAction(syscall string) error {
-	switch syscall {
-	case "":
-	case "SCMP_ACT_KILL":
-	case "SCMP_ACT_TRAP":
-	case "SCMP_ACT_ERRNO":
-	case "SCMP_ACT_TRACE":
-	case "SCMP_ACT_ALLOW":
-	default:
-		return fmt.Errorf("seccomp-syscall action must be empty or " +
-			"one of SCMP_ACT_KILL|SCMP_ACT_TRAP|SCMP_ACT_ERRNO|" +
-			"SCMP_ACT_TRACE|SCMP_ACT_ALLOW")
-	}
-	return nil
-}
-
-func checkSeccompSyscallArg(arg string) error {
-	switch arg {
-	case "":
-	case "SCMP_CMP_NE":
-	case "SCMP_CMP_LT":
-	case "SCMP_CMP_LE":
-	case "SCMP_CMP_EQ":
-	case "SCMP_CMP_GE":
-	case "SCMP_CMP_GT":
-	case "SCMP_CMP_MASKED_EQ":
-	default:
-		return fmt.Errorf("seccomp-syscall args must be " +
-			"empty or one of SCMP_CMP_NE|SCMP_CMP_LT|" +
-			"SCMP_CMP_LE|SCMP_CMP_EQ|SCMP_CMP_GE|" +
-			"SCMP_CMP_GT|SCMP_CMP_MASKED_EQ")
-	}
-	return nil
-}
-
-func parseSeccompSyscall(s string) (rspec.Syscall, error) {
-	syscall := strings.Split(s, ":")
-	if len(syscall) != 3 {
-		return rspec.Syscall{}, fmt.Errorf("seccomp sysctl must consist of 3 parameters")
-	}
-	name := syscall[0]
-	if err := checkSeccompSyscallAction(syscall[1]); err != nil {
-		return rspec.Syscall{}, err
-	}
-	action := rspec.Action(syscall[1])
-
-	var Args []rspec.Arg
-	if strings.EqualFold(syscall[2], "") {
-		Args = nil
-	} else {
-		argsslice := strings.Split(syscall[2], ",")
-		for _, argsstru := range argsslice {
-			args := strings.Split(argsstru, "/")
-			if len(args) == 4 {
-				index, err := strconv.Atoi(args[0])
-				value, err := strconv.Atoi(args[1])
-				value2, err := strconv.Atoi(args[2])
-				if err != nil {
-					return rspec.Syscall{}, err
-				}
-				if err := checkSeccompSyscallArg(args[3]); err != nil {
-					return rspec.Syscall{}, err
-				}
-				op := rspec.Operator(args[3])
-				Arg := rspec.Arg{
-					Index:    uint(index),
-					Value:    uint64(value),
-					ValueTwo: uint64(value2),
-					Op:       op,
-				}
-				Args = append(Args, Arg)
-			} else {
-				return rspec.Syscall{}, fmt.Errorf("seccomp-sysctl args error: %s", argsstru)
-			}
-		}
-	}
-
-	return rspec.Syscall{
-		Name:   name,
-		Action: action,
-		Args:   Args,
-	}, nil
-}
-
-// ClearLinuxSeccompSyscall clears g.spec.Linux.Seccomp.Syscalls.
-func (g *Generator) ClearLinuxSeccompSyscall() {
-	if g.spec == nil || g.spec.Linux == nil || g.spec.Linux.Seccomp == nil {
-		return
-	}
-
-	g.spec.Linux.Seccomp.Syscalls = []rspec.Syscall{}
-}
-
-// AddLinuxSeccompSyscall adds sSyscall into g.spec.Linux.Seccomp.Syscalls.
-func (g *Generator) AddLinuxSeccompSyscall(sSyscall string) error {
-	f, err := parseSeccompSyscall(sSyscall)
-	if err != nil {
-		return err
-	}
-
-	g.initSpecLinuxSeccomp()
-	g.spec.Linux.Seccomp.Syscalls = append(g.spec.Linux.Seccomp.Syscalls, f)
-	return nil
-}
-
-// AddLinuxSeccompSyscallAllow adds seccompAllow into g.spec.Linux.Seccomp.Syscalls.
-func (g *Generator) AddLinuxSeccompSyscallAllow(seccompAllow string) {
-	syscall := rspec.Syscall{
-		Name:   seccompAllow,
-		Action: "SCMP_ACT_ALLOW",
-	}
-
-	g.initSpecLinuxSeccomp()
-	g.spec.Linux.Seccomp.Syscalls = append(g.spec.Linux.Seccomp.Syscalls, syscall)
-}
-
-// AddLinuxSeccompSyscallErrno adds seccompErrno into g.spec.Linux.Seccomp.Syscalls.
-func (g *Generator) AddLinuxSeccompSyscallErrno(seccompErrno string) {
-	syscall := rspec.Syscall{
-		Name:   seccompErrno,
-		Action: "SCMP_ACT_ERRNO",
-	}
-
-	g.initSpecLinuxSeccomp()
-	g.spec.Linux.Seccomp.Syscalls = append(g.spec.Linux.Seccomp.Syscalls, syscall)
-}
-
-// RemoveSeccompSyscallByName removes all the seccomp syscalls with the given
-// name from g.spec.Linux.Seccomp.Syscalls.
-func (g *Generator) RemoveSeccompSyscallByName(name string) error {
-	if g.spec == nil || g.spec.Linux == nil || g.spec.Linux.Seccomp == nil {
-		return nil
-	}
-
-	var r []rspec.Syscall
-	for _, syscall := range g.spec.Linux.Seccomp.Syscalls {
-		if strings.Compare(name, syscall.Name) != 0 {
-			r = append(r, syscall)
-		}
-	}
-	g.spec.Linux.Seccomp.Syscalls = r
-	return nil
-}
-
-// RemoveSeccompSyscallByAction removes all the seccomp syscalls with the given
-// action from g.spec.Linux.Seccomp.Syscalls.
-func (g *Generator) RemoveSeccompSyscallByAction(action string) error {
-	if g.spec == nil || g.spec.Linux == nil || g.spec.Linux.Seccomp == nil {
-		return nil
-	}
-
-	if err := checkSeccompSyscallAction(action); err != nil {
-		return err
-	}
-
-	var r []rspec.Syscall
-	for _, syscall := range g.spec.Linux.Seccomp.Syscalls {
-		if strings.Compare(action, string(syscall.Action)) != 0 {
-			r = append(r, syscall)
-		}
-	}
-	g.spec.Linux.Seccomp.Syscalls = r
-	return nil
-}
-
-// RemoveSeccompSyscall removes all the seccomp syscalls with the given
-// name and action from g.spec.Linux.Seccomp.Syscalls.
-func (g *Generator) RemoveSeccompSyscall(name string, action string) error {
-	if g.spec == nil || g.spec.Linux == nil || g.spec.Linux.Seccomp == nil {
-		return nil
-	}
-
-	if err := checkSeccompSyscallAction(action); err != nil {
-		return err
-	}
-
-	var r []rspec.Syscall
-	for _, syscall := range g.spec.Linux.Seccomp.Syscalls {
-		if !(strings.Compare(name, syscall.Name) == 0 &&
-			strings.Compare(action, string(syscall.Action)) == 0) {
-			r = append(r, syscall)
-		}
-	}
-	g.spec.Linux.Seccomp.Syscalls = r
-	return nil
 }
 
 func parseIDMapping(idms string) (rspec.IDMapping, error) {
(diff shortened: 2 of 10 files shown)


GrantSeltzer:
@mrunalp Not sure how to fix this error.

mrunalp:
@GrantSeltzer Your commit is missing a Signed-off-by.

wking:
ae08020d7 has recursive Godeps :p.  I'm not sure if that's an expected
thing, but 103 vendored files and 10k lines is a lot to pull in for
this change.  Maybe we can get a smaller library with the bits we need
without Manhattan → runc → docker → libtrust and similar dependency
chains?

generate.go
```diff
@@ -27,7 +29,6 @@ var generateFlags = []cli.Flag{
 	cli.StringSliceFlag{Name: "groups", Usage: "supplementary groups for the process"},
 	cli.StringSliceFlag{Name: "cap-add", Usage: "add capabilities"},
 	cli.StringSliceFlag{Name: "cap-drop", Usage: "drop capabilities"},
-	cli.StringFlag{Name: "cgroup", Usage: "cgroup namespace"},
```
wking: You probably didn't intend to remove this (it landed in #122).  Maybe a busted rebase?

man/ocitools-generate.1.md
```diff
@@ -153,26 +153,43 @@ inside of the container.
...
-
-**--seccomp-allow**=SYSCALL
-  Specifies syscalls to be added to the ALLOW list.
-  See --seccomp-syscalls for setting limits on arguments.
-
-**--seccomp-errno**=SYSCALL
-  Specifies syscalls to be added to the ERRNO list.
-  See --seccomp-syscalls for setting limits on arguments.
+  **--seccomp-arch**=ARCH
+    Specifies Additional architectures permitted to be used for system calls.
```
wking: You're increasing the indent (the rest of these entries have no whitespace before the initial `**`, and only two spaces before subsequent lines.

GrantSeltzer:
@wking Not  sure what's causing that godeps recursive chain. Do you know why that's happening?

wking:
On Fri, Jul 01, 2016 at 04:52:03PM -0700, Grant Seltzer Richman wrote:
> @wking Not sure what's causing that godeps recursive chain. Do you
> know why that's happening?

I'm not familiar with Godeps, but if it is just based on Git clones
maybe it always pulls in the whole repository (and not just a
particular referenced package)?  If so, the fix is probably “get
whatever portion of $DEPENDENCY you need pulled out into an
independent Git repository”, ideally by getting upstream to split
things up for you.  But maybe the nesting is happening for a different
reason.

GrantSeltzer:
> I'm not familiar with Godeps, but if it is just based on Git clones maybe it always pulls in the whole repository (and not just a particular referenced package)? If so, the fix is probably “get whatever portion of $DEPENDENCY you need pulled out into an independent Git repository”, ideally by getting upstream to split things up for you. But maybe the nesting is happening for a different reason.

Yea, the issue is that I have the library ociseccompgen and the tool manhattan in the same git repo. I didn't expect  `go get` to pull the whole repo instead of just the package.

For the sake of this this PR at least `Godeps.json` should be updated to include the ociseccompgen library, have the actual library stored and remove the embedded Godeps directory which is holding dependencies that ocitools already has (cli, logrus, runtime-sepc)

I can do that along with the fixing spacing and re-adding cgroup flag that I accidentally removed. Thoughts?

wking:
On Fri, Jul 01, 2016 at 05:12:30PM -0700, Grant Seltzer Richman wrote:
> I can do that along with the fixing spacing and re-adding cgroup
> flag that I accidentally removed. Thoughts?

I don't know enough about Godeps to know what the right fix is ;).
But fixing these other two nits would be good.

generate.go
```diff
@@ -9,6 +9,8 @@ import (
 	"strconv"
 	"strings"
 
+	seccomp "github.com/grantseltzer/Manhattan/oci-seccomp-gen"
+
```
liangchenye: remove this line.

generate.go
```diff
@@ -54,10 +55,13 @@ var generateFlags = []cli.Flag{
...
-	cli.StringSliceFlag{Name: "seccomp-syscalls", Usage: "specifies Additional architectures permitted to be used for system calls, e.g Name:Action:Arg1_index/Arg1_value/Arg1_valuetwo/Arg1_op, Arg2_index/Arg2_value/Arg2_valuetwo/Arg2_op "},
-	cli.StringSliceFlag{Name: "seccomp-allow", Usage: "specifies syscalls to be added to allowed"},
-	cli.StringSliceFlag{Name: "seccomp-errno", Usage: "specifies syscalls to be added to list that returns an error"},
+	cli.StringFlag{Name: "seccomp-arch", Usage: "specifies Additional architectures permitted to be used for system calls"},
+	cli.StringFlag{Name: "seccomp-allow", Usage: "specifies syscalls to be added to allowed"},
+	cli.StringFlag{Name: "seccomp-trap", Usage: "specifies syscalls to be added to list that returns an error"},
+	cli.StringFlag{Name: "seccomp-errno", Usage: "specifies syscalls to be added to list that returns an error"},
+	cli.StringFlag{Name: "seccomp-trace", Usage: "specifies syscalls to be added to list that returns an error"},
+	cli.StringFlag{Name: "seccomp-kill", Usage: "specifies syscalls to be added to list that returns an error"},
+
```
liangchenye: also remove this line

liangchenye:
@GrantSeltzer I follow [instruction](https://github.com/tools/godep) to update your dependency, it works, no extra 'Godeps' added. I'm not sure if it also helps in your environment.

generate.go
```diff
@@ -232,172 +236,58 @@ func modify(spec *rspec.Spec, context *cli.Context) error {
...
 		return err
 	}
 
-	for _, a := range context.StringSlice("seccomp-allow") {
-		sysCall := rspec.Syscall{
-			Name:   a,
-			Action: "SCMP_ACT_ALLOW",
-		}
-		secc.Syscalls = append(secc.Syscalls, sysCall)
+	err = seccomp.ParseSyscallFlag("trace", seccompTrace, &secc)
```
liangchenye: maybe put all the five `ParseSyscallFlag` functions to one loop?

generate.go
```diff
@@ -232,172 +236,58 @@ func modify(spec *rspec.Spec, context *cli.Context) error {
...
+	if err != nil {
+		return err
+	}
+
+	err = seccomp.ParseSyscallFlag("allow", seccompAllow, &secc)
+	if err != nil {
+		return err
 	}
 
 	spec.Linux.Seccomp = &secc
```
liangchenye: The function `parseArgs()` follows this one is useless with your commit, so you can remove that.

liangchenye:
Hi @GrantSeltzer ,  `ocitools generate` is a small and basic tool, `oci-seccomp-gen` looks more like a part of it. So personally I don't like `ocitools generete` to depend on `oci-seccomp-gen`. But I think it is good to have `oci-seccomp-gen` as an independent pkg since it could be used in different scenarios (now we have ocitools and docker). Would you mind/like to 'donate' your project to ocitools? 

 @mrunalp @wking  @Mashimiao what you do think about this? The code struct could look like this:

```
ocitools/pkg/seccomp
ocitools/contrib/seccomp-generete (originial from https://github.com/GrantSeltzer/Manhattan/blob/master/main.go)
```

GrantSeltzer:
@rhatdan @runcom @mheon what do you think of that idea? Manhattan can import its functionality from ocitools instead of inside it's own repo. I'm personally fine with it.

runcom:
I'm +1k to include that library directly in ocitools and have Manhattan or ocitools itself to manage and generate seccomp profiles or OCI configs

wking:
On Mon, Jul 04, 2016 at 02:22:52AM -0700, 梁辰晔 (Liang Chenye) wrote:
> ocitools/pkg/seccomp

Is ‘pkg’ more idiomatic than ‘lib’?  I don't read enough Go to know,
but for a C project I'd expect ‘lib’.

> ocitools/contrib/seccomp-generete (originial from
> https://github.com/GrantSeltzer/Manhattan/blob/master/main.go)

Why keep this executable separate?  Isn't this PR just rolling it into
the bigger ‘ocitool generate …’ picture?

runcom:
No pkg nor lib, just the name of the library is go idiomatic

runcom:
+1 also on having the main tool work with the library directly instead of creating a contrib dir with binaries in there. So just:

./seccomp

And code from that used in the main ocitools

GrantSeltzer:
I like the idea of having the ociseccompgen be part of the ocitools library, that makes a lot of sense.

 Just to clarify, are we in disagreement that seccomp should be a flag for the `./ocitools generate` command? That way it's consistent with other configuration options that generate supports (i.e. `--cgroups-path` or `--hostname`) 

Manhattan is fully functioning for generated just the seccomp configurations: https://github.com/GrantSeltzer/Manhattan

mrunalp:
Could add a flag --seccomp-only to ocitools generate

Mashimiao:
`ocitools/contrib/seccomp-generete (originial from https://github.com/GrantSeltzer/Manhattan/blob/master/main.go)`

I'm +1 for this.

liangchenye:
OK, +1 for  `a flag --seccomp-only to ocitools generate`.

GrantSeltzer:
Alright, I'm working on the following points:

1) Rehost the `ociseccompgen` library into the ocitools repo

2) Add all of the flags like `seccomp-allow` that I added in this PR

3) Add `--seccomp-only` flag to just export a seccomp file

Should this be it's own PR (this would be closed) or squashed with the commit for this one?

wking:
On Tue, Jul 05, 2016 at 12:23:18PM -0700, Grant Seltzer Richman wrote:
> 3) Add `--seccomp-only` flag to just export a seccomp file

What is the use-case for this?  Can we accomplish the same thing by
having ‘ocitools generate’ write (optionally?) to stdout and use:

  $ ocitools generate … | jq .linux.seccomp

GrantSeltzer:
The use case is for use with other tools like docker that you can pass a seccomp configuration file to a container or (soon) the daemon. And yes piping into jq would do the same thing, But I think this would be more user friendly and feel less like a hack.

wking:
On Tue, Jul 05, 2016 at 12:39:19PM -0700, Grant Seltzer Richman wrote:
> The use case is for use with other tools like docker that you can
> pass a seccomp configuration file to a container or (soon) the
> daemon. And yes piping into jq would do the same thing, But I think
> this would be more user friendly and feel less like a hack.

I'm just not sure how generically we should scope the solution.  I'd
rather not end up with an API that has lots of --*-only flags.  If we
don't want to use jq, we could also ‘--subsection linux.seccomp’ or
some such to handle the drilling internally.

And I think using jq for JSON dicing is using the proper (already
written) tool for the task.  It doesn't feel like a hack ;).

And regardless of how we address this, always writing to config.json
doesn't seem appropriate for the seccomp-only use case.  What
command-line API do we want for “please write the generated config to
stdout”?  I'd rather do that by default, with an ‘--output PATH’
option to write to a file (for situations where shell redirection is
not available).

GrantSeltzer:
I think having a flag for `generate` to rename the output file should be added, although that's a more general issue than this PR. Certainly the default for seccomp output could be something like "config.seccomp". 

I'm not sure what other configurations can be independently used besides seccomp but I think the `seccomp-only` flag would definitely be used and makes more sense than having an independent tool ([Manhattan](https://github.com/grantseltzer/manhattan)). I definitely concede that using jq wouldn't be a hack but having this feature would give more visibility to the idea of using ocitools for it. I also think the feature adds minimal change to the code base.

wking:
On Tue, Jul 05, 2016 at 01:51:53PM -0700, Grant Seltzer Richman wrote:
> I think having a flag for `generate` to rename the output file
> should be added, although that's a more general issue than this PR.

I've filed --output as #129.

> I'm not sure what other configurations can be independently used
> besides seccomp but I think the `seccomp-only` flag would definitely
> be used and makes more sense than having an independent tool
> ([Manhattan](https://github.com/grantseltzer/manhattan)).

I agree that having an independent, OCI-specific tool like Manhattan
is less ideal than either jq or an ocitools flag.  If the current
consensus is that --seccomp-only is unlikely to be the first step down
a slippery slope, I'm ok with going down that path for now.  We can
always revisit the decision if/when we find ourselves sliding down the
--*-only slope ;).

GrantSeltzer:
Alright, if people could re-review this PR that would be great. I made the changes as discussed above:

> 1) Rehost the ociseccompgen library into the ocitools repo

> 2) Add flags for each seccomp action for specifying syscalls and possibly actions. As well as for setting architectures and default action

> 3) Add --seccomp-only flag to just export a seccomp file

> 4) Updated README and man page to document these changes

Godeps/Godeps.json
```diff
@@ -14,6 +13,10 @@
 			"Rev": "aea32c919a18e5ef4537bbd283ff29594b1b0165"
 		},
 		{
+			"ImportPath": "github.com/opencontainers/ocitools/contrib/oci-seccomp-gen",
```
liangchenye: Don't need to have this dependency in Godeps.

contrib/oci-seccomp-gen/parse_arguments.go
```diff
@@ -0,0 +1,71 @@
...
+	"errors"
+	"fmt"
+	"strconv"
+
+	types "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+// parseArguments takes a list of arguments (delimArgs)  and a pointer to a
+// corresponding syscall struct. It parses and fills out the argument information
+func parseArguments(delimArgs []string) (*[]types.Arg, error) {
```
wking: This comment does not seem to match the function signature (I don't see any input `*types.Syscall`).

contrib/oci-seccomp-gen/parse_location.go
```diff
@@ -0,0 +1,49 @@
...
+	"os/user"
+	"strconv"
+	"strings"
+
+	"github.com/Sirupsen/logrus"
+)
+
+// DefaultFullPath returns the default full path/name for output configuration files
+func DefaultFullPath() string {
+	return (parseLocation(pwd(), parseNameWithNumber("manhattan", 0)))
```
wking: This `manhattan` reference seems stale, and I can't find any consumers using `parse_location.go` code outside of this file.  Maybe drop it from the import?

contrib/oci-seccomp-gen/syscall_compare.go
```diff
@@ -0,0 +1,145 @@
...
+	"strings"
+
+	types "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+/*******************************************************
+This file is a bunch of helper functions for determining
+what action should be taken for new syscalls.
+See github.com/grantseltzer/manhattan/Resources/Logic.png
+for a flowchart and explanation
```
wking: Are these docs coming over too?

wking:
On Tue, Jul 05, 2016 at 07:50:27PM -0700, Grant Seltzer Richman wrote:
> > 1) Rehost the ociseccompgen library into the ocitools repo

Manhattan was under the GPL [1], but with only you and @runcom
contributing [2], I suspect you're both fine relicensing it under the
Apache v2 [3].

With ocitools in the URL, I'd also rather see the package called
‘seccomp’ [4] instead of ‘contrib/oci-seccomp-gen’, but that's not a
big deal either way.

[1]: https://github.com/GrantSeltzer/Manhattan/blob/22051082012dfe7bea030f85b8f600aa1bd39354/LICENSE
[2]: https://github.com/GrantSeltzer/Manhattan/graphs/contributors
[3]: https://github.com/opencontainers/ocitools/blob/239e8e959e5dc236d54d007ec162b11919df0c9e/LICENSE
[4]: https://github.com/opencontainers/ocitools/pull/128#issuecomment-230353540

liangchenye:
The seccomp-gen should not be in the 'contrib' directory. 
If there will be more libs, I prefer to put it in `ocitools/pkg/seccom-gen`.
@wking I only read few golang projects and feel `pkg` is used more frequently. ([docker](https://github.com/docker/docker/blob/master/pkg/README.md)/rkt/kubernetes)

runcom:
Why don't just have seccomp-gen directly without any contrib or pkg. The pkg you're referring to is for huge projects where it makes sense to do that. I still vote to have the library along with the other files without any fancy directly....

wking:
On Tue, Jul 05, 2016 at 11:38:43PM -0700, Antonio Murdaca wrote:
> Why don't just have seccomp-gen directly without any contrib or
> pkg. The pkg you're referring to is for huge projects where it makes
> sense to do that. I still vote to have the library along with the
> other files without any fancy directly....

And (as I said before) I'm not familiar with the Go idioms here, but
using just ‘seccomp’ (or, I guess, ‘seccomp-gen’) sounds good to me.

liangchenye:
@runcom  I'm fine with `ocitools/seccomp-gen`.  I add a `pkg` just in case `If there will be more libs`, so now it is not necessary.

runcom:
Great! :)

runcom:
Wrt the license, I'm not entirely sure, @rhatdan?

contrib/oci-seccomp-gen/parse_location.go
GrantSeltzer: `parse_location.go` was used specifically for the `manhattan` tool so I can see dropping it.

Godeps/Godeps.json
GrantSeltzer: Makes sense :)

contrib/oci-seccomp-gen/syscall_compare.go
GrantSeltzer: I think it's a good idea, What do you think?

wking: On Wed, Jul 06, 2016 at 07:00:14AM -0700, Grant Seltzer Richman wrote:
> I think it's a good idea, What do you think?

The flow-chart looks useful, but I'd like to have whatever vector
source you used to develop it alongside the raster PNG.

GrantSeltzer:
Is there a license issue? Can't the library be issued under a different license?

runcom:
Nope, just wanted Dan to confirm we don't have issue here

rhatdan:
I have no problem re-licensing to Apache.

GrantSeltzer:
Fixing conflicts now, I have no problem with it either

seccomp-gen/parse_arguments.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"strconv"
+
+	types "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+// parseArguments takes a list of arguments (delimArgs).It parses and fills out
```
wking: Missing whitespace after the period.

GrantSeltzer:
Alright so now all of the seccomp actions have flags, and there's a flag for exporting only the seccomp portion. The seccomp-gen library has been moved to within the oci-tools repo. How's it look?

wking:
9fd82e5 → 9d1203e restored the old parseArgs in generate.go.  This is probably an accidental side-effect of the rebase.

I sent that^ comment in via email ~40 minutes ago (so maybe GitHub ate it), in case the email eventually gets through and folks wonder why I double-posted it ;).

GrantSeltzer:
@wking Woops! Fixing now. I'm not very good with git yet.

GrantSeltzer:
@mrunalp Could you take a look at this today?

mrunalp:
This PR can now be reworked on top of #131

GrantSeltzer:
working on it now @mrunalp

GrantSeltzer:
@mrunalp @hmeng-19 

I want to double check that it's ok to remove the `seccomp-syscalls` flag. We would only need the `seccomp-ACTION` flags and ones for architectures, and default actions. Right now there's only ones for `allow` and `errno` anyway. 

Sytanx should like this:

`ocitools --seccomp-kill clone:0:1:2:NE,getcwd --seccomp-default allow --seccomp-arch mips,arm,arm64`

mrunalp:
@GrantSeltzer SGTM

GrantSeltzer:
@mrunalp Please take a look

cmd/ocitools/generate.go
```diff
@@ -48,11 +51,14 @@ var generateFlags = []cli.Flag{
...
-	cli.StringSliceFlag{Name: "seccomp-arch", Usage: "specifies Additional architectures permitted to be used for system calls"},
-	cli.StringSliceFlag{Name: "seccomp-syscalls", Usage: "specifies Additional architectures permitted to be used for system calls, e.g Name:Action:Arg1_index/Arg1_value/Arg1_valuetwo/Arg1_op, Arg2_index/Arg2_value/Arg2_valuetwo/Arg2_op "},
-	cli.StringSliceFlag{Name: "seccomp-allow", Usage: "specifies syscalls to be added to allowed"},
-	cli.StringSliceFlag{Name: "seccomp-errno", Usage: "specifies syscalls to be added to list that returns an error"},
+	cli.BoolFlag{Name: "seccomp-only", Usage: "specifies to export just a seccomp configuration file"},
+	cli.StringFlag{Name: "seccomp-arch", Usage: "specifies Additional architectures permitted to be used for system calls"},
+	cli.StringFlag{Name: "seccomp-default", Usage: "specifies default aciton to be used for system calls"},
+	cli.StringFlag{Name: "seccomp-allow", Usage: "specifies syscalls to be added to allowed"},
+	cli.StringFlag{Name: "seccomp-trap", Usage: "specifies syscalls to be added to list that returns an error"},
```
mrunalp: should the help text be same for all these options?

cmd/ocitools/generate.go
```diff
@@ -301,64 +310,79 @@ func setupSpec(g generate.Generator, context *cli.Context) error {
...
+	seccompErrno := context.String("seccomp-errno")
+	seccompTrap := context.String("seccomp-trap")
+	seccompAllow := context.String("seccomp-allow")
 
 	// Set the DefaultAction of seccomp
-	if context.IsSet("seccomp-default") {
-		if err := g.SetLinuxSeccompDefault(sd); err != nil {
-			return err
-		}
+	if seccompDefault == "" {
```
mrunalp: This shouldn't be defaulted to allow. We should figure out intent and error out if we can't. For e.g. if user only passes seccomp-allow, then we can assume user is building a whitelist and we can default to `errno`.

generate/seccomp/consts.go
```diff
@@ -0,0 +1,7 @@
+package seccomp
+
+const (
+	overwrite = "overwrite"
+	appnd     = "append"
```
mrunalp: typo: `appnd` -> `append`

generate/seccomp/parse_action.go
```diff
@@ -0,0 +1,105 @@
...
+
+	"github.com/Sirupsen/logrus"
+	types "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+// ParseSyscallFlag takes the name of the action, the arguments (syscalls) that were
+// passed with it at the command line and a pointer to the config struct. It parses
+// the action and syscalls and updates the config accordingly
+func ParseSyscallFlag(action string, arguments string, config *types.Seccomp) error {
+
```
mrunalp: nit: Can remove the empty line.

generate/seccomp/consts.go
GrantSeltzer: append is a reserved word

wking: On Thu, Jul 21, 2016 at 11:41:09AM -0700, Grant Seltzer Richman wrote:
> > @@ -0,0 +1,7 @@
> > +package seccomp
> > +
> > +const (
> > +	overwrite = "overwrite"
> > +	appnd     = "append"
> 
> append is a reserved word

APPEND?  (I think Go is case sensitive, and CAPS for package constants
is idiomatic in Python, although I'm not sure about Go).

cmd/ocitools/generate.go
GrantSeltzer: Agreed :+1:

generate/seccomp/consts.go
GrantSeltzer: Sure, I'm fine with that. I'll make `overwrite` all caps as well

GrantSeltzer: Well, the only issue is that it would then be exported, not sure if that makes a difference

wking: On Thu, Jul 21, 2016 at 11:48:02AM -0700, Grant Seltzer Richman wrote:
> > @@ -0,0 +1,7 @@
> > +package seccomp
> > +
> > +const (
> > +	overwrite = "overwrite"
> > +	appnd     = "append"
>
> Well, the only issue is that it would then be exported, not sure if
> that makes a difference

If these are just for internal use, maybe prefix them (seccomp_append,
seccomp_overwrite).

GrantSeltzer: Sure, good idea

GrantSeltzer:
Fixed suggestions, @mrunalp You can take another look :)

cmd/ocitools/generate.go
```diff
@@ -48,11 +51,14 @@ var generateFlags = []cli.Flag{
...
 	cli.StringFlag{Name: "apparmor", Usage: "specifies the the apparmor profile for the container"},
-	cli.StringFlag{Name: "seccomp-default", Usage: "specifies the the defaultaction of Seccomp syscall restrictions"},
-	cli.StringSliceFlag{Name: "seccomp-arch", Usage: "specifies Additional architectures permitted to be used for system calls"},
-	cli.StringSliceFlag{Name: "seccomp-syscalls", Usage: "specifies Additional architectures permitted to be used for system calls, e.g Name:Action:Arg1_index/Arg1_value/Arg1_valuetwo/Arg1_op, Arg2_index/Arg2_value/Arg2_valuetwo/Arg2_op "},
-	cli.StringSliceFlag{Name: "seccomp-allow", Usage: "specifies syscalls to be added to allowed"},
-	cli.StringSliceFlag{Name: "seccomp-errno", Usage: "specifies syscalls to be added to list that returns an error"},
+	cli.BoolFlag{Name: "seccomp-only", Usage: "specifies to export just a seccomp configuration file"},
+	cli.StringFlag{Name: "seccomp-arch", Usage: "specifies additional architectures permitted to be used for system calls"},
+	cli.StringFlag{Name: "seccomp-default", Usage: "specifies default aciton to be used for system calls"},
```
wking: “aciton” → “action”

wking: And you're following a pre-existing pattern, but I don't think “specifies …” adds much vs. “default action to be used for system calls”, etc.

cmd/ocitools/generate.go
```diff
@@ -382,3 +406,15 @@ func setupLinuxNamespaces(g generate.Generator, needsNewUser bool, nsMaps map[st
...
 	}
 }
+
+func onlyExportFlagSpecified(context *cli.Context) bool {
+	onlyExportFlags := []bool{
+		context.Bool("seccomp-only"),
+	}
+	for _, flag := range onlyExportFlags {
+		if flag {
+			return true
```
wking: What is this function about?  Why can't we just `context.Bool("seccomp-only")` instead of `onlyExportFlagSpecified`?

GrantSeltzer: It's a more general approach. If we wanted to add a feature to export just the configuration for one thing then it can be added into this function.

wking: On Thu, Jul 21, 2016 at 01:03:31PM -0700, Grant Seltzer Richman wrote:
> > +func onlyExportFlagSpecified(context *cli.Context) bool {
> > +	onlyExportFlags := []bool{
> > +		context.Bool("seccomp-only"),
> > +	}
> > +	for _, flag := range onlyExportFlags {
> > +		if flag {
> > +			return true
>
> It's a more general approach. If we wanted to add a feature to
> export just the configuration for one thing then it can be added
> into this function.

How does the caller specify the thing it's interested in?  And it
looks like you're ignoring --output here [1].  I'd rather extend
SaveToFile and Save to take context.  SaveToFile would pass it through
to Save, and Save would do any --*-only switching internally (without
needing a separate helper function).

[1]: https://github.com/opencontainers/ocitools/pull/128/files#diff-3e0e09fa01dec94676136d9c0cf52025R379

GrantSeltzer: I was saying so that we can add functionality for the same way `seccomp-only` would work

wking: On Thu, Jul 21, 2016 at 01:22:06PM -0700, Grant Seltzer Richman wrote:
> > +func onlyExportFlagSpecified(context *cli.Context) bool {
> > +	onlyExportFlags := []bool{
> > +		context.Bool("seccomp-only"),
> > +	}
> > +	for _, flag := range onlyExportFlags {
> > +		if flag {
> > +			return true
> 
> I was saying so that we can add functionality for the same way
> `seccomp-only` would work

But the way seccomp-only works in 3a5ad1e7 is by bypassing the usual
Save logic and rolling its own.  I'd rather have it work by teaching
the usual Save logic about --*-only.

GrantSeltzer: Not disagreeing, but why?

wking: On Thu, Jul 21, 2016 at 01:55:45PM -0700, Grant Seltzer Richman wrote:
> Not disagreeing, but why? 

For one, it means that the implementation would respect --output
without needing special code in lots of places ;).  More broadly, if
this functionality is useful enough for a command-line option, it
seems like it deserves a place in the library API.  Passing Context
through SaveToFile to Save is a classic use of Context for tunneling
parameters [1], and makes it easy for library callers to say “Please
save just the seccomp stuff to this path” without having to roll their
own duplicate of [2] or pulling in all the other things setupSpec is
looking at.

[1]: https://blog.golang.org/context#TOC_5.
     “[A] Context parameter as the first argument to every function on
     the call path… ensures that critical values like security
     credentials transit Go programs properly.”
[2]: https://github.com/opencontainers/ocitools/pull/128/files#diff-3e0e09fa01dec94676136d9c0cf52025R374

runcom: No, strongly disagreeing here to pass down the context - what you're referring to is only being done for Request scoped stuff and I'm not sure we have any request here (on my phone, I'll backup my sentence with soueces asap)

runcom: And this isn't even the official context you're linking the golang blob

wking: On Thu, Jul 21, 2016 at 02:42:22PM -0700, Antonio Murdaca wrote:
> No, strongly disagreeing here to pass down the context - what you're
> referring to is only being done for Request scoped stuff and I'm not
> sure we have any request here…

I'm pretty sure there aren't HTTP requests in ocitools, but the
command-line call creates a similar Context, and I don't see a reason
to distinguish between “serving an HTTP request”, “serving a CLI
request”, and “serving a library API request”.  What I'm looking for
is something like Python's **kwargs [1] for tunneling
low-level-function config through an intermediate.  If there's a more
canonical Go way of doing that, it would be fine too.  But exposing
seccomp_only (and similar stuff if we add more in the future) as
separate parameters in SaveToFile just seems like busywork.

[1]: https://docs.python.org/3/faq/programming.html?highlight=kwargs#how-can-i-pass-optional-or-keyword-parameters-from-one-function-to-another

runcom: You can have a struct to configure optional arguments to this Save function (or SaveToFile). I'm assuming those functions are from the libraries and in this case that context (coming from a cli package) should not clutter the library.

wking: On Thu, Jul 21, 2016 at 03:13:27PM -0700, Antonio Murdaca wrote:
> You can have a struct to configure optional arguments to this Save
> function (or SaveToFile). I'm assuming those functions are from the
> libraries and in this case that context (coming from a cli package)
> should not clutter the library.

Who cares where the Context comes from?  You'd just want to document
(on Save) the properties it looked at and (on SaveToFile) the fact
that it was a thin wrapper around Save.  A config struct accomplishes
the same result, but it feels kludgier to me.  But if a config struct
is more idiomatic Go than a Context argument, that's more important
than how I feel ;).

mrunalp: Yes to passing a config struct but no to reusing the context from the cli package.

GrantSeltzer: @mrunalp @runcom @wking  PTAL

generate/generate.go
```diff
@@ -21,7 +21,14 @@ var (
 
 // Generator represents a generator for a container spec.
 type Generator struct {
-	spec *rspec.Spec
+	spec       *rspec.Spec
+	ExportOpts *ExportOptions
```
wking: Can we just pass this to the `Save*` functions?  I think this is [per-call enough][1] to not want a Generator-wide setting.

[1]: https://github.com/opencontainers/ocitools/pull/138#discussion_r71803604

runcom: if it's a generator setting it should stay (and be configured) along with the generator to me

runcom: if, of course, it isn't I'm all for moving that to `Save*` but afaict it's a generator option

cmd/ocitools/generate.go
```diff
@@ -83,19 +87,30 @@ var generateCommand = cli.Command{
...
 
+		specgen.ExportOpts.Seccomp = context.Bool("seccomp-only")
+
 		if context.IsSet("output") {
-			output := context.String("output")
-			err = specgen.SaveToFile(output)
+			specgen.ExportOpts.Output = context.String("output")
 		} else {
-			err = specgen.Save(os.Stdout)
+			specgen.ExportOpts.Output = "STDOUT"
```
wking: I'd rather not have this magic string.

And the library API already has `Save` and `SaveToFile`.  I think keeping the current switch here is fine, in which case you can drop `exportConfig`.

generate/generate.go
wking: On Fri, Jul 22, 2016 at 11:20:20AM -0700, Antonio Murdaca wrote:
> > @@ -21,7 +21,14 @@ var (
> >  
> >  // Generator represents a generator for a container spec.
> >  type Generator struct {
> > -	spec *rspec.Spec
> > +	spec       *rspec.Spec
> > +	ExportOpts *ExportOptions
>
> if, of course, it isn't I'm all for moving that to `Save*` but
> afaict it's a generator option

Can you unpack “generator option” for me?  Since #131 we've been
trying to maintain two APIs: a library API and a command line API.  To
me, that means we develop the library API with an eye to expression
and flexibility, and then bind on some (probably less flexible)
convenient wrapper for the command line API.

I can imagine someone wanting to save both a full config and also only
the seccomp part, and don't want those calls competing for control of
the same Generator.ExportOpts resource.

cmd/ocitools/generate.go
GrantSeltzer: Not sure what you mean by keeping the current switch, can you clarify?

generate/generate.go
```diff
@@ -139,12 +146,14 @@ func New() Generator {
 			Devices: []rspec.Device{},
 		},
 	}
-	return Generator{&spec}
+	var outputOptions ExportOptions
+	return Generator{&spec, &outputOptions}
```
runcom: you can directly do `return Generator{&spec, &ExportOptions{}}`

generate/generate.go
```diff
@@ -163,10 +172,11 @@ func NewFromFile(path string) (Generator, error) {
 // NewFromTemplate loads the template from io.Reader into a spec Generator.
 func NewFromTemplate(r io.Reader) (Generator, error) {
 	var spec rspec.Spec
+	var expOpt ExportOptions
 	if err := json.NewDecoder(r).Decode(&spec); err != nil {
 		return Generator{}, err
 	}
-	return Generator{&spec}, nil
+	return Generator{&spec, &expOpt}, nil
```
runcom: ditto

generate/generate.go
runcom: Oh I see what you're saying (hopefully). I think the solution is not to tie this `ExportOption` to the `Generator` but just be a _dangling_ option struct which is directly provided to `Save*`. Does this work for you?

generate/generate.go
GrantSeltzer: Ah cool, didn't realize you could do that.

cmd/ocitools/generate.go
wking: On Fri, Jul 22, 2016 at 11:33:29AM -0700, Grant Seltzer Richman wrote:
> >  		} else {
> > -			err = specgen.Save(os.Stdout)
> > +			specgen.ExportOpts.Output = "STDOUT"
> 
> Not sure what you mean by keeping the current switch, can you clarify?

I meant keep the current:

  if context.IsSet("output") {
    output := context.String("output")
    err = specgen.SaveToFile(output)
  } else {
		err = specgen.Save(os.Stdout)
  }

which switches (if/else) in the caller instead of punting that choice
to exportConfig looking at specgen.ExportOpts.Output.

generate/generate.go
wking: On Fri, Jul 22, 2016 at 11:38:11AM -0700, Antonio Murdaca wrote:
> I think the solution is not to tie this `ExportOption` to the
> `Generator` but just be a _dangling_ option struct which is directly
> provided to `Save*`. Does this work for you?

Yeah, I want:

  Save(w io.Writer, ExportOpts *ExportOptions) error
  SaveToFile(path string, ExportOpts *ExportOptions) error

runcom: > Save(w io.Writer, ExportOpts *ExportOptions) error
>   SaveToFile(path string, ExportOpts *ExportOptions) error

+1

cmd/ocitools/generate.go
GrantSeltzer: Without having a function like `exportConfig` you would have to have the check for stdout/path inside both the if and the else, no? Doesn't having that relegated to one function call seem cleaner?

generate/generate.go
GrantSeltzer: Sure

cmd/ocitools/generate.go
wking: On Fri, Jul 22, 2016 at 12:09:48PM -0700, Grant Seltzer Richman wrote:
> >  		} else {
> > -			err = specgen.Save(os.Stdout)
> > +			specgen.ExportOpts.Output = "STDOUT"
>
> Without having a function like `exportConfig` you would have to have
> the check for stdout/path inside both the if and the else, no?

I don't understand.  [1] only checks the output path if
context.IsSet("output"), not in the else branch.

[1]: https://github.com/opencontainers/ocitools/pull/128/files#r71928565

GrantSeltzer: Ah, I see. Gotcha.

generate/generate.go
GrantSeltzer: Ready for another look

generate/generate.go
```diff
@@ -139,12 +145,13 @@ func New() Generator {
 			Devices: []rspec.Device{},
 		},
 	}
-	return Generator{&spec}
+	return Generator{&spec, &ExportOptions{}}
```
wking: You don't need these now that you've dropped it from `Generator`.

wking: Actually, it looks like you forgot to drop it from `Generator` ;).

GrantSeltzer: ughhh, fixing now :P

GrantSeltzer:
@mrunalp @rhatdan @runcom 

I'd like to open a fresh branch/PR if it's alright, the commit's for this one have gotten unmanageable because of multiple merge commits in the time it's been sitting. Besides the fail for the non-dco, everything seems good to me. Let me know whenever you're ready to look at this, Mrunal.

wking:
On Tue, Jul 26, 2016 at 12:33:19PM -0700, Grant Seltzer Richman wrote:
> I'd like to open a fresh branch/PR if it's alright, the commit's for
> this one have gotten unmanageable because of multiple merge commits
> in the time it's been sitting.

No need for a new PR.  Once you have a squashed commit, just
force-push it to your seccomp-cli-features branch.

wking:
Ah, and the base for the new PR should be somewhere before #140 landed
in f0ecc98 so we can merge this into the v1.0.0.rc1 branch too.  If
you do open a new PR, my personal preference would be for you to open
it on top of and targetting v1.0.0.rc1 [1,2].

[1]: https://github.com/opencontainers/ocitools/pull/140#issuecomment-234602988
[2]: https://github.com/opencontainers/ocitools/pull/140#issuecomment-234626841

runcom:
> Ah, and the base for the new commit/PR should be somewhere before #140 landed
> in f0ecc98 so we can merge this into the v1.0.0.rc1 branch too.  If
> you do open a new PR, my personal preference would be for you to open
> it on top of and targetting v1.0.0.rc1 [1,2].

I believe the last commit is rebased on top of the current ocitools master branch

GrantSeltzer:
@runcom just helped me out, ignore my last comment.

@mrunalp can you please review :)

wking:
On Tue, Jul 26, 2016 at 01:01:39PM -0700, Antonio Murdaca wrote:
> > Ah, and the base for the new commit/PR should be somewhere before
> > #140 landed in f0ecc98 so we can merge this into the v1.0.0.rc1
> > branch too.  If you do open a new PR, my personal preference would
> > be for you to open it on top of and targetting v1.0.0.rc1 [1,2].
> 
> I believe the last commit is rebased on top of the current ocitools
> master branch

I've rebased it backwards with:

  $ git rebase --onto 564b30e HEAD^

After resolving some conflicts in code you'd deleted (but which had
been updated in master), I ended up with 69526b9, which I've pushed to
[1].  You can compare with your current commit using:

  $ diff -u <(git show --color=always cbd3a2e4) <(git show --color=always 69526b9)

If that looks right to you, I'd like to see 69526b9 pushed to this PR
so it can be merged into both master and v1.0.0.rc1.  The alternative
is for me to PR 69526b9 into v1.0.0.rc1, but then we get to unify
those slightly-different versions of the commit at some later point
when we merge v1.0.0.rc1 into master to pick up other
v1.0.0.rc1-compatible changes.

[1]: https://github.com/wking/ocitools-v2/tree/seccomp-library

runcom:
> I've rebased it backwards with:
> 
>   $ git rebase --onto 564b30e HEAD^
> 
> After resolving some conflicts in code you'd deleted (but which had
> been updated in master), I ended up with 69526b9, which I've pushed to
> [1].  You can compare with your current commit using:
> 
>   $ diff -u <(git show --color=always cbd3a2e) <(git show --color=always 69526b9)
> 
> If that looks right to you, I'd like to see 69526b9 pushed to this PR
> so it can be merged into both master and v1.0.0.rc1.  The alternative
> is for me to PR 69526b9 into v1.0.0.rc1, but then we get to unify
> those slightly-different versions of the commit at some later point
> when we merge v1.0.0.rc1 into master to pick up other
> v1.0.0.rc1-compatible changes.

sorry, can't be just wait to land this in master and cherry-pick for the release? or create another pr for the release branch? it's so confusing for everyone like this.

runcom:
@GrantSeltzer will of course take care of opening a PR against the release branch once this is finished and afaict this isn't yet fully reviewed and lgtm'ed

wking:
On Tue, Jul 26, 2016 at 01:13:53PM -0700, Antonio Murdaca wrote:
> > I've rebased it backwards with:
> > 
> >   $ git rebase --onto 564b30e HEAD^
> > 
> > After resolving some conflicts in code you'd deleted (but which had
> > been updated in master), I ended up with 69526b9, which I've pushed to
> > [1].  You can compare with your current commit using:
> > 
> >   $ diff -u <(git show --color=always cbd3a2e) <(git show --color=always 69526b9)
> > 
> > If that looks right to you, I'd like to see 69526b9 pushed to this PR
> > so it can be merged into both master and v1.0.0.rc1.  The alternative
> > is for me to PR 69526b9 into v1.0.0.rc1, but then we get to unify
> > those slightly-different versions of the commit at some later point
> > when we merge v1.0.0.rc1 into master to pick up other
> > v1.0.0.rc1-compatible changes.
>
> sorry, can't be just wait to land this in master and cherry-pick for
> the release? or create another pr for the release branch?

That's the “The alternative is …” route above.  I'd rather not
cherry-pick, because it makes the later merge of v1.0.0.rc1 into
master more confusing.

> it's so confusing for everyone like this.

This is how lots of projects work.  For example see Git's section on
picking which branch to base your work on [1].

At some point ocitools is (I think) going to have to support multiple
runtime-spec releases at the same time, so maybe we want to work out
how we see that working now, implement it, and drop the v1.0.0.rc1
branch?  I have some thoughts on that in #98 for config validation,
but the thoughts there aren't all that useful for generate or runtime
validation.

[1]: https://git.kernel.org/cgit/git/git.git/tree/Documentation/SubmittingPatches?h=v2.9.2#n4

wking:
On Tue, Jul 26, 2016 at 01:14:36PM -0700, Antonio Murdaca wrote:
> @GrantSeltzer will of course take care of opening a PR against the
> release branch once this is finished and afaict this isn't yet fully
> reviewed and lgtm'ed

Sounds good, and it is good to wait until the branch is reviewed and
merged into one branch before filing other PRs.  If (for
v1.0.0.rc1-compatible features) we lead with a v1.0.0.rc1-based PR
[1], there's no need for a second PR.  The maintainer just merges
v1.0.0.rc1 into master after each v1.0.0.rc1 bump.

[1]: https://github.com/opencontainers/ocitools/pull/140#issuecomment-234602988

mrunalp:
I think the API should be easy to use just like rest of generate.
For e.g.
```
g := generate.New()
g.SetSeccompDefault(...)
g.AddSeccompErrno(...)
g.AddSeccompErrno(...)
g.SaveToFile(...)
```

The command line option parsing should be left to integration with the cmd.

GrantSeltzer:
@mrunalp How's that? I created a seccomp generator in the same way the generate library has.

mrunalp:
@GrantSeltzer No, this doesn't help. User shouldn't have to deal with two different generators when trying to generate a config. You could create a seccomp generator API and then use it from the regular generate API. The generate API should be simple and comprehensive.

For e.g.
```go
gen := generate.New()
gen.SetSeccompDefault(..) // These calls could use your own seccomp API behind the scenes.
gen.AddSeccompErrno(..)
gen.AddSeccompKill(..)
gen.SetRootfs(..)
gen.AddCapability(..)
gen.SaveToFile(..)
```

GrantSeltzer:
Just to clarify do you want all of the command line parsing done in the main package?

mrunalp:
@GrantSeltzer Yes, it shouldn't be in the API.

GrantSeltzer:
I think the check is failing because of vendoring, looking into it now, it's working locally

mrunalp:
Can you open a new PR with your commit cherry picked to see if we can make the travis issue go away?

wking:
On Wed, Jul 27, 2016 at 04:43:03PM -0700, Mrunal Patel wrote:
> Can you open a new PR with your commit cherry picked to see if we
> can make the travis issue go away?

New PR should not matter and this PR has already had its history
compacted.  Current failures are pointer issues [1], so we're probably
missing an update around #150.

[1]: https://travis-ci.org/opencontainers/ocitools/builds/147899681#L199

mrunalp:
@wking It builds fine locally.

GrantSeltzer:
@wking If you actually check the code you can see that it has been fixed but travis is not finding the correct commit.

wking:
On Wed, Jul 27, 2016 at 04:53:37PM -0700, Grant Seltzer Richman wrote:
> @wking If you actually check the code you can see that it has been
> fixed but travis is not finding the correct commit.

Ah, in that case pushing an amended commit (to bump the commit
timestamp and get a new hash) may work.  Although sometimes we've had
issues with Travis just being slow ;).  Opening a new PR is fine too
if we don't want to wait for Travis to catch up.

wking:
Actually, here's the merge commit Travis is testing *not* using a
pointer receiver [1].  So I don't think any “kick Travis” fix will
work.

[1]: https://github.com/opencontainers/ocitools/commit/5d4ed7ae2c6cd63944df987574ecde84db41b207#diff-3e0e09fa01dec94676136d9c0cf52025R322

wking:
And here's your current tip not using a pointer receiver:

  $ git grep addSeccomp d297bc1
  d297bc1:cmd/ocitools/generate.go:       g, err := addSeccomp(g, context)
  d297bc1:cmd/ocitools/generate.go:func addSeccomp(g generate.Generator, context *cli.Context) (generate.Generator, error) {

GrantSeltzer:
#159

wking:
On Wed, Jul 27, 2016 at 05:56:21PM -0700, Grant Seltzer Richman wrote:
> #159 

So we should close this one?

GrantSeltzer:
Yup @mrunalp

wking:
On Thu, Jul 28, 2016 at 07:48:10PM -0700, Grant Seltzer Richman wrote:
> Yup @mrunalp 

@GrantSeltzer, as the PR author, you should be able to close it too.

PR #129, opened 2016-07-05
generate: Add --output=PATH and write to stdout by default

wking:
This makes it easier to post-process output before writing it to disk.  For example, you may want to make adjustments with jq and/or [write part of the configuration to a different file for consumption by non-OCI tools][1].

[1]: https://github.com/opencontainers/ocitools/pull/128#issuecomment-230580568

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,11 @@ ocitools is a collection of tools for working with the [OCI runtime specificatio
 
 ## Generating an OCI runtime spec configuration files
 
-[`ocitools generate`][generate.1] generates a [`config.json`][config.json] for an [OCI bundle][bundle].
-This `config.json` file can be placed into a directory and used by an [OCI compatible runtime][runtime-spec] like [runC][] to run a container.
+[`ocitools generate`][generate.1] generates [configuration JSON][config.json] for an [OCI bundle][bundle].
+[OCI-compatible runtimes][runtime-spec] like [runC][] expect to read the configuration from `config.json`.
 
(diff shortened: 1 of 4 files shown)


liangchenye:
LGTM

Mashimiao:
+1

GrantSeltzer:
I think having the default output be a file should remain the same. Maybe there should be an option to output to stdout, and an option to to rename the file output? It could be the same flag too, either a path or "STDOUT" possibly.

wking:
On Wed, Jul 06, 2016 at 08:48:31AM -0700, Grant Seltzer Richman wrote:
> I think having the default output be a file should remain the same.

I like having ‘--output PATH’ always mean “please write to the file at
PATH”, and not having to rely on ‘--output /dev/stdout’ with the
/dev/stdout symlinks, /proc/self/fd/1, or magic paths like ‘-’.
Stdout is a generic Unix/C concept, and shouldn't require magic paths
or Linux-specific knowledge.

On the other hand, with the default-to-stdout approach, folks don't
need any special knowledge to say “I want to write this to
config.json”.  Folks who do a lot of this in interactive shells can
always alias ocitools-generate to ‘ocitools generate --output
config.json’, although I expect most ocitools consumers will be
scripts.

iamanush:
+1 I like the idea of defaulting to STDOUT and providing the --output flag for the file.

mrunalp:
LGTM

PR #131, opened 2016-07-09
Move generate code into a separate library

### LINKED ISSUE #130: Create a reusable library for ocitools generate
The generate code could be converted into a library that could be consumed in other projects.

hmeng-19:
Move generate code into a separate library, `libgen`. 

To avoid using `cli.Context` inside `libgen`, the `modify` function responds to get the command options, and transfer them into the functions inside `libgen`.

The PR tries to fix #130 .


hmeng-19:
@mrunalp , PTAL.

hmeng-19:
@mrunalp , @wking , this PR only aims to move the generate code into a separate library.  
We may do another PR to coordinate with #128 to make reviewing PRs easier.

generate.go
```diff
@@ -9,9 +9,8 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/Sirupsen/logrus"
+	"github.com/opencontainers/ocitools/libgen"
```
wking: There was a reasonable amount of “what should we call this?” discussion in #128, which resolved around [`seccomp-gen`][1].  If we're aiming for separate generate/validate/([test][1][?][2]) commands, we probably want something like:

    ocitools
    |-- cmd
    |    `-- ocitools
    |         |-- main.go
    |         |-- generate.go
    |         |-- validate.go
    |         …
    |-- generate
    |    |-- generate.go
    |    |-- utils.go, etc.
    |    `-- seccomp (#128)
    |         |-- consts.go
    |         …
    |-- validate
    |    |-- validate.go
    |    `-- utils.go
    …

Shifting the import paths around while we settle into whatever layout we choose isn't a big deal as long as we don't cut releases.  But I expect that the libraries will start picking up external consumers shortly after they're born, and we'll make life easier on those downstream folks if we don't churn the layout too much ;).

[1]: https://github.com/opencontainers/ocitools/pull/128#issuecomment-230688108
[2]: https://github.com/opencontainers/ocitools/pull/47#issuecomment-215328958
[3]: https://github.com/opencontainers/ocitools/pull/61#issuecomment-217910342

libgen/utils.go
```diff
@@ -0,0 +1,649 @@
...
+	cf, err := os.Open(path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, fmt.Errorf("template configuration at %s not found", path)
+		}
+	}
+	defer cf.Close()
+
+	if err = json.NewDecoder(cf).Decode(&spec); err != nil {
+		return
```
wking: I'm not sure how this compiles (although it seems to), but I expect it should be returning `nil, err`.

wking:
On Fri, Jul 08, 2016 at 08:02:51PM -0700, hmeng-19 wrote:
> We may do another PR to coordinate with #128 to make reviewing PRs
> easier.

This is also going to create some tedious conflicts with #54 (now
stable for almost two months ;).  Can we get an up/down ruling on that
PR in the near future.  Having these major PRs all touching the same
code and in flight at the same time is just going to create busywork.

libgen/utils.go
wking: And I see that this is just a copy/paste from the old code, but we might as well fix it now ;).

libgen/utils.go
```diff
@@ -0,0 +1,649 @@
...
+			Args = append(Args, &Arg)
+		} else {
+			return nil, fmt.Errorf("seccomp-sysctl args error: %s", argstr)
+		}
+	}
+	return Args, nil
+}
+
+// AddIDMappings add the mappings of uid and gid into spec.
+func AddIDMappings(spec *rspec.Spec, uidMaps, gidMaps []string) error {
```
wking: This should probably be a single:

    func ParseIDMappings(spec *rspec.Spec, idType string, maps []string) error {…}

which the command implementation can call with something like (untested, and I don't do much Go ;):

    idTypes := ["uid", "gid"]
    for idType := range idTypes {
        maps := context.StringSlice(idType + "mappings")
        if err := libgen.AddIDMappings(spec, idType, maps) err != nil {
          return err
        }
     }

The approach your copy/pasting was acceptable for an internal helper, but it seems like we should aim for something DRYer for a library method.  On the other hand, I'm fine cleaning this stuff up in follow-up PRs, as long as we don't wait long enough for API-breaking to become a concern.

libgen/utils.go
```diff
@@ -0,0 +1,649 @@
...
+	case "rshared":
+	default:
+		return fmt.Errorf("rootfs-propagation must be empty or one of private|rprivate|slave|rslave|shared|rshared")
+	}
+	spec.Linux.RootfsPropagation = rp
+	return nil
+}
+
+// AddHooks add the (prestart|poststop|poststart)hooks into spec.
+func AddHooks(spec *rspec.Spec, preStartHooks, postStopHooks, postStartHooks []string) error {
```
wking: I would like to DRY this up [too][1], but again, that can happen in a follow up PR.  On the other hand, shifting functions out into a separate package seems like a good time to put some thought into the API we want to expose.  Just moving things without cleaning up the APIs seems like it's breaking #128 and #54 without much benefit.

[1]: https://github.com/opencontainers/ocitools/pull/131/files#r70162128

libgen/utils.go
```diff
@@ -0,0 +1,649 @@
...
+			return
+		}
+	}
+}
+
+// StrPtr returns the pointer pointing to the string s.
+func StrPtr(s string) *string { return &s }
+
+// GetDefaultTemplate returns the default spec.
+func GetDefaultTemplate() *rspec.Spec {
```
wking: I'd put this in `default.go`.  And does Go have a deep copy?  If so, I'd consider making this a variable.  That way it's easy for folks to swap in an alternative default template or mutate the original if they want, and they can make a deep copy if they don't want to mutate the default template.

hmeng-19: @wking , I thought of putting this into `default.go`, and did not do it when I noticed that there are lots of maps, pointers, and slices embedded in the `rspec.Spec`.
I will try to see whether it is easy to do a deep copy of a `rspec.Spec` in Go.

libgen/utils.go
hmeng-19: @wking , reconsidering the API seems a good idea.
For this PR, I was just trying to separate `generate.go` and the library, and tried to avoid drastic changes of the code.
I will change the code after the new API is decided.

generate.go
hmeng-19: @wking , this new package structure looks very different from the current package structure.

    ocitools
    |-- cmd
    |    `-- runtimetest
    |         |-- main.go
    |         |-- rlimit_linux.go
    |         |-- mount
    |               |-- mountinfo_freebsd.go
    |               |-- mountinfo.go
    |               |-- mountinfo_linux.go
    |               |-- mountinfo_solaris.go
    |               |-- mountinfo_unsupported.go
    |               |-- mountinfo_windows.go
    |-- main.go
    |-- generate.go
    |-- validate.go

hmeng-19: @wking , @mrunalp , I think I will first get myself familiar with #54 , and #128 . 
Then we can come back to this PR and decide how we want it to be.

wking: On Mon, Jul 11, 2016 at 07:09:30AM -0700, hmeng-19 wrote:
> > @@ -9,9 +9,8 @@ import (
> >  	"strconv"
> >  	"strings"
> >  
> > -	"github.com/Sirupsen/logrus"
> > +	"github.com/opencontainers/ocitools/libgen"
>
> @wking , this new package structure looks very different from the
> current package structure…

Yeah, but if we're going to restructure things, I think we want to
take a big step to get to where we (think we) want to be, instead of
taking a bunch of short steps.  The big step may not be perfect, but
it's got a better chance at being right than a small step after which
we already plan on making further changes.

mrunalp: @wking proposed layout scheme looks reasonable to me.

libgen/utils.go
mrunalp: Named returns ;) But yes should be similar to other returns in the code for consistency.

mrunalp:
I agree with @wking that we should combine this with #128 so we have a single cohesive change/library.

generate.go
hmeng-19: @wking , @mrunalp . Sounds good. I will modify the code structure as @wking  suggested.
@wking , in your proposal, there are two `generate.go` - one is under `cmd/ocitools`, another is under `generate`. What is their difference? The same problem happens to `validate.go`.
I think all the files under `generate`  and `validate` should be libraries, right?

hmeng-19:
@mrunalp , I am okay with the plan. 
Should I also try to modify the code structure of the current `validate.go` file? If so, we probabaly should rename the title of #130 to be more generic.

generate.go
wking: On Mon, Jul 11, 2016 at 11:36:11AM -0700, hmeng-19 wrote:
> @wking , in your proposal, there are two `generate.go` - one is
> under `cmd/ocitools`, another is under `generate`. What is their
> difference? The same problem happens to `validate.go`.

The cmd/ocitools stuff is just thin bindings between the library code
(under generate/, validate/, etc.) and urfave/cli.

> I think all the files under `generate` and `validate` should be
> libraries, right?

Yup.

hmeng-19:
@mrunalp , @wking , the CI checks failed because this PR tries to introduce a new package `cmd/ocitools`:
 
    The command "$HOME/gopath/bin/git-validation -run DCO,short-subject -v -range ${TRAVIS_COMMIT_RANGE}" exited with 0.
    0.01s$ make
    go build -tags "" -o ocitools ./cmd/ocitools
    can't load package: package github.com/opencontainers/ocitools/cmd/ocitools: cannot find package "github.com/opencontainers/ocitools/cmd/ocitools" in any of:
        /home/travis/.gimme/versions/go1.5.3.linux.amd64/src/github.com/opencontainers/ocitools/cmd/ocitools (from $GOROOT)
        /home/travis/gopath/src/github.com/opencontainers/ocitools/Godeps/_workspace/src/github.com/opencontainers/ocitools/cmd/ocitools (from $GOPATH)
        /home/travis/gopath/src/github.com/opencontainers/ocitools/cmd/ocitools
    make: *** [all] Error 1

hmeng-19:
@mrunalp , @wking , the new changes I made include:
 
    rename libgen -> generate
    move package main into cmd/ocitools
    Add an empty generate/seccomp package
  
The new changes focus on the package structure, does not really polish the function API. 
I think we can merge this PR to unblock #128. Then we can make a new PR to polish the new function API.

wking:
On Mon, Jul 11, 2016 at 02:37:44PM -0700, hmeng-19 wrote:
> @mrunalp , @wking , the CI checks failed because this PR tries to
> introduce a new package `cmd/ocitools`:
>  
> …
>     can't load package: package github.com/opencontainers/ocitools/cmd/ocitools: cannot find package "github.com/opencontainers/ocitools/cmd/ocitools" in any of:
>         /home/travis/.gimme/versions/go1.5.3.linux.amd64/src/github.com/opencontainers/ocitools/cmd/ocitools (from $GOROOT)
>         /home/travis/gopath/src/github.com/opencontainers/ocitools/Godeps/_workspace/src/github.com/opencontainers/ocitools/cmd/ocitools (from $GOPATH)
>         /home/travis/gopath/src/github.com/opencontainers/ocitools/cmd/ocitools

Actually, I think that's failing because you haven't added
cmd/ocitools [1].  Did you forget a ‘git add cmd/ocitools’ with
d2f10ede?

hmeng-19:
@wking , you are right. My `.gitignore` confused me and I thought I added those files.
Fix it. Thanks. PTAL.

wking:
On Mon, Jul 11, 2016 at 02:55:12PM -0700, hmeng-19 wrote:
> My `.gitignore` confused me and I thought I added those files.

Ah, the ‘ocitools’ entry in .gitignore should probably be ‘/ocitools’
to anchor the ignore (and not ignore cmd/ocitools).  9777ceb made a
similar fix for runtimetest.

I'm not sure what the ‘oci’ entry in .gitignore is for.  It's from
7ac06868 (2015-09-15), and maybe is just stale?

hmeng-19:
@wking , I think the problem is the pattern `ocitools` in the `.gitignore` file, which automatically
ignore all the files or directories with the name of `ocitools`. In our case, `cmd/ocitools`, together with all the entries under it, gets ignored.
However, by adding a leading slash into a pattern, it would only match the beginning of the pathname. That is how Mrunal solved the problem in 9777ceb473cf23005ac74f177d83fc7398845a7b. `/runtimetest` only matches `runtimetest` directly under the git repo.

Please refer to here https://git-scm.com/docs/gitignore .

hmeng-19:
@wking , @mrunalp , fix the `.gitignore` file. PTAL.

wking:
Through fed976fa6 look fine to me if we want to shuffle things around
without changing APIs (much), although I'd rather it was squashed down
into a single commit.

But I don't see a need to split “restructure and reconsider the API”
into two separate PRs [1].  And I think it would be nice if #54 and
#128 (which were both filed before this change was proposed in #130)
got reviewed first.

[1]: https://github.com/opencontainers/ocitools/pull/131#discussion_r70261507

GrantSeltzer:
@hmeng-19 and I are in agreement that this PR should get merged before #128, and i'll probably close #128 and make a new one that will add all of the seccomp stuff into the generate library that she is separating out.

wking:
On Tue, Jul 12, 2016 at 07:13:44AM -0700, Grant Seltzer Richman wrote:
> @hmeng-19 and I are in agreement that this PR should get merged
> before #128, and i'll probably close #128 and make a new one that
> will add all of the seccomp stuff into the generate library that she
> is separating out.

Whatever's easier, but you can rebase your existing #128 branch once
this PR lands.  And if the plan is to land this PR first, I'll survive
a rebase of #54 (but it would be nice if PRs were reviewed and
merged/rejected more quickly to avoid these parallel-work conflicts.

cmd/ocitools/generate.go
```diff
@@ -0,0 +1,244 @@
...
+	cli.StringSliceFlag{Name: "label", Usage: "add annotations to the configuration e.g. key=value"},
+}
+
+var generateCommand = cli.Command{
+	Name:   "generate",
+	Usage:  "generate a OCI spec file",
+	Flags:  generateFlags,
+	Before: before,
+	Action: func(context *cli.Context) error {
+		spec := generate.GetDefaultTemplate()
```
mrunalp: This could be `generate.New()`

cmd/ocitools/generate.go
```diff
@@ -0,0 +1,244 @@
...
+		template := context.String("template")
+		if template != "" {
+			var err error
+			spec, err = generate.LoadTemplate(template)
```
mrunalp: `generate.NewFromTemplate()`

mrunalp:
commit history needs to be cleaned up or they should be squashed.

cmd/ocitools/generate.go
hmeng-19: @mrunalp , are you suggesting move the following code into a new func `generate.NewFromTemplate`?

    template := context.String("template")
    if template != "" {
        var err error
        spec, err = generate.LoadTemplate(template)

hmeng-19: @mrunalp , should we name it `NewSpecFromTemplate`?

mrunalp: Yeah, I am suggesting that we rename it.

cmd/ocitools/generate.go
mrunalp: Again, a rename suggestion for `GetDefaultTemplate` to `New`

hmeng-19:
@mrunalp , @wking , renamed the two functions you suggested, and squashed all my changes into a single commit. 
PTAL.

mrunalp:
While this is a step in the right direction,  I think API like the following will be easier to use:

```go

type generator struct {
    spec *rspec.Spec
}

// Start from default template
specgen := generator.New()

// Make modifications
specgen.SetRootPath("/path/to/rootfs")
specgen.AddCapabilities(...)
specgen.AddTmpfsMount(...)
specgen.SetSelinuxLabel(...)
...
...
// Write to disk
specgen.Save("/path/to/file")

spec := specgen.GetSpec()

// Load a template.json
specgen := generator.NewFromTemplate("/path/to/file")
specgen.AddMount(src, dest, options..)
specgen.Save("/some/other/path")

// Load existing spec object and modify it to generate a new spec
specgen := generator.NewFromSpec(someSpecObject)
specgen.SetNoNewPriviliges()
specgen.Save("/yet/another/path")
```

wking:
On Wed, Jul 13, 2016 at 02:53:54PM -0700, Mrunal Patel wrote:
> specgen := generator.NewFromTemplate("/path/to/file")

I'd load this from an io.Reader for more flexibility.  Wrapping that
in something that opened a file, deferred a close, and passed the
writer in to generator.NewFromReader is acceptable noise for the
command-line binding package.

mrunalp:
Also, we should have both Set and Add for items that are lists (like Capabilities and Namespaces for e.g.). That allows one to reset the list or set it wholesale using Set and add items to it using Add. Drop should also be added for Capabilities and Namespaces to begin with.

hmeng-19:
@mrunalp , @wking , sounds good.

hmeng-19:
@mrunalp , @wking , PTAL.
I rebased the PR on top of the latest master. 
All the new change since the last review is in the second commit https://github.com/opencontainers/ocitools/pull/131/commits/5b4ed54f15dc2f39d396d36a67f35aaaa7d12490 .

hmeng-19:
@mrunalp , @wking ,  ignore the last message.
I still missed some `Add` function. Will fix it soon.

cmd/ocitools/generate.go
```diff
@@ -0,0 +1,299 @@
...
+		template := context.String("template")
+		if template != "" {
+			var err error
+			specgen, err = generate.NewFromFile(template)
+			if err != nil {
+				return err
+			}
+		}
+
+		err := modify(specgen, context)
```
mrunalp: We should rename `modify` to `setupSpec` or something similar.

cmd/ocitools/generate.go
```diff
@@ -0,0 +1,299 @@
...
+	},
+}
+
+func modify(g generate.Generator, context *cli.Context) error {
+	spec := g.GetSpec()
+
+	if len(spec.Version) == 0 {
+		spec.Version = rspec.Version
+	}
+	spec.Root.Path = context.String("rootfs")
```
mrunalp: The code from here through line 127 should all have functions in the generator.

cmd/ocitools/generate.go
hmeng-19: Sounds good. Work on it.

cmd/ocitools/generate.go
hmeng-19: @mrunalp , Okay. In general, we can provide get/set (plus add if applicable) functions for each field inside g.spec. Right?

hmeng-19:
@mrunalp , just let know. I am still trying to polishing each function. The current commit is not the final one.

cmd/ocitools/generate.go
mrunalp: We don't need get except for spec

hmeng-19: Yes. I think you are right. Realized it when I am fixing it.

cmd/ocitools/generate.go
```diff
@@ -0,0 +1,345 @@
...
+
+	if context.IsSet("sysctl") {
+		g.SetLinuxSysctl(context.StringSlice("sysctl"))
+	}
+
+	privileged := false
+	if context.IsSet("privileged") {
+		privileged = context.Bool("privileged")
+	}
+	g.SetupPrivilege(privileged)
```
mrunalp: `SetupPrivilege` -> `SetPrivileged`

cmd/ocitools/generate.go
```diff
@@ -0,0 +1,345 @@
...
+
+	if context.IsSet("cap-add") {
+		addCaps := context.StringSlice("cap-add")
+		if err := g.AddProcessCapabilities(addCaps); err != nil {
```
mrunalp: `AddProcessCapabilities` -> `AddProcessCapability`

cmd/ocitools/generate.go
hmeng-19: Accept.

cmd/ocitools/generate.go
hmeng-19: Accept

cmd/ocitools/generate.go
```diff
@@ -0,0 +1,345 @@
...
+	}
+
+	mountCgroupOption := context.String("mount-cgroups")
+	if err := g.AddCgroupsMounts(mountCgroupOption); err != nil {
+		return err
+	}
+
+	if context.IsSet("bind") {
+		binds := context.StringSlice("bind")
+		if err := g.AddBindMounts(binds); err != nil {
```
mrunalp: API should add one bind mount at a time and should be called AddBindMount

cmd/ocitools/generate.go
```diff
@@ -0,0 +1,345 @@
...
+			return err
+		}
+	}
+
+	if context.IsSet("prestart") {
+		preStartHooks := context.StringSlice("prestart")
+		if err := g.AddPreStartHooks(preStartHooks); err != nil {
```
mrunalp: Same here; one at a time.

cmd/ocitools/generate.go
hmeng-19: Is it okay we expect the user to put a single bind mount setting into a slice?
Another I can do is use variadic function.

mrunalp: We can loop in the calling code to add multiple bind mounts.

hmeng-19: Sounds good.

cmd/ocitools/generate.go
```diff
@@ -0,0 +1,345 @@
...
+	for _, nsName := range generate.Namespaces {
+		if context.IsSet(nsName) {
+			nsMaps[nsName] = context.String(nsName)
+		}
+	}
+	setupLinuxNamespaces(g, needsNewUser, nsMaps)
+
+	if context.IsSet("tmpfs") {
+		tmpfsSlice := context.StringSlice("tmpfs")
+		if err := g.AddTmpfsMounts(tmpfsSlice); err != nil {
```
mrunalp: Same here. This should be `AddTmpfsMount`

cmd/ocitools/generate.go
```diff
@@ -0,0 +1,345 @@
...
+	if context.IsSet("prestart") {
+		preStartHooks := context.StringSlice("prestart")
+		if err := g.AddPreStartHooks(preStartHooks); err != nil {
+			return err
+		}
+	}
+
+	if context.IsSet("poststop") {
+		postStopHooks := context.StringSlice("poststop")
+		if err := g.AddPostStopHooks(postStopHooks); err != nil {
```
mrunalp: Same for hooks.

hmeng-19:
@mrunalp , I am fixing all these problems. The new commit is coming soon.

generate/utils.go
```diff
@@ -0,0 +1,1147 @@
+// Package generate implements functions generating container config files.
```
mrunalp: rename this file to generate.go?

generate/utils.go
```diff
@@ -0,0 +1,1147 @@
...
+		if ns.Type == namespace.Type {
+			g.spec.Linux.Namespaces = append(g.spec.Linux.Namespaces[:i], g.spec.Linux.Namespaces[i+1:]...)
+			return nil
+		}
+	}
+	return nil
+}
+
+// StrPtr returns the pointer pointing to the string s.
+func StrPtr(s string) *string { return &s }
```
mrunalp: this function could be private to the generate package.

generate/utils.go
hmeng-19: Okay.

generate/utils.go
hmeng-19: The validate.go is currently using this function once.

wking: On Fri, Jul 15, 2016 at 03:46:51PM -0700, hmeng-19 wrote:
> > +// StrPtr returns the pointer pointing to the string s.
> > +func StrPtr(s string) *string { return &s }
> 
> The validate.go is currently using this function once.

ocitools/util.go ;).

hmeng-19: Sounds good. Fixing it.

hmeng-19:
@mrunalp , @wking , I am finishing my change on my side for now, and also rebase and squash my commit.
Do you want to push my changes to this PR now? or you guys want to finish review the current version?

mrunalp:
Please push your updates so the latest version could be reviewed.

generate/utils.go
hmeng-19: I may remember the wrong variable. `strPtr` is only used by the generate library and can be private.

hmeng-19:
@mrunalp , @wking , the latest PR is in.

generate/generate.go
```diff
@@ -0,0 +1,1015 @@
...
+	idMapping := rspec.IDMapping{
+		HostID:      uint32(hid),
+		ContainerID: uint32(cid),
+		Size:        uint32(size),
+	}
+	return idMapping, nil
+}
+
+// ClearLinuxUIDMapping clears g.spec.Linux.UIDMappings.
+func (g Generator) ClearLinuxUIDMapping() {
```
mrunalp: ClearLinuxUIDmappings

generate/generate.go
```diff
@@ -0,0 +1,1015 @@
...
+	if err != nil {
+		return err
+	}
+
+	g.spec.Linux.UIDMappings = append(g.spec.Linux.UIDMappings, r)
+	return nil
+}
+
+// ClearLinuxGIDMapping clears g.spec.Linux.GIDMappings.
+func (g Generator) ClearLinuxGIDMapping() {
```
mrunalp: ClearLinuxGIDMappings

generate/generate.go
```diff
@@ -0,0 +1,1015 @@
...
+		return rspec.Namespace{Type: rspec.UserNamespace, Path: path}, nil
+	case "cgroup":
+		return rspec.Namespace{Type: rspec.CgroupNamespace, Path: path}, nil
+	default:
+		return rspec.Namespace{}, fmt.Errorf("Should not reach here!")
+	}
+}
+
+// ClearLinuxNamespace clears g.spec.Linux.Namespaces.
+func (g Generator) ClearLinuxNamespace() {
```
mrunalp: ClearLinuxNamespaces

generate/generate.go
```diff
@@ -0,0 +1,1015 @@
...
+	args := []string{}
+	path := parts[0]
+	if len(parts) > 1 {
+		args = parts[1:]
+	}
+	return rspec.Hook{Path: path, Args: args}
+}
+
+// ClearPreStartHook clears g.spec.Hooks.Prestart.
+func (g Generator) ClearPreStartHook() {
```
mrunalp: ClearPreStartHooks

generate/generate.go
```diff
@@ -0,0 +1,1015 @@
...
+			},
+			Rlimits: []rspec.Rlimit{
+				{
+					Type: "RLIMIT_NOFILE",
+					Hard: uint64(1024),
+					Soft: uint64(1024),
+				},
+			},
+		},
+		Hostname: "shell",
```
mrunalp: This should be something else. Maybe `example`.

wking: On Fri, Jul 15, 2016 at 04:25:19PM -0700, Mrunal Patel wrote:
> > +		Hostname: "shell",
> 
> This should be something else. Maybe `example`.

I usually echo the spec's example [1].

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc1/config.md#L188

hmeng-19:
@mrunalp , @wking , fixed the problems you pointed out in a separate commit https://github.com/opencontainers/ocitools/pull/131/commits/0e6dfc00c26a3ca89008b81eb08684b370908ad7 . PTAL.

liangchenye:
0e6dfc0  LGTM

generate/generate.go
```diff
@@ -0,0 +1,1015 @@
...
+		g.spec.Process.Capabilities = finalCapList
+		g.spec.Process.SelinuxLabel = ""
+		g.spec.Process.ApparmorProfile = ""
+		g.spec.Linux.Seccomp = nil
+	}
+}
+
+func checkCap(c string) error {
+	isValid := false
+	cp := fmt.Sprintf("CAP_%s", c)
```
mrunalp: This isn't required. The list you are checking against doesn't have CAP_

generate/generate.go
```diff
@@ -0,0 +1,1015 @@
...
+
+	for _, cap := range capability.List() {
+		if strings.ToUpper(cp) == strings.ToUpper(cap.String()) {
```
mrunalp: Move this out of the loop to avoid calling ToUpper in each iteration for cp

mrunalp:
--seccomp-default doesn't get populated
```
[root@dhcp-16-129 busybox]# ocitools generate --tty --output=config.json --seccomp-default SCMP_ACT_ALLOW --seccomp-errno getcwd
[root@dhcp-16-129 busybox]# tail -n 20 config.json 
                        {
                                "type": "mount"
                        }
                ],
                "seccomp": {
                        "defaultAction": "",
                        "architectures": null,
                        "syscalls": [
                                {
                                        "name": "getcwd",
                                        "action": "SCMP_ACT_ERRNO"
                                }
                        ]
                }
        },
        "solaris": {
                "cappedCPU": {},
                "cappedMemory": {}
        }
```

mrunalp:
Also trying to add a capability or dropping it fails
Try --cap-add SYS_ADMIN. I have pointed to the problem in earlier comment.

generate/generate.go
hmeng-19: The list I am checking against is `capability.List()`, which includes `CAP_CHOWN`, `CAP_DAC_OVERRIDE`, and so on.

hmeng-19: Ignore my comment. I got your point.

hmeng-19:
@mrunalp , fixed the problems you pointed out and also tested all the other options, PTAL.

mrunalp:
Can you re-order or squash the commits?

hmeng-19:
@mrunalp , no problem. Are you done checking the latest changes?

mrunalp:
@hmeng-19 Yeah, the changes look good and am ready to merge.

hmeng-19:
@mrunalp , I sqashed the commits.

mrunalp:
LGTM

PR #135, opened 2016-07-18
runtimetest: Fix check for device nodes

mrunalp:

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -283,7 +283,7 @@ func validateDefaultDevices(spec *rspec.Spec) error {
 				return err
 			}
 		} else {
-			if fi.Mode() != os.ModeDevice {
+			if fi.Mode()&os.ModeDevice != os.ModeDevice {
 				return fmt.Errorf("file %v is not a device as expected", device)
 			}
 		}


test_runtime.sh
```diff
@@ -78,7 +78,7 @@ tar -xf  rootfs.tar.gz -C ${TESTDIR}
 cp runtimetest ${TESTDIR}
 
 pushd $TESTDIR > /dev/null
-ocitools generate --output config.json "${TEST_ARGS[@]}" --rootfs '.'
+ocitools generate --tty --output config.json "${TEST_ARGS[@]}" --rootfs '.'
```
wking: Drive by?

wking:
7fd96c123's ‘& os.ModeDevice’ addition looks good to me, but I'm not
sure what the --tty addition is for.

test_runtime.sh
mrunalp: /dev/console is only setup if we set terminal to true.

wking: On Mon, Jul 18, 2016 at 03:41:25PM -0700, Mrunal Patel wrote:
> > -ocitools generate --output config.json "${TEST_ARGS[@]}" --rootfs '.'
> > +ocitools generate --tty --output config.json "${TEST_ARGS[@]}" --rootfs '.'
> 
> /dev/console is only setup if we set terminal to true. 

Can we stop adding “runC isn't compliant out of the box” workarounds
now that we're post 1.0.0-rc1?  If there's really a reason to not
setup /dev/console when terminal is false, can we PR the spec with
that information?

mrunalp: @wking I have created a runtime-spec PR https://github.com/opencontainers/runtime-spec/pull/518

wking: On Mon, Jul 18, 2016 at 03:54:39PM -0700, Mrunal Patel wrote:
> > @@ -78,7 +78,7 @@ tar -xf  rootfs.tar.gz -C ${TESTDIR}
> >  cp runtimetest ${TESTDIR}
> >  
> >  pushd $TESTDIR > /dev/null
> > -ocitools generate --output config.json "${TEST_ARGS[@]}" --rootfs '.'
> > +ocitools generate --tty --output config.json "${TEST_ARGS[@]}" --rootfs '.'
> 
> @wking I have created a runtime-spec PR https://github.com/opencontainers/runtime-spec/pull/518

+1.  In the meantime, can we land this PR without this change?

mrunalp: I have taken out that change for now pending upstream PR.

wking:
6f9de0a0 looks good to me.

The --tty bit should be easy to add back in if we still need it after
azurestandard/runtime-spec#518 gets sorted.

PR #137, opened 2016-07-21
Add resource set funcs into generate library

hmeng-19:
The PR adds functions setting the linux resource cpu and memory fields into the generate library.

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -342,6 +342,71 @@ func (g Generator) SetLinuxMountLabel(label string) {
 	g.spec.Linux.MountLabel = label
 }
 
+// SetLinuxResourcesCPUShares sets g.spec.Linux.Resources.CPU.Shares.
+func (g Generator) SetLinuxResourcesCPUShares(shares uint64) {
+	g.spec.Linux.Resources.CPU.Shares = &shares
+}
+
+// SetLinuxResourcesCPUQuota sets g.spec.Linux.Resources.CPU.Quota.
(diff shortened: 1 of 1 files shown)


hmeng-19:
@mrunalp , PTAL.
This PR only adds the library funcs, and does not expand the ocitools generate command arguments.

I plan to expand the ocitools generate in a separate  issue and PR.

mrunalp:
LGTM

wking:
I'm too late to beat the merge here, but do we really prefer these
one-line setters to:

  type Generator struct {
    Spec *rspec.Spec
  }

and letting folks touch g.Spec directly?

mrunalp:
@wking Good point and I have thought about it and gone back and forth.
One option is for the user to get back the spec using GetSpec and then modify it directly. For now adding the most common options seems fine. If it becomes too many we can reconsider the approach.

hmeng-19:
@mrunalp , @wking , I indeed felt the pain when I tried to add some func like `SetLinuxResourcesCPURealtimePeriod` into the generate library. 
I kind of want to make the func name corresponding to the hierarchy of the runtime-spec config, but it looks it is a little too much. 
Let me know if you have some good suggestions for this.

wking:
On Thu, Jul 21, 2016 at 01:57:41PM -0700, Mrunal Patel wrote:
> One option is for the user to get back the spec using GetSpec and
> then modify it directly.

Just make the spec public [1].  No need for a getter unless you want
to do something to the private content or perform a auth check or some
such on the way out.

PR #138, opened 2016-07-21
Check CAP_LAST_CAP while setting privileged

hmeng-19:
The PR checks whether each cap inside `capability.List()` is <= `capability.CAP_LAST_CAP` before adding it into `finalCapList`.
This avoids add unsupported cap into `finalCapList`. For example, `CAP_AUDIT_READ` is added into  the kernel since Linux 3.16, and not supported on RHEL7.

### DIFF
--- a/cmd/ocitools/generate.go
+++ b/cmd/ocitools/generate.go
@@ -78,7 +78,7 @@ var generateCommand = cli.Command{
 			}
 		}
 
-		err := setupSpec(specgen, context)
+		err := setupSpec(&specgen, context)
 		if err != nil {
 			return err
 		}
@@ -96,8 +96,12 @@ var generateCommand = cli.Command{
 	},
 }
 
-func setupSpec(g generate.Generator, context *cli.Context) error {
-	spec := g.GetSpec()
+func setupSpec(g *generate.Generator, context *cli.Context) error {
+	if context.GlobalBool("host-specific") {
+		g.HostSpecific = true
+	}
+
+	spec := g.Spec()
 
 	if len(spec.Version) == 0 {
 		g.SetVersion(rspec.Version)
@@ -369,7 +373,7 @@ func checkNs(nsMaps map[string]string, nsName string) bool {
 	return true
 }
 
-func setupLinuxNamespaces(g generate.Generator, needsNewUser bool, nsMaps map[string]string) {
+func setupLinuxNamespaces(g *generate.Generator, needsNewUser bool, nsMaps map[string]string) {
 	for _, nsName := range generate.Namespaces {
 		if !checkNs(nsMaps, nsName) && !(needsNewUser && nsName == "user") {
 			continue
--- a/cmd/ocitools/validate.go
+++ b/cmd/ocitools/validate.go
@@ -24,7 +24,6 @@ type configCheck func(rspec.Spec, string, bool) []string
 
 var bundleValidateFlags = []cli.Flag{
 	cli.StringFlag{Name: "path", Value: ".", Usage: "path to a bundle"},
-	cli.BoolFlag{Name: "host-specific", Usage: "Check host specific configs."},
 }
 
 var (
@@ -99,7 +98,7 @@ var bundleValidateCommand = cli.Command{
 			return fmt.Errorf("The root path %q is not a directory.", rootfsPath)
 		}
 
-		hostCheck := context.Bool("host-specific")
+		hostCheck := context.GlobalBool("host-specific")
 
 		checks := []configCheck{
 			checkMandatoryFields,
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -21,7 +21,8 @@ var (
 
 // Generator represents a generator for a container spec.
 type Generator struct {
-	spec *rspec.Spec
+	spec         *rspec.Spec
+	HostSpecific bool
 }
 
 // New creates a spec Generator with the default spec.
@@ -139,12 +140,16 @@ func New() Generator {
 			Devices: []rspec.Device{},
 		},
 	}
-	return Generator{&spec}
+	return Generator{
+		spec: &spec,
+	}
 }
 
 // NewFromSpec creates a spec Generator from a given spec.
 func NewFromSpec(spec *rspec.Spec) Generator {
-	return Generator{spec}
+	return Generator{
+		spec: spec,
+	}
 }
 
 // NewFromFile loads the template specifed in a file into a spec Generator.
@@ -166,7 +171,9 @@ func NewFromTemplate(r io.Reader) (Generator, error) {
 	if err := json.NewDecoder(r).Decode(&spec); err != nil {
 		return Generator{}, err
 	}
-	return Generator{&spec}, nil
+	return Generator{
+		spec: &spec,
+	}, nil
 }
 
 // SetSpec sets the spec in the Generator g.
@@ -174,8 +181,8 @@ func (g *Generator) SetSpec(spec *rspec.Spec) {
 	g.spec = spec
 }
 
-// GetSpec gets the spec in the Generator g.
-func (g *Generator) GetSpec() *rspec.Spec {
+// Spec gets the spec in the Generator g.
+func (g *Generator) Spec() *rspec.Spec {
 	return g.spec
 }
 
@@ -953,6 +960,9 @@ func (g *Generator) SetupPrivileged(privileged bool) {
 		// Add all capabilities in privileged mode.
 		var finalCapList []string
 		for _, cap := range capability.List() {
+			if g.HostSpecific && cap > capability.CAP_LAST_CAP {
+				continue
+			}
 			finalCapList = append(finalCapList, fmt.Sprintf("CAP_%s", strings.ToUpper(cap.String())))
 		}
 		g.initSpecLinux()
@@ -963,12 +973,15 @@ func (g *Generator) SetupPrivileged(privileged bool) {
 	}
 }
 
-func checkCap(c string) error {
+func checkCap(c string, hostSpecific bool) error {
 	isValid := false
 	cp := strings.ToUpper(c)
 
 	for _, cap := range capability.List() {
 		if cp == strings.ToUpper(cap.String()) {
+			if hostSpecific && cap > capability.CAP_LAST_CAP {
+				return fmt.Errorf("CAP_%s is not supported on the current host", cp)
+			}
 			isValid = true
 			break
 		}
@@ -990,7 +1003,7 @@ func (g *Generator) ClearProcessCapabilities() {
 
 // AddProcessCapability adds a process capability into g.spec.Process.Capabilities.
 func (g *Generator) AddProcessCapability(c string) error {
-	if err := checkCap(c); err != nil {
+	if err := checkCap(c, g.HostSpecific); err != nil {
 		return err
 	}
 
@@ -1009,7 +1022,7 @@ func (g *Generator) AddProcessCapability(c string) error {
 
 // DropProcessCapability drops a process capability from g.spec.Process.Capabilities.
 func (g *Generator) DropProcessCapability(c string) error {
-	if err := checkCap(c); err != nil {
+	if err := checkCap(c, g.HostSpecific); err != nil {
 		return err
 	}
 
--- a/man/ocitools.1.md
+++ b/man/ocitools.1.md
@@ -15,10 +15,25 @@ ocitools is a collection of tools for working with the [OCI runtime specificatio
 
 # OPTIONS
 **--help**
-  Print usage statement
+  Print usage statement.
 
 **-v**, **--version**
   Print version information.
+
+**--log-level**
+  Log level (panic, fatal, error, warn, info, or debug) (default: "error").
+
+**--host-specific**
+  Generate host-specific configs or do host-specific validations.
+
+  By default, generator generates configs without checking whether they are
+  supported on the current host. With this flag, generator will first check
+  whether each config is supported on the current host, and only add it into
+  the config file if it passes the checking.
+
+  By default, validation only tests for compatibility with a hypothetical host.
+  With this flag, validation will also run more specific tests to see whether
+  the current host is capable of launching a container from the configuration.
 
 # COMMANDS
 **validate**
--- a/cmd/ocitools/main.go
+++ b/cmd/ocitools/main.go
@@ -18,6 +18,10 @@ func main() {
 			Value: "error",
 			Usage: "Log level (panic, fatal, error, warn, info, or debug)",
 		},
+		cli.BoolFlag{
+			Name:  "host-specific",
+			Usage: "generate host-specific configs or do host-specific validations",
+		},
 	}
 
 	app.Commands = []cli.Command{
--- a/man/ocitools-validate.1.md
+++ b/man/ocitools-validate.1.md
@@ -18,16 +18,6 @@ Validate an OCI bundle
 **--path=PATH
   Path to bundle
 
-**--host-specific**
-  Check host specific configs.
-  By default, validation only tests for compatibility with a hypothetical host.
-  With this flag, validation will also run more specific tests to see whether
-  the current host is capable of launching a container from the configuration.
-  For example, validating a compliant Windows configuration on a Linux machine
-  will pass without this flag ("there may be a Windows host capable of
-  launching this container"), but will fail with it ("this host is not capable
-  of launching this container").
-
 # SEE ALSO
 **ocitools**(1)
 


hmeng-19:
@mrunalp , PTAL.

wking:
On Thu, Jul 21, 2016 at 09:39:36AM -0700, hmeng-19 wrote:
> The PR checks whether each cap inside `capability.List()` is <=
> `capability.CAP_LAST_CAP` before adding it into `finalCapList`.

This is host-specific (e.g. maybe you're using a host with an old
kernel to build a container for a host with a new kernel).  ‘validate’
has a --host-specific flag to cover this sort of thing [1].  Maybe we
should promote that to a global option?

[1]: https://github.com/opencontainers/ocitools/blame/e05b84720adfb66c34152ace2d554a82174ad631/man/ocitools-validate.1.md#L21

hmeng-19:
@wking , I am testing octools on RHEL7 recently. 
After running the following command, `CAP_AUDIT_READ` was also put into `config.json`. However, rhel7 does not support this cap, because its kernel is 3.10.

    ocitools generate --privileged --tty --output=config.json --uidmappings=0:0:1024 --gidmappings=0:0:1024

wking:
On Thu, Jul 21, 2016 at 09:49:08AM -0700, hmeng-19 wrote:
> After running the following command, `CAP_AUDIT_READ` was also put
> into `config.json`. However, rhel7 does not support this cap,
> because its kernel is 3.10.

That's only a problem if you're also trying to *run* the container on
the same kernel that was used to generate the config.  A system
running Linux 3.10 is capable of *generating* a config that uses
CAP_AUDIT_READ (or a Solaris config, or a Windows config).  It just
can't run containers using those configs.

A --host-specific flag says “Don't be generic, I want to
generate/validate for the current host”.  For you, that would end up
looking like:

  $ ocitools --host-specific generate --privileged --tty --output=config.json --uidmappings=0:0:1024 --gidmappings=0:0:1024

hmeng-19:
@wking , sounds good. It would be nice we can take care both cases:
1) create config.json  with `ocitools` and use the config.json with `runc` on the same machine;
2) create config.json  with `ocitools` and use the config.json with `runc` on different machines;

It is easy to picture both use cases.

wking:
On Thu, Jul 21, 2016 at 10:01:54AM -0700, hmeng-19 wrote:
> It would be nice we can take care both cases:

Agreed, that's why I'm suggesting a global option ;).

hmeng-19:
@mrunalp , how do you think about this?

mrunalp:
Sounds good to me.

hmeng-19:
@mrunalp , @wking , PTAL.

cmd/ocitools/validate.go
```diff
@@ -84,7 +84,7 @@ var bundleValidateCommand = cli.Command{
 			return fmt.Errorf("The root path %q is not a directory.", rootfsPath)
 		}
 
-		hostCheck := context.Bool("host-specific")
+		hostCheck := hostSpecific || context.Bool("host-specific")
```
wking: I don't think we want both a global and a per-command --host-specific here.  We should drop the per-command option.

And I'm not all that familiar with urfave/cli, but do we really need the package-global `hostSpecific`?  I'd expect you to be able to pull that from the `context`, even though it's set via a global variable.

hmeng-19: > I don't think we want both a global and a per-command --host-specific here. We should drop the per-command option.

 I am okay with drop the per-command --host-specific option once @mrunalp  and you are okay with it. 
I keep it for now just in case some important users are using it.

hmeng-19: > And I'm not all that familiar with urfave/cli, but do we really need the package-global hostSpecific? I'd expect you to be able to pull that from the context, even though it's set via a global variable.

I use a global var here to avoid multiple func calls to `context.GlobalBool`.

cmd/ocitools/validate.go
```diff
@@ -109,10 +109,9 @@ var bundleValidateCommand = cli.Command{
 			errMsg = fmt.Sprintf("%d Errors detected:\n%s", i-1, errMsg)
 			return errors.New(errMsg)
 
-		} else {
-			fmt.Println("Bundle validation succeeded.")
-			return nil
 		}
+		fmt.Println("Bundle validation succeeded.")
+		return nil
```
wking: +1 to this change, but it probably belongs in a separate commit.

cmd/ocitools/validate.go
```diff
@@ -270,10 +269,9 @@ func supportedMountTypes(OS string, hostCheck bool) (map[string]bool, error) {
 		supportedTypes["bind"] = true
 
 		return supportedTypes, nil
-	} else {
-		logrus.Warn("Checking linux mount types without --host-specific is not supported yet")
-		return nil, nil
 	}
+	logrus.Warn("Checking linux mount types without --host-specific is not supported yet")
+	return nil, nil
```
wking: Also +1, but a separate commit (maybe a “drop superfluous elses” PR?).

hmeng-19: Sounds good. Will fix it.

cmd/ocitools/validate.go
wking: Ah, [here][1]'s an example of accessing a global option.

[1]: https://github.com/opencontainers/ocitools/pull/138/files#diff-46eb56a79aaa28b90fd1b953e23bc8c9R40

wking: On Thu, Jul 21, 2016 at 12:31:53PM -0700, hmeng-19 wrote:
> I keep it for now just in case some important users are using it. 

We haven't tagged an ocitools release yet or said anything about
forward compat in it's command-line API, so I think we're free to
change whatever we want.  But if we know of any consumers, giving them
a heads up on the change would be good (ideally as part of a
release-tagging process ;).

wking: On Thu, Jul 21, 2016 at 12:33:31PM -0700, hmeng-19 wrote:
> I use a global var here to avoid multiple func calls to
> `context.GlobalBool`.

What's the drawback to multiple GlobalBool calls?  I like that more
than a package global, since it keeps the “single, unambiguous,
authoritative” representation of that setting [1].

[1]: https://en.wikipedia.org/wiki/Don't_repeat_yourself

mrunalp: Yeah, multiple calls to `GlobalBool` are fine. Also, can drop per command option in favor of global.

hmeng-19:
@mrunalp , @wking , PTAL.

man/ocitools.1.md
```diff
@@ -15,11 +15,25 @@ ocitools is a collection of tools for working with the [OCI runtime specificatio
...
+  Print usage statement.
 
 **-v**, **--version**
   Print version information.
 
+**--log-level**
+  Log level (panic, fatal, error, warn, info, or debug) (default: "error").
+
+**--host-specific**
+  Generate host-specific configs or do host-specific validations.
```
wking: This section (including the following paragraph) needs to be genericized now that it's covering both generate and validate.

generate/generate.go
```diff
@@ -904,11 +904,14 @@ func (g Generator) AddBindMount(bind string) error {
...
-func (g Generator) SetupPrivileged(privileged bool) {
+func (g Generator) SetupPrivileged(privileged bool, hostSpecific bool) {
 	if privileged {
 		// Add all capabilities in privileged mode.
 		var finalCapList []string
 		for _, cap := range capability.List() {
+			if hostSpecific && cap > capability.CAP_LAST_CAP {
+				continue
+			}
 			finalCapList = append(finalCapList, fmt.Sprintf("CAP_%s", strings.ToUpper(cap.String())))
```
wking: This section should probably be using `AddProcessCapability` and passing `hostSpecific` through too that function.  `AddProcessCapability` (or maybe `checkCap`) should actually be doing the conditional `CAP_LAST_CAP` comparison.

hmeng-19: @wking , Thanks for pointing out.
I would rather keep this function as it is for code efficiency.
I will fix `AddProcessCapability` and other related functions.

hmeng-19:
@wking , @mrunalp , PTAL.

generate/generate.go
wking: On Thu, Jul 21, 2016 at 03:03:04PM -0700, hmeng-19 wrote:
> I would rather keep this function as it is for code efficiency.

Are we really worried about performance for compiling a capability
list?  Can you spell out your concern in more detail or give a dummy
benchmark or something?

mrunalp: This could be addressed in a different PR.

wking: On Thu, Jul 21, 2016 at 03:39:31PM -0700, Mrunal Patel wrote:
> >  		// Add all capabilities in privileged mode.
> >  		var finalCapList []string
> >  		for _, cap := range capability.List() {
> > +			if hostSpecific && cap > capability.CAP_LAST_CAP {
> > +				continue
> > +			}
> >  			finalCapList = append(finalCapList, fmt.Sprintf("CAP_%s", strings.ToUpper(cap.String())))
> 
> This could be addressed in a different PR.

Using AddProcessCapability here?  I guess, if we can't decide in this
one.  I'm happy to file that PR if you want to land this without
addressing that point (although I still think we want to handle [1]
here).

[1]: https://github.com/opencontainers/ocitools/pull/138/files#r71790239

mrunalp: @wking Yes, you can open another PR to use AddProcessCapability here.

cmd/ocitools/generate.go
```diff
@@ -184,16 +184,18 @@ func setupSpec(g generate.Generator, context *cli.Context) error {
...
 	}
 
+	hostSpecific := context.GlobalBool("host-specific")
+
 	privileged := false
 	if context.IsSet("privileged") {
 		privileged = context.Bool("privileged")
 	}
-	g.SetupPrivileged(privileged)
+	g.SetupPrivileged(privileged, hostSpecific)
```
mrunalp: Thinking about this more, I think that we shouldn't pass this as part of the API. We should have a call to enable hostSpecific on the generate object and then use that variable in individual APIs

```go
g := generate.New()
g.EnableHostSpecific()
g.SetupPrivileged()
...
g.SaveToFile()
```

wking: On Thu, Jul 21, 2016 at 04:01:34PM -0700, Mrunal Patel wrote:
> >  	privileged := false
> >  	if context.IsSet("privileged") {
> >  		privileged = context.Bool("privileged")
> >  	}
> > -	g.SetupPrivileged(privileged)
> > +	g.SetupPrivileged(privileged, hostSpecific)
>
> Thinking about this more, I think that we shouldn't pass this as
> part of the API. We should have a call to enable hostSpecific on the
> generate object and then use that variable in individual APIs
> 
> ```go
> g := generate.New()
> g.EnableHostSpecific()
> g.SetupPrivileged()

This works for me.  I think it depends on the setting though; I prefer
seccomp-only to be a Save(ToFile) parameter and not a Generator-wide
setting [1].  The threshold is “can we think of a reason why a single
Generator would need different values in different contexts?”, and
that makes me a bit jumpy (my crystal ball isn't always that clear ;).
But in this case a Generator setting seems safe enough.

[1]: https://github.com/opencontainers/ocitools/pull/128#discussion_r71778273

hmeng-19:
@mrunalp , PTAL.

hmeng-19:
@mrunalp , `EnableHostSpecific` has a pointer receiver. Do you think I should change all the other receivers of other generate methods to be pointers?
I can also change `hostSpecific` to be a package-level var.

generate/generate.go
```diff
@@ -21,7 +21,8 @@ var (
 
 // Generator represents a generator for a container spec.
 type Generator struct {
-	spec *rspec.Spec
+	spec         *rspec.Spec
+	hostSpecific bool
```
wking: Why make this private with a setter (and no getter)?  Is there a reason to not make it public?

wking:
On Fri, Jul 22, 2016 at 09:45:37AM -0700, hmeng-19 wrote:
> Do you think I should change all the other receivers of other
> generate methods to be pointers?

I think so, but this should probably be a separate commit, since it's
going to be noisy.  Copy-by-value Generators relying on the underlying
Generator.spec being a pointer seems like it could be confusing.

man/ocitools.1.md
```diff
@@ -15,11 +15,26 @@ ocitools is a collection of tools for working with the [OCI runtime specificatio
...
+
+  By default, generator addes process capabilities into the config without
```
wking: “addes” → “adds”

man/ocitools.1.md
wking: This comment has been addressed.

man/ocitools.1.md
```diff
@@ -15,11 +15,26 @@ ocitools is a collection of tools for working with the [OCI runtime specificatio
...
+  checking whether it is supported on the current host. With this flag,
+  generator will first check whether a process capability is supported on the
+  current host, and only add it into the config if it passes the checking.
```
wking: I recommend dropping this last sentence, since it will go stale as we add more host-specific stuff.

Alternatively, you could keep it as a “for example”, in which case we are clearly not aspiring towards completely documenting all host-specific generation.  If you go this route, I'd restore the Linux/Windows example to the validation paragraph below.

generate/generate.go
hmeng-19: I currently did not think of any cases where `Generator.hostSpsecific` is queried outside the package.

man/ocitools.1.md
hmeng-19: Accept. Thanks.

man/ocitools.1.md
hmeng-19: Using `for example` sounds good.

hmeng-19:
> I think so, but this should probably be a separate commit, since it's
going to be noisy.  Copy-by-value Generators relying on the underlying
Generator.spec being a pointer seems like it could be confusing.

@wking , sounds good. I will work on it after the PR is done.

hmeng-19:
@wking , @mrunalp , PTAL.

generate/generate.go
wking: On Fri, Jul 22, 2016 at 10:16:03AM -0700, hmeng-19 wrote:
> > @@ -21,7 +21,8 @@ var (
> >  
> >  // Generator represents a generator for a container spec.
> >  type Generator struct {
> > -	spec *rspec.Spec
> > +	spec         *rspec.Spec
> > +	hostSpecific bool
> 
> I currently did not think of any cases where
> `Generator.hostSpsecific` is queried outside the package.

“I can't think of any reason you'd care” is different from “we don't
want people touching this directly”.  Since a public field doesn't
require getters/setters, I'd rather keep things public unless we have
reasons to make them private.

hmeng-19:
@wking , @mrunalp , I added `getHostSpecific` method and rebased the PR on the upstream master. PTAL.

wking:
On Mon, Jul 25, 2016 at 11:45:14AM -0700, hmeng-19 wrote:
> @wking , @mrunalp , I added `getHostSpecific` method and rebased the
> PR on the upstream master. PTAL.

Go suggests not using a ‘Get’ prefix [1], and I'd still rather just
make the property public and skip the getter/setter [2].

[1]: https://golang.org/doc/effective_go.html#Getters
[2]: https://github.com/opencontainers/ocitools/pull/138#discussion_r71917014

hmeng-19:
@wking , @mrunalp , I added a new commit which:
converts `hostSpecific` into a public field `HostSpecific`;
replace `GetSpec` with `Spec`.
PTAL. 
I will squash the commits before it is ready to be merged.

mrunalp:
LGTM

hmeng-19:
@mrunalp , PTAL.
I rebased the PR on top of the upstream master and squashed the commits into a single one.

cmd/ocitools/generate.go
```diff
@@ -96,8 +96,13 @@ var generateCommand = cli.Command{
...
 }
 
-func setupSpec(g generate.Generator, context *cli.Context) error {
-	spec := g.GetSpec()
+func setupSpec(gp *generate.Generator, context *cli.Context) error {
+	if context.GlobalBool("host-specific") {
+		gp.HostSpecific = true
+	}
+
+	g := *gp
```
wking: I don't think we need this.  It seemed to compile and run fine for me with the following change on top of 751e18b91:

    diff --git a/cmd/ocitools/generate.go b/cmd/ocitools/generate.go
    index ab1f6cc..a6423a0 100644
    --- a/cmd/ocitools/generate.go
    +++ b/cmd/ocitools/generate.go
    @@ -96,12 +96,11 @@ var generateCommand = cli.Command{
     	},
     }
     
    -func setupSpec(gp *generate.Generator, context *cli.Context) error {
    +func setupSpec(g *generate.Generator, context *cli.Context) error {
     	if context.GlobalBool("host-specific") {
    -		gp.HostSpecific = true
    +		g.HostSpecific = true
     	}
     
    -	g := *gp
     	spec := g.Spec()
     
     	if len(spec.Version) == 0 {
    @@ -374,7 +373,7 @@ func checkNs(nsMaps map[string]string, nsName string) bool {
     	return true
     }
     
    -func setupLinuxNamespaces(g generate.Generator, needsNewUser bool, nsMaps map[string]string) {
    +func setupLinuxNamespaces(g *generate.Generator, needsNewUser bool, nsMaps map[string]string) {
     	for _, nsName := range generate.Namespaces {
     		if !checkNs(nsMaps, nsName) && !(needsNewUser && nsName == "user") {
     			continue

wking:
On Tue, Jul 26, 2016 at 11:31:38AM -0700, hmeng-19 wrote:
> I rebased the PR on top of the upstream master and squashed the
> commits into a single one.

Except for [1] 751e18b9 looks good to me.

I'd rather see this v1.0.0-rc1-compatible change based on the
v1.0.0.rc1 branch so we could have it in both master and that branch
without cherry picking and subsequent post-v1.0.0.rc1 merge
duplication.  More on this in #140 and the comments after [2].

[1]: https://github.com/opencontainers/ocitools/pull/138/files#r72337364
[2]: https://github.com/opencontainers/ocitools/pull/128#issuecomment-235382768

hmeng-19:
@wking , thanks for pointing out. Fixed it. PTAL.

wking:
bf3f0d4 looks good to me, although I'd still rather not have to
cherry-pick it back into v1.0.0.rc1 [1].

[1]: https://github.com/opencontainers/ocitools/pull/138#issuecomment-235408339

hmeng-19:
@mrunalp , PTAL.

mrunalp:
LGTM

PR #139, opened 2016-07-21
remove extra else

hmeng-19:
This PR removes two extra else {}. This makes the code clean, also stops vim-bundle complaining.


hmeng-19:
@mrunalp , PTAL.

mrunalp:
LGTM 
cc: @wking

wking:
4a8e72d1 looks great :).

PR #141, opened 2016-07-22
test_runtime.sh: Drop pushd/popd from around 'generate'

wking:
The switch to `--output` in 06d61a34 (generate: Add `--output=PATH` and write to stdout by default, 2016-07-05, #129) was too narrow.  There's no need to change directories now that the output path is configurable.


mrunalp:
LGTM

PR #142, opened 2016-07-22
man: Shift examples from the README to man pages

wking:
If the examples are worth documenting, they're worth putting in the man
pages where users who install (but don't develop) ocitools can find them.

Also update the validation example to catch up with recent changes, and add
a --host-specific example to advertize that feature (since folks skimming
the man page quickly might expect the bare 'ocitools validate' to be
host-specific).

Also change `#` → `$` in the generate man page to show where you're
likely to need elevated permissions (you don't for `generate` calls).

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,25 +7,11 @@ ocitools is a collection of tools for working with the [OCI runtime specificatio
 [`ocitools generate`][generate.1] generates [configuration JSON][config.json] for an [OCI bundle][bundle].
 [OCI-compatible runtimes][runtime-spec] like [runC][] expect to read the configuration from `config.json`.
 
-```sh
-$ ocitools generate --output config.json
-$ cat config.json
-{
-        "ociVersion": "0.5.0",
-        …
-}
-```
-
 ## Validating an OCI bundle
 
 [`ocitools validate`][validate.1] validates an OCI bundle.
 The error message will be printed if the OCI bundle failed the validation procedure.
 
-```sh
-$ ocitools generate
-$ ocitools validate
-INFO[0000] Bundle validation succeeded.
-```
 
 ## Testing OCI runtimes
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -7,25 +7,11 @@ ocitools is a collection of tools for working with the [OCI runtime specificatio
...
 [`ocitools validate`][validate.1] validates an OCI bundle.
 The error message will be printed if the OCI bundle failed the validation procedure.
 
-```sh
-$ ocitools generate
-$ ocitools validate
-INFO[0000] Bundle validation succeeded.
-```
 
 ## Testing OCI runtimes
```
runcom: why this _example_ is here? should it go the man page as well?

runcom: I guess the question is: shouldn't testruntime.sh be a part of ocitools?

wking: On Fri, Jul 22, 2016 at 09:43:41AM -0700, Antonio Murdaca wrote:
> >  ## Testing OCI runtimes
> 
> I guess the question is: shouldn't testruntime.sh be a part of ocitools? 

Yes, but it's hard to do right ;).  See discussion in #61 and #98 (I'm
partial to [1] ;).

[1]: https://github.com/opencontainers/ocitools/pull/61#issuecomment-217910342

Mashimiao:
We should have examples in man pages, but I think should also leave examples in README.

wking:
On Mon, Feb 13, 2017 at 07:00:55PM -0800, Ma Shimiao wrote:
> We should have examples in man pages, but I think should also leave
> examples in README.

That's not very DRY, but if runtime-tools maintainers want to go that
way, they're welcome to.  I'm not interested in supporting that
approach myself, so if it's a maintainer requirement, feel free to
close this PR.

Mashimiao:
DRY is good but not required.  I think by now it more important to let users know how to use runtime-tools easily, just by README.md.
Maybe reopen this PR is necessary in the future.

PR #140, opened 2016-07-21
Update runtime-spec package

hmeng-19:
The PR vendors the latest version of the runtime-spec package, and updates generate.go.

PR #144, opened 2016-07-24
Check pointer fields of g.spec

hmeng-19:
The PR tries to check whether a pointer field is `nil` before modifying the field. This avoids modifying the field of a `nil` pointer.

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -206,34 +206,37 @@ func (g *Generator) SaveToFile(path string) error {
 
 // SetVersion sets g.spec.Version.
 func (g *Generator) SetVersion(version string) {
+	g.initSpec()
 	g.spec.Version = version
 }
 
 // SetRootPath sets g.spec.Root.Path.
 func (g *Generator) SetRootPath(path string) {
+	g.initSpec()
 	g.spec.Root.Path = path
 }
 
 // SetRootReadonly sets g.spec.Root.Readonly.
 func (g *Generator) SetRootReadonly(b bool) {
+	g.initSpec()
 	g.spec.Root.Readonly = b
 }
 
 // SetHostname sets g.spec.Hostname.
 func (g *Generator) SetHostname(s string) {
+	g.initSpec()
 	g.spec.Hostname = s
 }
 
 // ClearAnnotations clears g.spec.Annotations.
 func (g *Generator) ClearAnnotations() {
+	g.initSpec()
 	g.spec.Annotations = make(map[string]string)
 }
 
 // AddAnnotation adds an annotation into g.spec.Annotations.
 func (g *Generator) AddAnnotation(s string) error {
-	if g.spec.Annotations == nil {
-		g.spec.Annotations = make(map[string]string)
-	}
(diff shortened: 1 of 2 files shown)


hmeng-19:
@mrunalp , @wking , PTAL.
This PR is not ready to be merged directly, and should only be merged after #140.
I created the PR here, so you can see what I am up to.

wking:
On Sun, Jul 24, 2016 at 01:25:15PM -0700, hmeng-19 wrote:
> The PR tries to check whether a pointer field is `nil` before
> modifying the field.

I think we want ‘initialize*’ instead of ‘check*’, but the helpers are
more compact than the direct entries I'm currently using in #112.  Do
we have some threshold number of calls under which a helper function
is more mental overhead than its worth, or do we want to use
initializers for all of this sort of thing for consistency?

hmeng-19:
@wking , you are right. `initSpec*` is a better name. Fixed it.

hmeng-19:
@mrunalp , @wking , PTAL.
Since #140 has been merged, I think it is time to consider this PR.

wking:
Travis isn't happy with 889f775, possibly because of #140 landing [1]:

  go build -tags "" -o ocitools ./cmd/ocitools
  # github.com/opencontainers/ocitools/generate
  generate/generate.go:444: cannot convert nil to type specs.Linux
  …

I have some nil-Linux guards in #112, but they don't cover #137's
single-field setters like SetLinuxResourcesCPUShares.  We can add
guards to those methods in either this PR or #112, but I'd rather make
the config parameter public and drop the setters [2].

[1]: https://travis-ci.org/opencontainers/ocitools/builds/147189194#L193
[2]: https://github.com/opencontainers/ocitools/pull/137#issuecomment-234342815

hmeng-19:
@wking , I rebased this PR on top of the current master, the head of the PR should be https://github.com/opencontainers/ocitools/pull/144/commits/60df8fcbf16652ed394923cfb61d1f24a8285264 . On my side, github shows that `All checks have passed`.

wking:
On Tue, Jul 26, 2016 at 08:18:19AM -0700, hmeng-19 wrote:
> @wking , I rebased this PR on top of the current master, the head of
> the PR should be
> https://github.com/opencontainers/ocitools/pull/144/commits/60df8fcbf16652ed394923cfb61d1f24a8285264
> . On my side, github shows that `All checks have passed`.

Ah, looks like you pushed that while I was working on my comment.
60df8fcbf looks good to me, and with that this PR replaces #112.

wking:
On Tue, Jul 26, 2016 at 10:00:48AM -0700, W. Trevor King wrote:
> Ah, looks like you pushed that while I was working on my comment.
> 60df8fcbf looks good to me, and with that this PR replaces #112.

Actually, #112 still has the empty-Linux JSON serialization change, so
I'll keep it open.  After this PR lands, I'll rebase #112 on top of
the new master.

hmeng-19:
@wking , sounds good.

mrunalp:
LGTM

PR #147, opened 2016-07-25
move defaultCaps to validate

Mashimiao:

DefaultCaps is not used by generate any more. I think put it into validate is more suitable.


wking:
5f07d9d8 looks good to me.

mrunalp:
LGTM

PR #146, opened 2016-07-25
generate: remove unuseful default value for output

Mashimiao:
Signed-off-by: Ma Shimiao <mashimiao.fnst@cn.fujitsu.com>

PR #149, opened 2016-07-26
generate: remove unuseful default value for output

wking:
Backporting #146 to v1.0.0.rc1 following [the first procedure here][1].

[1]: https://github.com/opencontainers/ocitools/pull/140#issuecomment-234626841


mrunalp:
LGTM

PR #143, opened 2016-07-24
Use Generator pointers as methods receivers

hmeng-19:
Signed-off-by: Haiyan Meng <hmeng@redhat.com>

PR #150, opened 2016-07-26
Use Generator pointers as methods receivers

wking:
Backporting #143 to v1.0.0.rc1 following [the first procedure here][1].

[1]: https://github.com/opencontainers/ocitools/pull/140#issuecomment-234626841


mrunalp:
LGTM

PR #155, opened 2016-07-27
generate: fix tmpfs adding based on manpage

Mashimiao:
Signed-off-by: Ma Shimiao <mashimiao.fnst@cn.fujitsu.com>

PR #159, opened 2016-07-28
Use seccomp library for adding seccomp configuration flags

GrantSeltzer:
I added functionality to specify syscalls and their arguments for all actions. Also changed default action and archs flags to ociseccompgen library. 

Now we can do things like:

`ocitools generate --seccomp-errno clone,write --seccomp-trace getcwd --seccomp-trap umount:0:1:2:NE`

The behavior mimics that of: https://github.com/GrantSeltzer/Manhattan

Also updated man page to reflect this change. 

Signed-off-by: Grantseltzer <grantseltzer@gmail.com>

### DIFF
--- a/cmd/oci-runtime-tool/generate.go
+++ b/cmd/oci-runtime-tool/generate.go
@@ -9,6 +9,7 @@ import (
 
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/runtime-tools/generate"
+	"github.com/opencontainers/runtime-tools/generate/seccomp"
 	"github.com/urfave/cli"
 )
 
@@ -60,11 +61,17 @@ var generateFlags = []cli.Flag{
 	cli.BoolFlag{Name: "read-only", Usage: "make the container's rootfs read-only"},
 	cli.StringFlag{Name: "root-propagation", Usage: "mount propagation for root"},
 	cli.StringFlag{Name: "rootfs", Value: "rootfs", Usage: "path to the rootfs"},
-	cli.StringSliceFlag{Name: "seccomp-arch", Usage: "specifies Additional architectures permitted to be used for system calls"},
-	cli.StringSliceFlag{Name: "seccomp-allow", Usage: "specifies syscalls to be added to allowed"},
-	cli.StringFlag{Name: "seccomp-default", Usage: "specifies the the defaultaction of Seccomp syscall restrictions"},
-	cli.StringSliceFlag{Name: "seccomp-errno", Usage: "specifies syscalls to be added to list that returns an error"},
-	cli.StringSliceFlag{Name: "seccomp-syscalls", Usage: "specifies Additional architectures permitted to be used for system calls, e.g Name:Action:Arg1_index/Arg1_value/Arg1_valuetwo/Arg1_op, Arg2_index/Arg2_value/Arg2_valuetwo/Arg2_op "},
+	cli.StringFlag{Name: "seccomp-allow", Usage: "specifies syscalls to respond with allow"},
+	cli.StringFlag{Name: "seccomp-arch", Usage: "specifies additional architectures permitted to be used for system calls"},
+	cli.StringFlag{Name: "seccomp-default", Usage: "specifies default action to be used for system calls and removes existing rules with specified action"},
+	cli.StringFlag{Name: "seccomp-default-force", Usage: "same as seccomp-default but does not remove existing rules with specified action"},
+	cli.StringFlag{Name: "seccomp-errno", Usage: "specifies syscalls to respond with errno"},
+	cli.StringFlag{Name: "seccomp-kill", Usage: "specifies syscalls to respond with kill"},
+	cli.BoolFlag{Name: "seccomp-only", Usage: "specifies to export just a seccomp configuration file"},
+	cli.StringFlag{Name: "seccomp-remove", Usage: "specifies syscalls to remove seccomp rules for"},
+	cli.BoolFlag{Name: "seccomp-remove-all", Usage: "removes all syscall rules from seccomp configuration"},
+	cli.StringFlag{Name: "seccomp-trace", Usage: "specifies syscalls to respond with trace"},
+	cli.StringFlag{Name: "seccomp-trap", Usage: "specifies syscalls to respond with trap"},
 	cli.StringFlag{Name: "selinux-label", Usage: "process selinux label"},
 	cli.StringSliceFlag{Name: "sysctl", Usage: "add sysctl settings e.g net.ipv4.forward=1"},
 	cli.StringFlag{Name: "template", Usage: "base template to use for creating the configuration"},
@@ -102,11 +109,13 @@ var generateCommand = cli.Command{
 			return err
 		}
 
+		var exportOpts generate.ExportOptions
+		exportOpts.Seccomp = context.Bool("seccomp-only")
+
 		if context.IsSet("output") {
-			output := context.String("output")
-			err = specgen.SaveToFile(output)
+			err = specgen.SaveToFile(context.String("output"), exportOpts)
 		} else {
-			err = specgen.Save(os.Stdout)
+			err = specgen.Save(os.Stdout, exportOpts)
 		}
 		if err != nil {
 			return err
@@ -395,69 +404,8 @@ func setupSpec(g *generate.Generator, context *cli.Context) error {
 		g.SetLinuxResourcesMemorySwappiness(context.Uint64("linux-mem-swappiness"))
 	}
 
-	if context.IsSet("linux-pids-limit") {
-		g.SetLinuxResourcesPidsLimit(context.Int64("linux-pids-limit"))
-	}
-
-	var sd string
-	var sa, ss []string
-
-	if context.IsSet("seccomp-default") {
-		sd = context.String("seccomp-default")
-	}
-
-	if context.IsSet("seccomp-arch") {
-		sa = context.StringSlice("seccomp-arch")
-	}
-
-	if context.IsSet("seccomp-syscalls") {
-		ss = context.StringSlice("seccomp-syscalls")
-	}
-
-	if sd == "" && len(sa) == 0 && len(ss) == 0 {
-		return nil
-	}
-
-	// Set the DefaultAction of seccomp
-	if context.IsSet("seccomp-default") {
-		if err := g.SetLinuxSeccompDefault(sd); err != nil {
-			return err
-		}
-	}
-
-	// Add the additional architectures permitted to be used for system calls
-	if context.IsSet("seccomp-arch") {
-		for _, arch := range sa {
-			if err := g.AddLinuxSeccompArch(arch); err != nil {
-				return err
-			}
-		}
-	}
-
-	// Set syscall restrict in Seccomp
-	if context.IsSet("seccomp-syscalls") {
-		for _, syscall := range ss {
-			if err := g.AddLinuxSeccompSyscall(syscall); err != nil {
-				return err
-			}
-		}
-	}
-
-	if context.IsSet("seccomp-allow") {
-		seccompAllows := context.StringSlice("seccomp-allow")
-		for _, s := range seccompAllows {
-			g.AddLinuxSeccompSyscallAllow(s)
-		}
-	}
-
-	if context.IsSet("seccomp-errno") {
-		seccompErrnos := context.StringSlice("seccomp-errno")
-		for _, s := range seccompErrnos {
-			g.AddLinuxSeccompSyscallErrno(s)
-		}
-	}
-
-	return nil
+	err := addSeccomp(context, g)
+	return err
 }
 
 func setupLinuxNamespaces(context *cli.Context, g *generate.Generator, needsNewUser bool) {
@@ -542,4 +490,121 @@ func parseBindMount(s string) (string, string, string, error) {
 	}
 
 	return source, dest, options, nil
+}
+
+func addSeccomp(context *cli.Context, g *generate.Generator) error {
+
+	// Set the DefaultAction of seccomp
+	if context.IsSet("seccomp-default") {
+		seccompDefault := context.String("seccomp-default")
+		err := g.SetDefaultSeccompAction(seccompDefault)
+		if err != nil {
+			return err
+		}
+	} else if context.IsSet("seccomp-default-force") {
+		seccompDefaultForced := context.String("seccomp-default-force")
+		err := g.SetDefaultSeccompActionForce(seccompDefaultForced)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Add the additional architectures permitted to be used for system calls
+	if context.IsSet("seccomp-arch") {
+		seccompArch := context.String("seccomp-arch")
+		architectureArgs := strings.Split(seccompArch, ",")
+		for _, arg := range architectureArgs {
+			err := g.SetSeccompArchitecture(arg)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	if context.IsSet("seccomp-errno") {
+		err := seccompSet(context, "errno", g)
+		if err != nil {
+			return err
+		}
+	}
+
+	if context.IsSet("seccomp-kill") {
+		err := seccompSet(context, "kill", g)
+		if err != nil {
+			return err
+		}
+	}
+
+	if context.IsSet("seccomp-trace") {
+		err := seccompSet(context, "trace", g)
+		if err != nil {
+			return err
+		}
+	}
+
+	if context.IsSet("seccomp-trap") {
+		err := seccompSet(context, "trap", g)
+		if err != nil {
+			return err
+		}
+	}
+
+	if context.IsSet("seccomp-allow") {
+		err := seccompSet(context, "allow", g)
+		if err != nil {
+			return err
+		}
+	}
+
+	if context.IsSet("seccomp-remove") {
+		seccompRemove := context.String("seccomp-remove")
+		err := g.RemoveSeccompRule(seccompRemove)
+		if err != nil {
+			return err
+		}
+	}
+
+	if context.IsSet("seccomp-remove-all") {
+		err := g.RemoveAllSeccompRules()
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func seccompSet(context *cli.Context, seccompFlag string, g *generate.Generator) error {
+	flagInput := context.String("seccomp-" + seccompFlag)
+	flagArgs := strings.Split(flagInput, ",")
+	setSyscallArgsSlice := []seccomp.SyscallOpts{}
+	for _, flagArg := range flagArgs {
+		comparisonArgs := strings.Split(flagArg, ":")
+		if len(comparisonArgs) == 5 {
+			setSyscallArgs := seccomp.SyscallOpts{
+				Action:   seccompFlag,
+				Syscall:  comparisonArgs[0],
+				Index:    comparisonArgs[1],
+				Value:    comparisonArgs[2],
+				ValueTwo: comparisonArgs[3],
+				Operator: comparisonArgs[4],
+			}
+			setSyscallArgsSlice = append(setSyscallArgsSlice, setSyscallArgs)
+		} else if len(comparisonArgs) == 1 {
+			setSyscallArgs := seccomp.SyscallOpts{
+				Action:  seccompFlag,
+				Syscall: comparisonArgs[0],
+			}
+			setSyscallArgsSlice = append(setSyscallArgsSlice, setSyscallArgs)
+		} else {
+			return fmt.Errorf("invalid syscall argument formatting %v", comparisonArgs)
+		}
+
+		for _, r := range setSyscallArgsSlice {
+			err := g.SetSyscallAction(r)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
 }
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -7,10 +7,10 @@ import (
 	"io"
 	"os"
 	"runtime"
-	"strconv"
 	"strings"
 
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/opencontainers/runtime-tools/generate/seccomp"
 	"github.com/syndtr/gocapability/capability"
 )
 
@@ -23,6 +23,11 @@ var (
 type Generator struct {
 	spec         *rspec.Spec
 	HostSpecific bool
+}
+
+// ExportOptions have toggles for exporting only certain parts of the specification
+type ExportOptions struct {
+	Seccomp bool // seccomp toggles if only seccomp should be exported
 }
 
 // New creates a spec Generator with the default spec.
@@ -140,6 +145,7 @@ func New() Generator {
 			Devices: []rspec.Device{},
 		},
 	}
+	spec.Linux.Seccomp = seccomp.DefaultProfile(&spec)
 	return Generator{
 		spec: &spec,
 	}
@@ -187,8 +193,14 @@ func (g *Generator) Spec() *rspec.Spec {
 }
 
 // Save writes the spec into w.
-func (g *Generator) Save(w io.Writer) error {
-	data, err := json.MarshalIndent(g.spec, "", "\t")
+func (g *Generator) Save(w io.Writer, exportOpts ExportOptions) (err error) {
+	var data []byte
+
+	if exportOpts.Seccomp {
+		data, err = json.MarshalIndent(g.spec.Linux.Seccomp, "", "\t")
+	} else {
+		data, err = json.MarshalIndent(g.spec, "", "\t")
+	}
 	if err != nil {
 		return err
 	}
@@ -202,13 +214,13 @@ func (g *Generator) Save(w io.Writer) error {
 }
 
 // SaveToFile writes the spec into a file.
-func (g *Generator) SaveToFile(path string) error {
+func (g *Generator) SaveToFile(path string, exportOpts ExportOptions) error {
 	f, err := os.Create(path)
 	if err != nil {
 		return err
 	}
 	defer f.Close()
-	return g.Save(f)
+	return g.Save(f, exportOpts)
 }
 
 // SetVersion sets g.spec.Version.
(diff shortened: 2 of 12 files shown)


GrantSeltzer:
Failing for a different reason, fixing right now.

cmd/ocitools/generate.go
```diff
@@ -78,25 +82,27 @@ var generateCommand = cli.Command{
...
+
 		if err != nil {
 			return err
 		}
 		return nil
 	},
 }
 
-func setupSpec(g *generate.Generator, context *cli.Context) error {
+func setupSpec(g generate.Generator, context *cli.Context) error {
```
wking: Why are we removing the pointer here?

wking: They just landed via #138.

GrantSeltzer: Hmmm, Must have accidentally cut it when making the new branch. Thanks for the call out.

GrantSeltzer: @wking Fixed

cmd/ocitools/generate.go
```diff
@@ -305,61 +311,81 @@ func setupSpec(g *generate.Generator, context *cli.Context) error {
...
-	var sa, ss []string
+	err := addSeccomp(*g, context)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func addSeccomp(g generate.Generator, context *cli.Context) error {
```
wking: To stick with the new-since-around-#143 pattern, I'd rather see `*generate.Generator` here too.

wking: And you're following master's existing `setupSpec` pattern here, but I think we may want to flip the argument order (in both this and `setupSpec`) to [put `context` first][1].

[1]: https://blog.golang.org/context#TOC_5.

cmd/ocitools/generate.go
```diff
@@ -305,61 +311,81 @@ func setupSpec(g *generate.Generator, context *cli.Context) error {
...
+	seccompTrace := context.String("seccomp-trace")
+	seccompErrno := context.String("seccomp-errno")
+	seccompTrap := context.String("seccomp-trap")
+	seccompAllow := context.String("seccomp-allow")
 
-	if context.IsSet("seccomp-default") {
-		sd = context.String("seccomp-default")
+	// Set the DefaultAction of seccomp
+	if seccompDefault == "" {
+		seccompDefault = "errno"
```
wking: You should be able to accomplish this by setting [`Value`][1] when you define the option.

[1]: https://godoc.org/gopkg.in/urfave/cli.v1#StringFlag

cmd/ocitools/generate.go
```diff
@@ -305,61 +311,81 @@ func setupSpec(g *generate.Generator, context *cli.Context) error {
...
+	architectureArgs := strings.Split(seccompArch, ",")
+	err = g.SetSeccompArchitectures(architectureArgs)
+	if err != nil {
+		return err
 	}
 
-	// Set the DefaultAction of seccomp
-	if context.IsSet("seccomp-default") {
-		if err := g.SetLinuxSeccompDefault(sd); err != nil {
+	if seccompKill != "" {
```
wking: Can we use `IsSet` instead of comparing with the empty string here?  It would be nice to die with an error (instead of silently ignoring) folks that call `generate --seccomp-kill ''`.  Or maybe that should clear the kill action?  But I'm pretty sure silently ignoring is not what we want.

generate/generate.go
```diff
@@ -137,6 +143,7 @@ func New() Generator {
 					Type: "mount",
 				},
 			},
+			Seccomp: &rspec.Seccomp{},
```
wking: Instead of adding an empty object here, the current pattern (since the recent #144) is to use an `init*` guard in each `Generator` method that requires the struct.

generate/generate.go
```diff
@@ -1124,42 +865,22 @@ func (g *Generator) RemoveLinuxNamespace(ns string) error {
...
-				Op:       op,
-			}
-			Args = append(Args, &Arg)
-		} else {
-			return nil, fmt.Errorf("seccomp-sysctl args error: %s", argstr)
-		}
-	}
-	return Args, nil
+// SetSyscallActions adds rules for syscalls with the specified action
+func (g Generator) SetSyscallActions(action string, arguments []string) error {
```
wking: These should also be pointer receivers to match #143.

generate/seccomp/doc.go
```diff
@@ -1,2 +0,0 @@
-// Package seccomp includes functions for configuring seccomp.
-package seccomp
```
wking: I think it's worth keeping something like this around for godoc.

GrantSeltzer: Should we add it back later when it's more complete?

wking: On Thu, Jul 28, 2016 at 08:19:16AM -0700, Grant Seltzer Richman wrote:
> > @@ -1,2 +0,0 @@
> > -// Package seccomp includes functions for configuring seccomp.
> > -package seccomp
> 
> Should we add it back later when it's more complete? 

I'm fine filling it in now, if we have more to say.  But if we don't
have more to say I'd rather leave it alone.  Do we gain something by
removing it?

GrantSeltzer: I believe it was was just a place holder that Haiyan added to tell me where the seccomp library should go when she separated out generate. I can't think of anything more to say, so I don't think we gain anything from leaving it. It's also just a single line....If we want to add it back with more information then we're not being set back by removing it now.

wking: On Thu, Jul 28, 2016 at 11:02:04AM -0700, Grant Seltzer Richman wrote:
> I can't think of anything more to say, so I don't think we gain
> anything from leaving it…

I feel like if it's a wash we should leave it alone ;), but I'm not
strongly against removing it.

cmd/ocitools/generate.go
```diff
@@ -305,61 +311,73 @@ func setupSpec(g *generate.Generator, context *cli.Context) error {
...
-		sd = context.String("seccomp-default")
+	err := addSeccomp(g, context)
+	if err != nil {
+		return err
 	}
 
-	if context.IsSet("seccomp-arch") {
-		sa = context.StringSlice("seccomp-arch")
-	}
+	return nil
```
mrunalp: could just return err here.

cmd/ocitools/generate.go
```diff
@@ -305,61 +311,73 @@ func setupSpec(g *generate.Generator, context *cli.Context) error {
...
+	if err != nil {
+		return err
 	}
 
-	if sd == "" && len(sa) == 0 && len(ss) == 0 {
-		return nil
+	// Add the additional architectures permitted to be used for system calls
+	seccompArch := context.String("seccomp-arch")
+	architectureArgs := strings.Split(seccompArch, ",")
+	err = g.SetSeccompArchitectures(architectureArgs)
```
mrunalp: Can we add this one at a time similar to have to add/drop capabilities?

cmd/ocitools/generate.go
```diff
@@ -305,61 +311,73 @@ func setupSpec(g *generate.Generator, context *cli.Context) error {
...
+		return err
 	}
 
-	// Set the DefaultAction of seccomp
-	if context.IsSet("seccomp-default") {
-		if err := g.SetLinuxSeccompDefault(sd); err != nil {
+	if context.IsSet("seccomp-kill") {
+		seccompKill := context.String("seccomp-kill")
+		killArgs := strings.Split(seccompKill, ",")
+		err := g.SetSyscallActions("kill", killArgs)
```
mrunalp: We should do this one at a time if possible renaming it to AddSyscallAction.

mrunalp: Also, there should be a constant defined for "kill" rather than accepting raw strings in the API.

generate/seccomp/parse_action.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"github.com/Sirupsen/logrus"
+	types "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+// ParseSyscallFlag takes the name of the action, the arguments (syscalls) that were
+// passed with it at the command line and a pointer to the config struct. It parses
+// the action and syscalls and updates the config accordingly
+func ParseSyscallFlag(action string, syscallArgs []string, config *types.Seccomp) error {
+
```
mrunalp: nit: Remove empty line.

cmd/ocitools/generate.go
mrunalp: Rather than passing arguments for parsing in the library, we should pass an option struct here.

cmd/ocitools/generate.go
mrunalp: Option struct for this API too.

cmd/ocitools/generate.go
GrantSeltzer: Not sure what you mean here, can you clarify?

cmd/ocitools/generate.go
GrantSeltzer: Not disagreeing, but what is the design advantage to passing the arguments one at a time instead of as a slice? 

Agreed on having a constant.

Again, if you can clarify about your idea for an option struct.

wking:
dfe2722 → cfb6d90 talks about rebasing, and pulls in some recent
master changes like #170, but doesn't seem to have actually rebased
anything.

  $ git log --graph --oneline --decorate origin/pr/159 | head -n4
  * cfb6d90 (origin/pr/159) Rebase and clean up seccomp generate api
  * dfe2722 Change generator receivers to pointers, other fixes
  * 22752db Use seccomp library for adding seccomp configuration flags
  *   3c2125c Merge pull request #155 from Mashimiao/generate-fix-tmpfs-adding

You probably intended to rebase onto the current master.

GrantSeltzer:
@wking Cleaning up now

wking:
On Tue, Aug 02, 2016 at 12:55:58PM -0700, Grant Seltzer Richman wrote:
> @wking Cleaning up now

cfb6d90 → a9b3569 looks like it rebases onto master and then merges
your old tip:

  $ git glog a9b3569 | head -n7
  *   a9b3569 (origin/pr/159) Fix rebase
  |\  
  | * cfb6d90 Rebase and clean up seccomp generate api
  | * dfe2722 Change generator receivers to pointers, other fixes
  | * 22752db Use seccomp library for adding seccomp configuration flags
  * | a116901 Use seccomp library for adding seccomp configuration flags
  * |   9d3881d (origin/master) Merge pull request #174 from hmeng-19/update_cli

The difference between a116901 and a9b3569 is that the latter removes
duplicate entries for setupLinuxNamespaces, parseIDMapping, etc.  I
expect you:

1. Rebased cfb6d90 onto master to create a116901.
2. Tried to push a116901 to your public add-seccomp-cli-features, but
   Git warned you that it wasn't a fast-forward and suggested you merge.
3. Merged your old tip (cfb6d90) to get a9b3569.
4. Tried to push again (which succeeded).

What you probably wanted to do instead was use a --force push in (2)
to say “I know this isn't a fast forward, and I'm ok with that.
Please update the remote branch anyway”.

So I think you want to recover with something like:

  $ git checkout add-seccomp-cli-features
  $ git reset --hard a116901
  $ git checkout a9b3569 -- cmd/ocitools/generate.go
  $ git commit --amend --no-edit cmd/ocitools/generate.go
  $ git push --force GrantSeltzer add-seccomp-cli-features

GrantSeltzer:
@wking Wow thank you very much, My git skills have much improving to be done, I appreciate your help!

cmd/ocitools/generate.go
```diff
@@ -319,65 +325,8 @@ func setupSpec(g *generate.Generator, context *cli.Context) error {
...
-
-	if context.IsSet("seccomp-errno") {
-		seccompErrnos := context.StringSlice("seccomp-errno")
-		for _, s := range seccompErrnos {
-			g.AddLinuxSeccompSyscallErrno(s)
-		}
-	}
-
-	return nil
+	err := addSeccomp(g, context)
```
mrunalp: pass context first

cmd/ocitools/generate.go
```diff
@@ -463,3 +412,177 @@ func parseBindMount(s string) (string, string, string, error) {
 
 	return source, dest, options, nil
 }
+
+func addSeccomp(g *generate.Generator, context *cli.Context) error {
+
```
mrunalp: Remove empty line

cmd/ocitools/generate.go
```diff
@@ -463,3 +412,177 @@ func parseBindMount(s string) (string, string, string, error) {
...
+	seccompDefault := context.String("seccomp-default")
+	err := g.SetDefaultSeccompAction(seccompDefault)
+	if err != nil {
+		return err
+	}
+
+	// Add the additional architectures permitted to be used for system calls
+	seccompArch := context.String("seccomp-arch")
+	architectureArgs := strings.Split(seccompArch, ",")
+	err = g.SetSeccompArchitectures(architectureArgs)
```
mrunalp: I was thinking that this should be called like this (so a loop here):
```
g.SetSeccompArchitecture("x86")
g.SetSeccompArchitecutre("mips")
```

cmd/ocitools/generate.go
```diff
@@ -463,3 +412,177 @@ func parseBindMount(s string) (string, string, string, error) {
...
+			for _, r := range setSyscallArgsSlice {
+				err := g.SetSyscallAction(r)
+				if err != nil {
+					return err
+				}
+			}
+		}
+	}
+	if context.IsSet("seccomp-trace") {
+		seccompTrace := context.String("seccomp-trace")
```
mrunalp: Looks like we have a repeated code pattern which could be abstracted into a helper function.

GrantSeltzer:
@mrunalp PTAL

generate/seccomp/consts.go
```diff
@@ -0,0 +1,12 @@
+package seccomp
+
+const (
+	seccompOverwrite = "overwrite"
+	seccompAppend    = "append"
+	nothing          = "nothing"
```
mrunalp: need to run gofmt on this file?

mrunalp:
I think this looks fine overall. @hmeng-19 Could you test/review as well?

generate/seccomp/consts.go
GrantSeltzer: :+1:

hmeng-19:
@GrantSeltzer , `man/ocitools-generate.1.md` also needs to be updated to reflect the new options.

cmd/ocitools/generate.go
```diff
@@ -463,3 +413,95 @@ func parseBindMount(s string) (string, string, string, error) {
...
+	err := g.SetDefaultSeccompAction(seccompDefault)
+	if err != nil {
+		return err
+	}
+
+	// Add the additional architectures permitted to be used for system calls
+	seccompArch := context.String("seccomp-arch")
+	architectureArgs := strings.Split(seccompArch, ",")
+	for _, arg := range architectureArgs {
+		err = g.SetSeccompArchitecture(arg)
```
hmeng-19: @GrantSeltzer , it is not a real problem. But I believe you want to `err := ...`.

cmd/ocitools/generate.go
```diff
@@ -463,3 +413,95 @@ func parseBindMount(s string) (string, string, string, error) {
...
+			}
+			setSyscallArgsSlice = append(setSyscallArgsSlice, setSyscallArgs)
+		} else if len(comparisonArgs) == 1 {
+			setSyscallArgs := seccomp.SyscallOpts{
+				Action:  seccompFlag,
+				Syscall: comparisonArgs[0],
+			}
+			setSyscallArgsSlice = append(setSyscallArgsSlice, setSyscallArgs)
+		} else {
+			return errors.New("Invalid syscall argument formatting")
```
hmeng-19: @GrantSeltzer , you may want to try `fmt.Errorf` here. It allows you to output which `flagArg` is broken.
```
func Errorf(format string, a ...interface{}) error
```

generate/generate.go
```diff
@@ -25,6 +25,11 @@ type Generator struct {
 	HostSpecific bool
 }
 
+// ExportOptions has toggles for exporting only certain parts of the specification
```
hmeng-19: `has` -> `have`

generate/seccomp/parse_action.go
```diff
@@ -0,0 +1,116 @@
...
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/Sirupsen/logrus"
+	types "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+// SyscallOpts contains options for parsing syscall rules
```
hmeng-19: `contains` -> `contain`

generate/seccomp/parse_remove.go
```diff
@@ -0,0 +1,33 @@
...
+	"fmt"
+	"strings"
+
+	types "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+// RemoveAction takes the argument string that was passed with the --remove flag,
+// parses it, and updates the Seccomp config accordingly
+func RemoveAction(arguments string, config *types.Seccomp) error {
+
```
hmeng-19: This extra line can be removed.

generate/seccomp/syscall_compare.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	// Append has the lowest priority
+	for _, determinedAction := range sliceOfDeterminedActions {
+		if determinedAction == seccompAppend {
+			return determinedAction, nil
+		}
+	}
+
+	return "error", fmt.Errorf("Trouble determining action: %s", sliceOfDeterminedActions)
```
hmeng-19: @GrantSeltzer , do correct if I am wrong.
the first return value, `"error"` here seems unnecessary, right? The second return value already marks that an error happened.

hmeng-19:
@GrantSeltzer , @mrunalp , I left comments on several tiny points. 
Otherwise, the PR looks good to me. I tested it, and it works perfectly. 
Good job, Grant.

cmd/ocitools/generate.go
GrantSeltzer: Good idea

generate/seccomp/syscall_compare.go
GrantSeltzer: Yea, i'm not sure what the idiomatic way of doing this was. It didn't feel right to return `determinedAction`, it's certainly another check incase the user isn't properly checking errors. Would you use just an empty string?

cmd/ocitools/generate.go
GrantSeltzer: What difference does it make in this context? Both appear to work

generate/seccomp/syscall_compare.go
wking: On Thu, Aug 04, 2016 at 02:42:49PM -0700, Grant Seltzer Richman wrote:
> > +	return "error", fmt.Errorf("Trouble determining action: %s", sliceOfDeterminedActions)
> 
> Yea, i'm not sure what the idiomatic way of doing this was. It
> didn't feel right to return `determinedAction`, it's certainly
> another check incase the user isn't properly checking errors. Would
> you use just an empty string?

I see ‘"", err’ frequently, although I don't read a lot of Go.  In
master:

  $ git grep 'return .*""'
  Godeps/_workspace/src/github.com/Sirupsen/logrus/entry.go:              return "", err
  Godeps/_workspace/src/github.com/Sirupsen/logrus/hooks/sentry/sentry.go:                return "", false
  Godeps/_workspace/src/github.com/Sirupsen/logrus/hooks/sentry/sentry.go:                return "", false
  Godeps/_workspace/src/github.com/blang/semver/range.go:         return nil, fmt.Errorf("Could not parse comparator %q in %q", opStr, strings.Join([]string{opStr, vStr}, ""))
  Godeps/_workspace/src/github.com/blang/semver/range.go:         return nil, fmt.Errorf("Could not parse version %q in %q: %s", vStr, strings.Join([]string{opStr, vStr}, ""), err)
  Godeps/_workspace/src/github.com/blang/semver/range.go:         return "", "", fmt.Errorf("Could not get version from string: %q", s)
  Godeps/_workspace/src/github.com/blang/semver/semver.go:                return "", errors.New("Buildversion is empty")
  Godeps/_workspace/src/github.com/blang/semver/semver.go:                return "", fmt.Errorf("Invalid character(s) found in build meta data %q", s)
  Godeps/_workspace/src/github.com/blang/semver/semver_test.go:   return PRVersion{"", i, true}
  Godeps/_workspace/src/github.com/urfave/cli/context.go: return ""
  Godeps/_workspace/src/github.com/urfave/cli/context.go: return ""
  Godeps/_workspace/src/github.com/urfave/cli/context.go: return ""
  Godeps/_workspace/src/github.com/urfave/cli/flag.go:    return "", usage

GrantSeltzer: Yea, I've seen that too. I'm going to change it to that.

hmeng-19: @GrantSeltzer , `return "", err` is pretty authentic.

cmd/ocitools/generate.go
hmeng-19: This works when you declare `err` before using `var err error` or `err := ....`. 
I pointed out because all the other for loops in this function seem to declare its own `err`.

mrunalp: For consistency, let's change this to `err := `

GrantSeltzer: Sure

cmd/ocitools/generate.go
```diff
@@ -468,3 +417,95 @@ func parseBindMount(s string) (string, string, string, error) {
...
+			}
+			setSyscallArgsSlice = append(setSyscallArgsSlice, setSyscallArgs)
+		} else if len(comparisonArgs) == 1 {
+			setSyscallArgs := seccomp.SyscallOpts{
+				Action:  seccompFlag,
+				Syscall: comparisonArgs[0],
+			}
+			setSyscallArgsSlice = append(setSyscallArgsSlice, setSyscallArgs)
+		} else {
+			return fmt.Errorf("Invalid syscall argument formatting %v", comparisonArgs)
```
mrunalp: s/Invalid/invalid See: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

generate/seccomp/parse_action.go
```diff
@@ -0,0 +1,116 @@
...
+type SyscallOpts struct {
+	Action   string
+	Syscall  string
+	Index    string
+	Value    string
+	ValueTwo string
+	Operator string
+}
+
+// ParseSyscallFlag takes the name of the action, the arguments (syscalls) that were
```
mrunalp: comment out of date?

generate/seccomp/parse_architecture.go
```diff
@@ -0,0 +1,53 @@
...
+		"mipsel64n32": types.ArchMIPSEL64N32,
+		"ppc":         types.ArchPPC,
+		"ppc64":       types.ArchPPC64,
+		"ppc64le":     types.ArchPPC64LE,
+		"s390":        types.ArchS390,
+		"s390x":       types.ArchS390X,
+	}
+	a, ok := arches[arch]
+	if !ok {
+		return "", fmt.Errorf("Unrecognized architecutre: %s", arch)
```
mrunalp: typo: architecture

mrunalp: Also start error string lower case. This applies to call error strings in this PR.

generate/seccomp/parse_action.go
```diff
@@ -0,0 +1,116 @@
...
+	if len(arguments) == 6 {
+		typeStruct, err := parseArguments(arguments[1:])
+		if err != nil {
+			return err
+		}
+		newSyscall = newSyscallStruct(arguments[1], action, *typeStruct)
+	} else if len(arguments) == 2 {
+		newSyscall = newSyscallStruct(arguments[1], action, nil)
+	} else {
+		return errors.New("Incorrect number of arguments to ParseSyscall")
```
mrunalp: Let's prefer using fmt.Errorf and print how many were passed vs. expected.

generate/seccomp/parse_arguments.go
```diff
@@ -0,0 +1,71 @@
...
+			ValueTwo: syscallValueTwo,
+			Op:       syscallOp,
+		}
+
+		argSlice := new([]types.Arg)
+		*argSlice = append(*argSlice, argStruct)
+		return argSlice, nil
+	}
+
+	return nilArgSlice, errors.New("Incorrect number of arguments passed with syscall")
```
mrunalp: Same here.

generate/seccomp/parse_action.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	descison, err := decideCourseOfAction(&newSyscall, config.Syscalls)
+	if err != nil {
+		return err
+	}
+	delimDescison := strings.Split(descison, ":")
+
+	if delimDescison[0] == nothing {
+		logrus.Info("No action taken: ", newSyscall)
```
mrunalp: libraries shouldn't log unless there is a flag to enable/disable it.

generate/seccomp/parse_action.go
```diff
@@ -0,0 +1,116 @@
...
+	"kill":  types.ActKill,
+	"trace": types.ActTrace,
+	"trap":  types.ActTrap,
+}
+
+// Take passed action, return the SCMP_ACT_<ACTION> version of it
+func parseAction(action string) (types.Action, error) {
+	a, ok := actions[action]
+	if !ok {
+		return "", fmt.Errorf("Unrecognized action: %s", action)
```
mrunalp: s/Unrecognized/unrecognized

generate/seccomp/parse_arguments.go
```diff
@@ -0,0 +1,71 @@
...
+// parseArguments takes a list of arguments (delimArgs). It parses and fills out
+// the argument information and returns a slice of arg structs
+func parseArguments(delimArgs []string) (*[]types.Arg, error) {
+	nilArgSlice := new([]types.Arg)
+
+	if len(delimArgs) == 1 {
+		return nilArgSlice, nil
+	}
+
+	if len(delimArgs) == 5 {
```
mrunalp: Define a const for 5 or add a comment explaining why we are using 5 here.

generate/seccomp/parse_arguments.go
```diff
@@ -0,0 +1,71 @@
...
+		"LT": types.OpLessThan,
+		"LE": types.OpLessEqual,
+		"EQ": types.OpEqualTo,
+		"GE": types.OpGreaterEqual,
+		"GT": types.OpGreaterThan,
+		"ME": types.OpMaskedEqual,
+	}
+	o, ok := operators[operator]
+	if !ok {
+		return "", fmt.Errorf("Unrecognized operator: %s", operator)
```
mrunalp: s/Unrecognized/unrecognized

generate/seccomp/syscall_compare.go
```diff
@@ -0,0 +1,143 @@
+/*******************************************************************************
```
mrunalp: This comment doesn't fit recommended style. See https://golang.org/doc/effective_go.html#commentary.

mrunalp:
I just tested this and see an issue. Even if I don't pass any seccomp flags, this section is added by default. It should not be present at all.  
```
           "seccomp": {
                        "defaultAction": "SCMP_ACT_ERRNO",
                        "architectures": [
                                "SCMP_ARCH_X86_64",
                                "SCMP_ARCH_X86",
                                "SCMP_ARCH_X32"
                        ]
                }
```

GrantSeltzer:
@mrunalp Can you clarify on what you want to the behavior to be?

If no seccomp flag is specified nothing should be produced, that makes a lot of sense.

Do you still want to maintain defaults though? so if `seccomp-default` is passed with nothing it should default to `SCMP_ACT_ERNNO`? 

Also should `architecutres` and `defaultAction` be added if a seccomp syscall flag is passed?

mrunalp:
I think that for good templating support, it makes sense not to add anything unless specified.

GrantSeltzer:
Sure @mrunalp

GrantSeltzer:
So the behavior I think you want, @mrunalp, is having `ocitools generate` to not include anything with seccomp. 

Something like `ocitools generate --seccomp-kill clone` to create a kill rule for clone but with `defaultAction` and `architectures` be null or empty. 

`ocitools generate --seccomp-only` would just have a null output. 

If this is correct I can go ahead and push these changes.

mrunalp:
@GrantSeltzer Yes, that sounds good.

mrunalp:
It doesn't break runtime test anymore. I'll do some more manual testing.

GrantSeltzer:
Thanks, Mrunal!

GrantSeltzer:
@rhatdan and I discussed that we should have the [default seccomp profile](https://github.com/GrantSeltzer/Manhattan/blob/master/manhattan.json) be automatically placed as the seccomp section in generate since it's a white-list, not a black-list. 

This would just entail having that be set as the default, and have the other seccomp flags be used to edit that. A flag can also be included to build one from scratch. As for if this should be in this PR or taken care of in a direct follow up is up for discussion.

wking:
On Fri, Aug 12, 2016 at 09:08:18AM -0700, Grant Seltzer Richman wrote:
> @rhatdan and I discussed that we should have the default seccomp
> profile be automatically placed as the seccomp section in generate
> since it's a white-list, not a black-list.

I think that this should go in generate.New, like the rest of the
defaults recommended by OCI maintainers, and not be injected into
templates loaded via generate.NewFrom*.  That way, folks that want
advice from OCI maintainers can start from the default,
OCI-maintainer-recommended template, while folks who want more control
(via --template or some such) can avoid implicit injection.

If that sounds good, I think it should be part of this PR.

rhatdan:
Sounds good to me.  I would suggest you add a flag to override the default with a json file also.

ocitools generate --seccomp-file foobar.json

But having something close to the docker basic json file for  a white list would be a good default.

GrantSeltzer:
I'm on it

mrunalp:
Works for me!

wking:
On Fri, Aug 12, 2016 at 09:27:59AM -0700, Daniel J Walsh wrote:
> Sounds good to me.  I would suggest you add a flag to override the
> default with a json file also.
> 
> ocitools generate --seccomp-file foobar.json

I don't see the need for this, since you can already use --template
with your seccomp JSON (and other settings) included.  Folks that want
*most* of the OCI-maintainer-specified default but want to clobber the
seccomp stuff can use the other --seccomp-* options already in this PR
or:

  $ jq '.linux.seccomp = $seccomp' --argfile seccomp seccomp.json template.json >config.json

rhatdan:
Ok that works for me.

GrantSeltzer:
@mrunalp @rhatdan @wking 

Would it be preferred to have the default white list be a part of the default `ocitools generate` or should a flag be specified to add it?

mrunalp:
Should be part of default

rhatdan:
Part of default.

GrantSeltzer:
A default seccomp config is now a part of the output of `ocitools generate`, including detection of the system architecture.

mrunalp:
I just tried this out and the runtime tests fail:
```
-----------------------------------------------------------------------------------
VALIDATING RUNTIME: runc
-----------------------------------------------------------------------------------
runtime: failed to create new OS thread (have 2 already; errno=1)
fatal error: newosproc

runtime stack:
runtime.throw(0x57c714, 0x9)
        /usr/lib/golang/src/runtime/panic.go:566 +0x95
runtime.newosproc(0xc420028000, 0xc420037fc0)
        /usr/lib/golang/src/runtime/os_linux.go:160 +0x194
runtime.newm(0x5907f8, 0x0)
        /usr/lib/golang/src/runtime/proc.go:1572 +0x132
runtime.main.func1()
        /usr/lib/golang/src/runtime/proc.go:126 +0x36
runtime.systemstack(0x63bb00)
        /usr/lib/golang/src/runtime/asm_amd64.s:298 +0x79
runtime.mstart()
        /usr/lib/golang/src/runtime/proc.go:1079

goroutine 1 [running]:
runtime.systemstack_switch()
        /usr/lib/golang/src/runtime/asm_amd64.s:252 fp=0xc420022768 sp=0xc420022760
runtime.main()
        /usr/lib/golang/src/runtime/proc.go:127 +0x6c fp=0xc4200227c0 sp=0xc420022768
runtime.goexit()
        /usr/lib/golang/src/runtime/asm_amd64.s:2086 +0x1 fp=0xc4200227c8 sp=0xc4200227c0
Runtime runc failed validation
```

GrantSeltzer:
Which test is that? I can't reproduce that. I think what could be happening is there's a system call not in the white-list that is falling under the default action to `errno`

GrantSeltzer:
I've added the syscalls to the default whitelist based on allowed capabilities, pretty sure that was the issue.

GrantSeltzer:
@mrunalp PTAL

generate/seccomp/parse_action.go
```diff
@@ -0,0 +1,108 @@
...
+// Take passed action, return the SCMP_ACT_<ACTION> version of it
+func parseAction(action string) (types.Action, error) {
+	a, ok := actions[action]
+	if !ok {
+		return "", fmt.Errorf("unrecognized action: %s", action)
+	}
+	return a, nil
+}
+
+//ParseDefaultAction simply sets the default action of the seccomp configuration
```
mrunalp: nit: space after //

mrunalp:
From my testing:
```
 oci-runtime-tool generate --output config.json --tty --seccomp-errno getcwd 
```
I still see the default seccomp options with getcwd being allowed.

mrunalp:
Next I tried to use a template with no syscalls set.
```
[root@dhcp-16-129 busybox]# oci-runtime-tool generate --output config.json --tty --seccomp-errno getcwd --template template.json                                                                                                                              
[root@dhcp-16-129 busybox]# vim config.json 
[root@dhcp-16-129 busybox]# tail -n 20 config.json 
                        {
                                "type": "ipc"
                        },
                        {
                                "type": "uts"
                        },
                        {
                                "type": "mount"
                        }
                ],
                "seccomp": {
                        "defaultAction": "SCMP_ACT_ERRNO",
                        "architectures": [
                                "SCMP_ARCH_X86_64",
                                "SCMP_ARCH_X86",
                                "SCMP_ARCH_X32"
                        ]
                }
        }
```

GrantSeltzer:
For anyone else involved, based on my discussion with @mrunalp

There's a check that will error out if a syscall rule is specified that has the same action as the defaultAction (I've added an error message instead of it failing silently). 

I'm going to be adding a `seccomp-syscall-remove` flag for the edge case of a rule for a system call existing already but we want to change it to the default (For example, if getcwd is set to allow, and the defaultAction is errno, and we want to change getcwd to errno). I'm also adding a flag to remove all syscall rules. 

The other edge case that needed to be addressed is when a new default action is specified but there are already system calls that match it. This should cause an error and to report the system calls that have this matching action.

mrunalp:
@GrantSeltzer For changing the default action, we should add a force flag in the API, so it clears syscalls that match the new action.

GrantSeltzer:
@mrunalp Right, got it.

GrantSeltzer:
The build keeps failing for what seems to be a CI issue, I'm not sure if something can be reset or this issue will just go away on its own tomorrow. 

```
# github.com/golang/lint

../../golang/lint/lint.go:249: undefined: types.ImportMode

The command "go get github.com/golang/lint/golint" failed and exited with 2 during .
```

Regardless, I've added flags for removing specific syscall rules or just all of them. I've also added a force option for setting the default action without it deleting rules that already have the same action set.

mrunalp:
LGTM
@Mashimiao @liangchenye @wking If you want to give it a look.

cmd/oci-runtime-tool/generate.go
```diff
@@ -51,11 +52,17 @@ var generateFlags = []cli.Flag{
...
+	cli.StringFlag{Name: "seccomp-arch", Usage: "specifies additional architectures permitted to be used for system calls"},
+	cli.StringFlag{Name: "seccomp-default", Usage: "specifies default action to be used for system calls and removes existing rules with specified action"},
+	cli.StringFlag{Name: "seccomp-default-force", Usage: "same as seccomp-default but does not remove existing rules with specified action"},
+	cli.StringFlag{Name: "seccomp-allow", Usage: "specifies syscalls to respond with allow"},
+	cli.StringFlag{Name: "seccomp-trap", Usage: "specifies syscalls to respond with trap"},
+	cli.StringFlag{Name: "seccomp-errno", Usage: "specifies syscalls to respond with errno"},
+	cli.StringFlag{Name: "seccomp-trace", Usage: "specifies syscalls to respond with trace"},
+	cli.StringFlag{Name: "seccomp-kill", Usage: "specifies syscalls to respond with kill"},
+	cli.StringFlag{Name: "seccomp-remove", Usage: "specifies syscalls to remove seccomp rules for"},
+	cli.BoolFlag{Name: "seccomp-remove-all", Usage: "removes all syscall rules from seccomp configuration"},
```
Mashimiao: I think the options should be in sorted list as #204 does. So we don't need to do it again.

cmd/oci-runtime-tool/generate.go
```diff
@@ -73,7 +80,6 @@ var generateFlags = []cli.Flag{
 	cli.Uint64Flag{Name: "linux-mem-kernel-limit", Usage: "kernel memory limit (in bytes)"},
 	cli.Uint64Flag{Name: "linux-mem-kernel-tcp", Usage: "kernel memory limit for tcp (in bytes)"},
 	cli.Uint64Flag{Name: "linux-mem-swappiness", Usage: "how aggressive the kernel will swap memory pages (Range from 0 to 100)"},
-	cli.Int64Flag{Name: "linux-pids-limit", Usage: "maximum number of PIDs"},
```
Mashimiao: need rebased

GrantSeltzer: Woops, Thanks for catching that!

cmd/oci-runtime-tool/generate.go
GrantSeltzer: I went ahead and reordered it. While not in strict alphabetical order, I separated it logically so that the syscall action flags were together (and alphabetical), the default flags together, etc...

I think the way it is now makes the most sense.

Mashimiao: Personal opinion, separating them logically does not make too much sense. As seccomp related options all  start with `seccomp-*`, easily to see they are in one group and find the needed one.
I think Following one standard(alphabetical order) makes more sense. 
@wangkirin  @mrunalp  @liangchenye  any comments?

GrantSeltzer: Sure, I can concede that. Will fix now

Mashimiao:
@GrantSeltzer  can you add new options to `completions/bash/oci-runtime-tool`.
If you are not familiar with that, I can add them at #167 
And sorry for need another rebase.

GrantSeltzer:
Sure no problem

mrunalp:
@Mashimiao Does this look fine to you now?

Mashimiao:
Though there are some minimal issues(like part of new options not in alpha order in man), almost looks good.

GrantSeltzer:
@Mashimiao Just fixed the man page order

Mashimiao:
Thanks.
[60473d2](https://github.com/opencontainers/runtime-tools/pull/159/commits/60473d2edd6bc0ade85aea1ca762f5218fb81553)  looks good to me.

mrunalp:
LGTM

GrantSeltzer:
Thanks for the merge!

PR #160, opened 2016-07-28
generate: Remove superfluous err check from Save

wking:


wking:
Wrong PR target (trying out my preferred approach from #140).

PR #169, opened 2016-07-28
Modify generate API

hmeng-19:
This PR tries to make the generate API easier to use.  For example, instead of feeding a whole string to `AddAnnotation`, the new API allows the user to feed a key and a value to `AddAnnotation`.

Note: This PR does not touch any seccomp-relative function to avoid conflicts with Grant's work. 

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -244,14 +244,9 @@ func (g *Generator) ClearAnnotations() {
 }
 
 // AddAnnotation adds an annotation into g.spec.Annotations.
-func (g *Generator) AddAnnotation(s string) error {
+func (g *Generator) AddAnnotation(key, value string) error {
 	g.initSpecAnnotations()
-
-	pair := strings.Split(s, "=")
-	if len(pair) != 2 {
-		return fmt.Errorf("incorrectly specified annotation: %s", s)
-	}
-	g.spec.Annotations[pair[0]] = pair[1]
+	g.spec.Annotations[key] = value
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


hmeng-19:
@mrunalp , PTAL.

hmeng-19:
@mrunalp , to make it easy for you to test the PR, here is the test command for testing all the changed functions in this PR:

    ocitools generate --uidmappings 0:19:1000 --gidmappings 0:29:1032 --tmpfs /hmeng/test/tmp:rw,noexec --bind /hmeng/test11:/hmeng/testcontainer12 --bind /hmeng/test:/hmeng/testcontainer:rw --prestart ls:/dev:/run --poststart date:-u --poststop uname:-a --sysctl net.ipv4.forward=1 --sysctl myname=hmeng --label tag=mytag1 --output config.json

mrunalp:
LGTM

generate/generate.go
```diff
@@ -461,14 +456,9 @@ func (g *Generator) ClearLinuxSysctl() {
 }
 
 // AddLinuxSysctl adds a new sysctl config into g.spec.Linux.Sysctl.
-func (g *Generator) AddLinuxSysctl(s string) error {
+func (g *Generator) AddLinuxSysctl(key, value string) error {
```
Mashimiao: It seems this function does not need `return error` any more. Same as other AddXXXX() function.

generate/generate.go
```diff
@@ -244,14 +244,9 @@ func (g *Generator) ClearAnnotations() {
...
+func (g *Generator) AddAnnotation(key, value string) error {
 	g.initSpecAnnotations()
-
-	pair := strings.Split(s, "=")
-	if len(pair) != 2 {
-		return fmt.Errorf("incorrectly specified annotation: %s", s)
-	}
-	g.spec.Annotations[pair[0]] = pair[1]
+	g.spec.Annotations[key] = value
 	return nil
```
wking: With this pattern, there are now a whole bunch of methods which call `g.init…()` and then do some simple struct work.  I still think the API would be easier to use and maintain if we made the `init…` helpers and `g.Config` public and dropped the thin wrappers around struct manipulation.

generate/generate.go
hmeng-19: @Mashimiao , thanks for pointing out. Fixed it. PTAL.
I also change the param of `AddProcessAdditionalGid` from `string` to `uint32`.

hmeng-19:
@mrunalp , PTAL.

mrunalp:
LGTM

PR #44, opened 2016-04-28
man/ocitools-generate: Alphabetize the OPTIONS section

wking:
With `--uid` so far from `--gid`, it seems clear that alphabetical order
was the original goal.  Make some adjustments to bring the current
content back in line.

PR #175, opened 2016-08-02
add new generate options for setting cpu/mem

hmeng-19:

### DIFF
--- a/cmd/ocitools/generate.go
+++ b/cmd/ocitools/generate.go
@@ -59,6 +59,19 @@ var generateFlags = []cli.Flag{
 	cli.StringFlag{Name: "template", Usage: "base template to use for creating the configuration"},
 	cli.StringSliceFlag{Name: "label", Usage: "add annotations to the configuration e.g. key=value"},
 	cli.IntFlag{Name: "oom-score-adj", Usage: "oom_score_adj for the container"},
+	cli.Uint64Flag{Name: "linux-cpu-shares", Usage: "the relative share of CPU time available to the tasks in a cgroup"},
+	cli.Uint64Flag{Name: "linux-cpu-period", Usage: "the CPU period to be used for hardcapping (in usecs)"},
+	cli.Uint64Flag{Name: "linux-cpu-quota", Usage: "the allowed CPU time in a given period (in usecs)"},
+	cli.Uint64Flag{Name: "linux-realtime-runtime", Usage: "the time realtime scheduling may use (in usecs)"},
+	cli.Uint64Flag{Name: "linux-realtime-period", Usage: "CPU period to be used for realtime scheduling (in usecs)"},
+	cli.StringFlag{Name: "linux-cpus", Usage: "CPUs to use within the cpuset (default is to use any CPU available)"},
+	cli.StringFlag{Name: "linux-mems", Usage: "list of memory nodes in the cpuset (default is to use any available memory node)"},
+	cli.Uint64Flag{Name: "linux-mem-limit", Usage: "memory limit (in bytes)"},
+	cli.Uint64Flag{Name: "linux-mem-reservation", Usage: "memory reservation or soft limit (in bytes)"},
+	cli.Uint64Flag{Name: "linux-mem-swap", Usage: "total memory limit (memory + swap) (in bytes)"},
+	cli.Uint64Flag{Name: "linux-mem-kernel-limit", Usage: "kernel memory limit (in bytes)"},
+	cli.Uint64Flag{Name: "linux-mem-kernel-tcp", Usage: "kernel memory limit for tcp (in bytes)"},
+	cli.Uint64Flag{Name: "linux-mem-swappiness", Usage: "how aggressive the kernel will swap memory pages (Range from 0 to 100)"},
 }
 
 var generateCommand = cli.Command{
@@ -322,6 +335,58 @@ func setupSpec(g *generate.Generator, context *cli.Context) error {
 
 	if context.IsSet("oom-score-adj") {
 		g.SetLinuxResourcesOOMScoreAdj(context.Int("oom-score-adj"))
+	}
+
+	if context.IsSet("linux-cpu-shares") {
+		g.SetLinuxResourcesCPUShares(context.Uint64("linux-cpu-shares"))
+	}
+
+	if context.IsSet("linux-cpu-period") {
+		g.SetLinuxResourcesCPUPeriod(context.Uint64("linux-cpu-period"))
+	}
+
+	if context.IsSet("linux-cpu-quota") {
+		g.SetLinuxResourcesCPUQuota(context.Uint64("linux-cpu-quota"))
+	}
+
+	if context.IsSet("linux-realtime-runtime") {
+		g.SetLinuxResourcesCPURealtimeRuntime(context.Uint64("linux-realtime-runtime"))
+	}
+
+	if context.IsSet("linux-realtime-period") {
+		g.SetLinuxResourcesCPURealtimePeriod(context.Uint64("linux-realtime-period"))
+	}
+
+	if context.IsSet("linux-cpus") {
+		g.SetLinuxResourcesCPUCpus(context.String("linux-cpus"))
+	}
+
+	if context.IsSet("linux-mems") {
+		g.SetLinuxResourcesCPUMems(context.String("linux-mems"))
+	}
+
+	if context.IsSet("linux-mem-limit") {
+		g.SetLinuxResourcesMemoryLimit(context.Uint64("linux-mem-limit"))
+	}
+
+	if context.IsSet("linux-mem-reservation") {
+		g.SetLinuxResourcesMemoryReservation(context.Uint64("linux-mem-reservation"))
+	}
+
+	if context.IsSet("linux-mem-swap") {
+		g.SetLinuxResourcesMemorySwap(context.Uint64("linux-mem-swap"))
+	}
+
+	if context.IsSet("linux-mem-kernel-limit") {
+		g.SetLinuxResourcesMemoryKernel(context.Uint64("linux-mem-kernel-limit"))
+	}
+
+	if context.IsSet("linux-mem-kernel-tcp") {
+		g.SetLinuxResourcesMemoryKernelTCP(context.Uint64("linux-mem-kernel-tcp"))
+	}
+
+	if context.IsSet("linux-mem-swappiness") {
+		g.SetLinuxResourcesMemorySwappiness(context.Uint64("linux-mem-swappiness"))
 	}
 
 	var sd string
--- a/man/ocitools-generate.1.md
+++ b/man/ocitools-generate.1.md
@@ -87,6 +87,45 @@ read the configuration from `config.json`.
 **--label**=[]
   Add annotations to the configuration e.g. key=value.
 
+**--linux-cpu-shares**=CPUSHARES
+  Specifies a relative share of CPU time available to the tasks in a cgroup.
+
+**--linux-cpu-period**=CPUPERIOD
+  Specifies a period of time in microseconds for how regularly a cgroup's access to CPU resources should be reallocated (CFS scheduler only).
+
+**--linux-cpu-quota**=CPUQUOTA
+  Specifies the total amount of time in microseconds for which all tasks in a cgroup can run during one period.
+
+**--linux-realtime-runtime**=REALTIMERUNTIME
+  Specifies a period of time in microseconds for the longest continuous period in which the tasks in a cgroup have access to CPU resources.
+
+**--linux-realtime-period**=REALTIMEPERIOD
+  Sets the CPU period to be used for realtime scheduling (in usecs). Same as **--linux-cpu-period** but applies to realtime scheduler only.
+
+**--linux-cpus**=CPUS
+  Sets the CPUs to use within the cpuset (default is to use any CPU available).
+
+**--linux-mems**=MEMS
+  Sets the list of memory nodes in the cpuset (default is to use any available memory node).
+
+**--linux-mem-limit**=MEMLIMIT
+  Sets the limit of memory usage in bytes.
+
+**--linux-mem-reservation**=MEMRESERVATION
+  Sets the soft limit of memory usage in bytes.
+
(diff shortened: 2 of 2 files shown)


hmeng-19:
@mrunalp , PTAL.

hmeng-19:
The test command is here:
```
ocitools generate --output config.json --cpu-shares 512 \
   --cpu-period 10000 \
   --cpu-quota 20000 \
   --realtime-runtime 35000000 \
   --realtime-period 343439999 \
   --cpus 1-3 \
   --mems 0 \
   --mem-limit 999000000 \
   --mem-reservation 6660000 \
   --mem-swap 1000000000 \
   --mem-kernel-limit 450000000 \
   --mem-kernel-tcp 3000000 \
   --mem-swappiness 30
```

man/ocitools-generate.1.md
```diff
@@ -234,6 +234,45 @@ inside of the container.
   The special *PATH* `host` removes any existing UTS namespace from
   the configuration.
 
+**--cpu-shares**=CPUSHARES
+  Sets the CPU shares (relative weight (ratio) vs. other cgroups with cpu shares).
```
wking: Instead of having two descriptions to discuss, I'd rather just copy the runtime-spec wording verbatim here.  For this setting, [that's][1]:

> Specifies a relative share of CPU time available to the tasks in a cgroup.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/config-linux.md#cpu

wking:
On Tue, Aug 02, 2016 at 12:00:25PM -0700, hmeng-19 wrote:
> The test command is here:

As ‘ocitools generate’ grows, it's probalby worth adding a
bats/sharness/… test suite locally to exercise all these options for
us.  I think we've passed that tipping point already, and am happy to
PR something if ocitools maintainers agree (and have a test-suite
preference, otherwise I'll probably use sharness).

man/ocitools-generate.1.md
hmeng-19: @wking , thanks for pointing out.
In fact, the current description I am using is from https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/specs-go/config.go#L272 . 
Is there any plan on runtime-spec side to sync the descriptions on the config-linux.md and config.go?

wking: On Tue, Aug 02, 2016 at 12:28:18PM -0700, hmeng-19 wrote:
> In fact, the current description I am using is from
> https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/specs-go/config.go#L272
> .  Is there any plan on runtime-spec side to sync the descriptions
> on the config-linux.md and config.go?

Heh.  I'm not aware of a plan, but I'll file a PR updating the Go
comments this evening if nobody beats me to it.  And if you feel moved
to PR the update, don't wait for me ;).

hmeng-19: @wking , go ahead with the runtime-spec PR to fix it.

hmeng-19:
@wking , @mrunalp , PTAL.

man/ocitools-generate.1.md
```diff
@@ -234,6 +234,45 @@ inside of the container.
   The special *PATH* `host` removes any existing UTS namespace from
   the configuration.
 
+**--cpu-shares**=CPUSHARES
```
wking: Options in this man page are currently alphabetized (since #44).  That's an easier pattern to follow than grouping related options.

cmd/ocitools/generate.go
```diff
@@ -58,6 +58,19 @@ var generateFlags = []cli.Flag{
 	cli.StringSliceFlag{Name: "seccomp-errno", Usage: "specifies syscalls to be added to list that returns an error"},
 	cli.StringFlag{Name: "template", Usage: "base template to use for creating the configuration"},
 	cli.StringSliceFlag{Name: "label", Usage: "add annotations to the configuration e.g. key=value"},
+	cli.Uint64Flag{Name: "cpu-shares", Usage: "CPU shares (relative weight (ratio) vs. other cgroups with cpu shares)"},
```
wking: Probably update these descriptions to match the man page which matches the runtime-spec Markdown.

hmeng-19: @wking , I tried to do that and eventually gave up. 
Those descriptions are too long, and the output of `ocitools generate -h` will overwhelm.

man/ocitools-generate.1.md
hmeng-19: @wking , will fix it. Thanks.

wking: It may be worth using `--linux-cpu-shares` here to leave room for other platforms with their own interpretation of CPU shares.  That will help avoid confusion when someone PRs [`--solaris-ncpus`][1], etc.  It's possible that there is common ground here and we could have a cross-platform `--cpu-shares` option, but if that's the case I'd rather have runtime-spec leading the push towards unification.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/config-solaris.md#cappedcpu

cmd/ocitools/generate.go
wking: On Tue, Aug 02, 2016 at 01:36:31PM -0700, hmeng-19 wrote:
> > @@ -58,6 +58,19 @@ var generateFlags = []cli.Flag{
> >  	cli.StringSliceFlag{Name: "seccomp-errno", Usage: "specifies syscalls to be added to list that returns an error"},
> >  	cli.StringFlag{Name: "template", Usage: "base template to use for creating the configuration"},
> >  	cli.StringSliceFlag{Name: "label", Usage: "add annotations to the configuration e.g. key=value"},
> > +	cli.Uint64Flag{Name: "cpu-shares", Usage: "CPU shares (relative weight (ratio) vs. other cgroups with cpu shares)"},
> 
> @wking , I tried to do that and eventually gave up. 
> Those descriptions are too long, and the output of `ocitools generate -h` will overwhelm.

I'm fine with abbreviating/excerpting here (and in the Go docs).  But
the wording should at least be related.

mrunalp:
@wking :+1: to a test harness. I don't care what it is as long as we have something :wink:

man/ocitools-generate.1.md
hmeng-19: @mrunalp , how to think about this?

hmeng-19:
@wking , @mrunalp , PTAL.
Rename the options to be `--linux-*`;
Reorder the options in `man/ocitools-generate.1.md`.

hmeng-19:
The new test command is:
```
ocitools generate --output config.json \
   --linux-cpu-shares 512 \
   --linux-cpu-period 10000 \
   --linux-cpu-quota 20000 \
   --linux-realtime-runtime 35000000 \
   --linux-realtime-period 343439999 \
   --linux-cpus 1-3 \
   --linux-mems 0 \
   --linux-mem-limit 999000000 \
   --linux-mem-reservation 6660000 \
   --linux-mem-swap 1000000000 \
   --linux-mem-kernel-limit 450000000 \
   --linux-mem-kernel-tcp 3000000 \
   --linux-mem-swappiness 30
```

mrunalp:
Needs rebase?

hmeng-19:
@mrunalp , rebased it. PTAL.

man/ocitools-generate.1.md
```diff
@@ -84,6 +84,45 @@ inside of the container.
...
+  Sets the CPUs to use within the cpuset (default is to use any CPU available).
+
+**--linux-mems**=MEMS
+  Sets the list of memory nodes in the cpuset (default is to use any available memory node).
+
+**--linux-mem-limit**=MEMLIMIT
+  Sets the limit of memory usage in bytes.
+
+**--linux-mem-reservation**=MEMRESERVATION
+  sets the soft limit of memory usage in bytes.
```
wking: “sets” → “Sets”

man/ocitools-generate.1.md
```diff
@@ -84,6 +84,45 @@ inside of the container.
...
+  Sets the total memory limit (memory + swap) in bytes.
+
+**--linux-mem-kernel-limit**=MEMKERNELLIMIT
+  Sets the hard limit of kernel memory in bytes.
+
+**--linux-mem-kernel-tcp**=MEMKERNELTCP
+  Sets the hard limit of kernel TCP buffer memory in bytes.
+
+**--linux-mem-swappiness**=MEMSWAPPINESS
+  Sets the swappiness of how the kernel will swap memory pages (Range from 0 to 100).
```
wking: If we're going to specify a range here, we probably want code in `SetLinuxResourcesMemorySwappiness` to enforce it (e.g. erroring out for values > 100).

man/ocitools-generate.1.md
hmeng-19: Fixed it.

man/ocitools-generate.1.md
hmeng-19: @wking , good points. In fact, there are several other fields having ranges we can enforce in the generate library.
Do you mind of opening an issue on it? I will fix it in a separate PR. This PR just aims to add the  command options.

hmeng-19: @wking, @mrunalp  , I am not sure how much overlap `ocitools/generate` and `ocitools/validate` should have in these fields. 
We may want to have a wider reflection about the functionalities of these two `ocitools` subcommands. Anyway, I think opening an issue should a good start.

man/ocitools-generate.1.md
```diff
@@ -84,6 +84,45 @@ inside of the container.
   The special *PATH* `host` removes any existing IPC namespace from the
   configuration.
 
+**--linux-cpu-shares**=CPUSHARES
+  Specifies a relative share of CPU time available to the tasks in a cgroup.
+
+**--linux-cpu-period**=CPUPERIOD
+  Specifies a period of time in microseconds for how regularly a cgroup's access to CPU resources should be reallocated (CFS scheduler only).
```
wking: “CFS only” seems like we want a `--host-specific` guard to error out if someone tries to set it with `--host-specific` when they aren't using a CFS scheduler.  But after reading over [sched(7)][1] and [sched_setscheduler(2)][2], I'm not quite sure how to do that.  It looks like you check for a `SCHED_OTHER` `policy`, but `sched_getscheduler` is per-thread.  Is the scheduler policy really set on a per-thread basis?  Is a host-specific check just supposed to look for internal consistency (e.g. the config doesn't set both a CFS-only and a realtime-only parameter)?

[1]: http://man7.org/linux/man-pages/man7/sched.7.html
[2]: http://man7.org/linux/man-pages/man2/sched_setscheduler.7.html

man/ocitools-generate.1.md
wking: On Fri, Aug 05, 2016 at 10:49:39AM -0700, hmeng-19 wrote:
> > +**--linux-mem-swappiness**=MEMSWAPPINESS
> > +  Sets the swappiness of how the kernel will swap memory pages (Range from 0 to 100).
> 
> @wking , good points. In fact, there are several other fields having ranges we can enforce in the generate library.
> Do you mind of opening an issue on it?

Filed as #186.

man/ocitools-generate.1.md
hmeng-19: @wking , I think there are two ways we can use to determine the current scheduler.
1) Read the kernel config file, refer to https://en.wikipedia.org/wiki/Completely_Fair_Scheduler:
```
[hmeng@localhost boot]$ grep CONFIG_SCHED_AUTOGROUP /boot/config-4.6.4-201.fc23.x86_64 
CONFIG_SCHED_AUTOGROUP=y
```
2) check the /sys/fs/cgroup/cpu dir to see whether `cpu.cfs_period_us` exists or not.

wking: On Fri, Aug 05, 2016 at 11:58:48AM -0700, hmeng-19 wrote:
> @wking , I think there are two ways we can use to determine the current scheduler.
> 1) Read the kernel config file, refer to https://en.wikipedia.org/wiki/Completely_Fair_Scheduler:
> ```
> [hmeng@localhost boot]$ grep CONFIG_SCHED_AUTOGROUP /boot/config-4.6.4-201.fc23.x86_64 
> CONFIG_SCHED_AUTOGROUP=y
> ```
> 2) check the /sys/fs/cgroup/cpu dir to see whether `cpu.cfs_period_us` exists or not.

These both sound like “does your kernel support that scheduler”, but
if scheduler policy is per-thread, then how do you know the container
will *use* that available scheduler (and not another available
scheduler)?

hmeng-19: @wking , @mrunalp , do correct me if I am wrong.
I always thought the scheduling policy is a system-level config. It is decided when you built your kernel. Right?

wking: On Fri, Aug 05, 2016 at 08:25:27PM -0700, hmeng-19 wrote:
> > +**--linux-cpu-period**=CPUPERIOD
> > +  Specifies a period of time in microseconds for how regularly a cgroup's access to CPU resources should be reallocated (CFS scheduler only).
> 
> @wking , @mrunalp , do correct me if I am wrong.
> I always thought the scheduling policy is a system-level config. It
> is decided when you built your kernel. Right?

That was my impression before poking around in the man pages today.
But see also [1], which has:

  Each process or thread is controlled by an associated scheduling
  policy and priority…

And [2] has:

  If at least one RT task is running, no other SCHED_OTHER task will
  be allowed to run in any CPU.

which makes it fairly clear that you can have threads/processes with
different schedulers running concurrently.  I'm not sure how that
works for the cgroup settings though.  Maybe you can set them all
(both real-time and CFS parameters) and the kernel will apply the
appropriate limits to each task depending on its policy?  In that
case, configs that specify settings for multiple schedulers are fair
game.

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sched.h.html
[2]: http://www.linuxjournal.com/article/3910

mrunalp:
LGTM

Mashimiao:
@hmeng-19  seems need rebase

hmeng-19:
@Mashimiao , @mrunalp , PTAL.

mrunalp:
LGTM

PR #176, opened 2016-08-02
support setting oom_score_adj

hmeng-19:

### DIFF
--- a/cmd/ocitools/generate.go
+++ b/cmd/ocitools/generate.go
@@ -58,6 +58,7 @@ var generateFlags = []cli.Flag{
 	cli.StringSliceFlag{Name: "seccomp-errno", Usage: "specifies syscalls to be added to list that returns an error"},
 	cli.StringFlag{Name: "template", Usage: "base template to use for creating the configuration"},
 	cli.StringSliceFlag{Name: "label", Usage: "add annotations to the configuration e.g. key=value"},
+	cli.IntFlag{Name: "oom-score-adj", Usage: "oom_score_adj for the container"},
 }
 
 var generateCommand = cli.Command{
@@ -317,6 +318,10 @@ func setupSpec(g *generate.Generator, context *cli.Context) error {
 		}
 
 		g.AddLinuxGIDMapping(hid, cid, size)
+	}
+
+	if context.IsSet("oom-score-adj") {
+		g.SetLinuxResourcesOOMScoreAdj(context.Int("oom-score-adj"))
 	}
 
 	var sd string
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -362,6 +362,12 @@ func (g *Generator) SetLinuxMountLabel(label string) {
 	g.spec.Linux.MountLabel = label
 }
 
+// SetLinuxResourcesOOMScoreAdj sets g.spec.Linux.Resources.OOMScoreAdj.
+func (g *Generator) SetLinuxResourcesOOMScoreAdj(adj int) {
+	g.initSpecLinuxResources()
+	g.spec.Linux.Resources.OOMScoreAdj = &adj
+}
+
 // SetLinuxResourcesCPUShares sets g.spec.Linux.Resources.CPU.Shares.
 func (g *Generator) SetLinuxResourcesCPUShares(shares uint64) {
 	g.initSpecLinuxResourcesCPU()
--- a/man/ocitools-generate.1.md
+++ b/man/ocitools-generate.1.md
@@ -117,6 +117,9 @@ inside of the container.
   using tools like setuid apps.  It is a good idea to run unprivileged
   containers with this flag.
 
+**--oom-score-adj**=adj
+  Specifies oom_score_adj for the container.
+
 **--output**=PATH
   Instead of writing the configuration JSON to stdout, write it to a
   file at *PATH* (overwriting the existing content if a file already


hmeng-19:
@mrunalp , PTAL.

wking:
Hmm.  Is there a way to clear oom_score_adj with this interface?  Does
IntFlag support explicitly setting a nil value (e.g. ‘--oom-score-adj=’
with no argument).

mrunalp:
IIRC, =FALSE works for setting nil value.

mrunalp:
@wking nvm, I misunderstood your question.

mrunalp:
LGTM

wking:
On Tue, Aug 02, 2016 at 02:25:07PM -0700, Mrunal Patel wrote:
> @wking nvm, I misunderstood your question. 

I haven't tested =FALSE, but I think we want to figure out how we
intend to handle clearing values before we dive too deeply into the
IntFlag approach.

cmd/ocitools/generate.go
```diff
@@ -58,6 +58,7 @@ var generateFlags = []cli.Flag{
 	cli.StringSliceFlag{Name: "seccomp-errno", Usage: "specifies syscalls to be added to list that returns an error"},
 	cli.StringFlag{Name: "template", Usage: "base template to use for creating the configuration"},
 	cli.StringSliceFlag{Name: "label", Usage: "add annotations to the configuration e.g. key=value"},
+	cli.IntFlag{Name: "oom-score-adj", Usage: "specify oom_score_adj for the container"},
```
Mashimiao: specify->specifies
I think usage should keep same with man

hmeng-19:
> Is there a way to clear oom_score_adj with this interface?

@wking, I am not sure what do you really want here. 
The range of `oom_score_adj` can be -1000 to 1000. 
If you want to increase the badness score of a process, just set a larger value.
If you want to decrease the badness score of a process, just set to a smaller value.
If you do not want to mess up with the badness score at all, just set it to be 0.

cmd/ocitools/generate.go
hmeng-19: @Mashimiao , I was struggling with this when I made the change.
The man doc always use `specifies`, `adds`, `uses`. However, most of flags defined in cmd/ocitools/generate.go are using `specify, `add`, `use`.

I will first change it to `a phrase` instead of `a verb + a noun`.

hmeng-19:
@mrunalp , @Mashimiao , PTAL.

mrunalp:
Still LGTM

wking:
GitHub seems to have eaten my emailed comment, so re-posting via the web UI:

On Wed, Aug 03, 2016 at 07:36:34AM -0700, hmeng-19 wrote:
> > Is there a way to clear oom_score_adj with this interface?
>
> @wking, I am not sure what do you really want here.
> …
> If you do not want to mess up with the badness score at all, just
> set it to be 0.

`Resources.OOMScoreAdj` is an `*int`, and it's a pointer (like many
other optional settings) because nil is clearly “don't mess with this”
while 0 is sometimes “don't mess with this” and sometimes not.  With
your current implementation (1915407):

    $ ./ocitools generate --template <(echo '{"linux": {"resources": {"oomScoreAdj": 5}}}') --oom-score-adj 0 | jq .linux
    {
      "resources": {
        "devices": null,
        "oomScoreAdj": 0
      }
    }

But I want a way to set the `*int` to nil so it doesn't get printed at
all (I've filed opencontainers/runtime-spec#526 to address the useless
‘devices’ entry).

wking:
On Wed, Aug 03, 2016 at 07:36:34AM -0700, hmeng-19 wrote:
> > Is there a way to clear oom_score_adj with this interface?
> 
> @wking, I am not sure what do you really want here. 
> …
> If you do not want to mess up with the badness score at all, just
> set it to be 0.

`Resources.OOMScoreAdj` is an `*int`, and it's a pointer (like many
other optional settings) because nil is clearly “don't mess with this”
while 0 is sometimes “don't mess with this” and sometimes not.  With
your current implementation (1915407):

  $ ./ocitools generate --template <(echo '{"linux": {"resources": {"oomScoreAdj": 5}}}') --oom-score-adj 0 | jq .linux
  {
    "resources": {
      "devices": null,
      "oomScoreAdj": 0
    }
  }

But I want a way to set the `*int` to nil so it doesn't get printed at
all (I've filed opencontainers/runtime-spec#526 to address the useless
‘devices’ entry).

hmeng-19:
@wking , first I am not sure we want to support the use cases you pointed out or not. If I understand correctly, you want to remove some fields from a template completely. Right?

I would suggest always feed the minimal config.json as the template, and then only add new fields into it.

If you really want to remove fields from a template, then we probably should add some new flags like `--remove-oom-score-adj`.

hmeng-19:
> GitHub seems to have eaten my emailed comment, so re-posting via the web UI:

@wking , it seems github was slow this morning. I noticed there was a 40 minutes delay before I got gmail notifications for others' comments.

wking:
On Wed, Aug 03, 2016 at 10:56:48AM -0700, hmeng-19 wrote:
> I would suggest always feed the minimal config.json as the template,
> and then only add new fields into it.

I think ‘ocitools generate --template ...’ is positioning itself to be
a generic tool for configuration edits, although I'm fine also using
jq for this sort of thing.  If you are using ocitools to edit an
existing config, the source config may be provided from a third party
who doesn't realize you won't want oom_score_adj.

> If you really want to remove fields from a template, then we
> probably should add some new flags like `--remove-oom-score-adj`.

That's one approach, and the implementation might be more
straightforward.  I think the command line API would be more
straightforward if it supported ‘--oom-score-adj=’, but that doesn't
work with IntFlag.  Because the IntFlag pattern is young (for
ocitools), I think now is a good time to decide:

1. Do we want to support setting nil for *int-type fields in general?
2. If we do, do we want the pattern to be ‘--remove-abc’ or ‘--abc=’?

With jq being a convenient, generic solution to config edits, I think
all of the possible choices are defensible.  But I'd like to see
explicit policy choices before forging ahead with the implementation
so we stay consistent.

hmeng-19:
@mrunalp , how do you think about this?

mrunalp:
>  1. Do we want to support setting nil for *int-type fields in general?

Yes. But, it shouldn't block PRs while we figure it out as starting with minimal config.json is a workaround.

>  2. If we do, do we want the pattern to be ‘--remove-abc’ or ‘--abc=’?

I think it would be good if we avoid adding new flags, but I don't see how --abc= will work for setting to nil unless the library we are using supports it by special treating some text like NIL to mean make it nil.

hmeng-19:
@wking , I tried to use the `--oom-score-adj=`, it would fail:
```
[hmeng@localhost c4]$ ocitools generate --output config.json --oom-score-adj=
FATA[0000] invalid value "" for flag -oom-score-adj: strconv.ParseInt: parsing "": invalid syntax 
```

wking:
On Wed, Aug 03, 2016 at 11:40:01AM -0700, Mrunal Patel wrote:
> >  2. If we do, do we want the pattern to be ‘--remove-abc’ or ‘--abc=’?
>
> I think it would be good if we avoid adding new flags, but I don't
> see how --abc= will work for setting to nil unless the library we
> are using supports it by special treating some text like NIL to mean
> make it nil.

I think that means we want to use a StringFlag here and add our own
StringFlag → *int, helpers.

hmeng-19:
> I think that means we want to use a StringFlag here and add our own
StringFlag → *int, etc. helpers.

@wking , I would suggest we first merge PR. 
Then we can open a new issue to discuss about which flags are going to be converted from IntFlag to StringFlag, or other possible solutions to tackle the problem here.

wking:
On Wed, Aug 03, 2016 at 12:02:36PM -0700, hmeng-19 wrote:
> > I think that means we want to use a StringFlag here and add our
> > own StringFlag → *int, etc. helpers.
> 
> @wking , I would suggest we first merge PR.  Then we can open a new
> issue to discuss about which flags are going to be converted from
> IntFlag to StringFlag, or other possible solutions to tackle the
> problem here.

If you like.  I don't see much need to push this through if we already
know we're going to revisit/discuss.  It isn't a big deal either way,
it's just more churn to backport to v1.0.0.rc1 if we merge fixup PRs
;).

hmeng-19:
@mrunalp , I think we can merge this PR for now, so that we can vendor it into ocid to set up oom_score_adj.

mrunalp:
@wking By using StringFlag we lose some of the checks that we get from the CLI library.

wking:
On Wed, Aug 03, 2016 at 03:37:05PM -0700, Mrunal Patel wrote:
> @wking By using StringFlag we lose some of the checks that we get
> from the CLI library.

So put the checks in our helper?  Are the checks really that
complicated?

mrunalp:
@wking Yeah, we could look into that. We can open an issue to track this.

mrunalp:
LGTM (we can figure out the CLI conversion on the side).

PR #180, opened 2016-08-03
test: Stub in sharness framework

wking:
Spun off from [this](https://github.com/opencontainers/ocitools/pull/175#issuecomment-237010187) [discussion](https://github.com/opencontainers/ocitools/pull/175#issuecomment-237051882).  This currently ignores most of the `generate` options and doesn't touch `validate` at all, but I thought I'd float it early for some feedback before sinking too much time into it.  If the current work looks solid, I'm happy to land it now and follow up with additional PRs to fill it out, or to keep this PR open until the coverage is more complete.

Details on the individual changes in the commit messages.

PR #188, opened 2016-08-09
add validate_test.go to do unit test

liangchenye:
Only add a test case on  checkSemVer in this PR.


wking:
On Tue, Aug 09, 2016 at 04:17:40AM -0700, 梁辰晔 (Liang Chenye) wrote:
> -- File Changes --
> 
>     A cmd/ocitools/validate_test.go (27)

You probably want to add a Makefile and/or .travis.yml entry to run
these tests from Travis.

cmd/ocitools/validate_test.go
```diff
@@ -0,0 +1,27 @@
...
+	}{
+		{rspec.Version, 0},
+		{"0.0.1", 1},
+		{"invalid", 2},
+	}
+
+	for _, c := range cases {
+		spec.Version = c.v
+		msg := checkSemVer(spec, "", false)
+		if len(msg) != c.expected {
```
wking: Is “number of error messages” really what we want to be checking here?  I'd rather have a bunch of potential version strings with true/false for whether we thought they were valid, and error if we thought a version was valid and `checkSemVer` gave us error messages at all (logging the error messages with `t.Fatalf`) or if we thought a version was invalid and `checkSemVer` did not give us error messages.

wking:
On Tue, Aug 09, 2016 at 04:17:40AM -0700, 梁辰晔 (Liang Chenye) wrote:
>     A cmd/ocitools/validate_test.go (27)

While I'm in favor of Go tests for our implementation, I'm not sure
it's worth testing cmd/… helpers.  I'd rather see the validation code
pulled out into a Go library (like generate/…) and test *that*.  Then
the cmd/… stuff should be a thin wrapper around that API, the API is
tested with ‘go test’ like you have here, and the command-line wrapper
is tested with sharness like I've floated in #180.

We'll also want a way to distinguish between “that's not valid” errors
[1] (which are like HTTP 400s) and “that may be valid, but we can't
check it yet” errors [2] (which are like HTTP 500s).  My preferred
approach for that is to use Python's unittest and skip which lets you
handle generic JSON easily (vs. the current Go based on the
runtime-spec types or a hypothetical Go framework based on
interface{}).  More on that in #98.

[1]: https://github.com/opencontainers/ocitools/blob/03e8b89f9eccc62d36f742a0af737adbfdcea635/cmd/ocitools/validate.go#L136-L139
[2]: https://github.com/opencontainers/ocitools/blob/03e8b89f9eccc62d36f742a0af737adbfdcea635/cmd/ocitools/validate.go#L140-L142

liangchenye:
I think it is a good idea to move 'validate' to a new directory like 'generate', I'll make a patch to move that.

(just back from a long time leaving)

cmd/ocitools/validate_test.go
liangchenye: You are right, it is not necessary to check the number of error messages.

liangchenye:
commit 82d0b38 moves validate directory out of cmd directory and add a 'validator' struct just like 'generator'.

commit 8b7ba22 adds a new unit test style based on Wking's comment

@wking @mrunalp  PTAL

validate/validate_test.go
```diff
@@ -0,0 +1,32 @@
...
+	}
+}
+
+func TestCheckSemVer(t *testing.T) {
+	cases := []struct {
+		val      string
+		expected bool
+	}{
+		{rspec.Version, true},
+		{"0.0.1", false},
```
wking: 0.0.1 looks like valid SemVer to me?  What's the error message?

initlove: Currently only support rspec.Version.
```
if version != rspec.Version {
		msgs = append(msgs, fmt.Sprintf("internal error: validate currently only handles version %s, but the supplied configuration targets %s", rspec.Version, version))
	}
```

wking: On Mon, Sep 05, 2016 at 03:38:57AM -0700, David Liang wrote:
> > +		{"0.0.1", false},
> 
> Currently only support rspec.Version.

So maybe leave a comment on this line that we're confirming a
shortcoming of the current test suite with this entry.

liangchenye: Thanks, I added a 'FIXME' to mention this.

validate/validate.go
```diff
@@ -0,0 +1,574 @@
...
+		"CAP_SYS_CHROOT",
+		"CAP_KILL",
+		"CAP_AUDIT_WRITE",
+	}
+)
+
+type Validator struct {
+	spec         rspec.Spec
+	rootfs       string
+	hostSpecific bool
```
wking: I'd rather [make][1] [these][2] [properties][3] [public][4], and `Generator.HostSpecific` *is* public.  Do you have any reason to keep any of them private?

[2]: https://github.com/opencontainers/ocitools/pull/138#discussion_r71912190
[3]: https://github.com/opencontainers/ocitools/pull/144#issuecomment-235299376
[4]:https://github.com/opencontainers/ocitools/pull/169/files/a217d8c86c9349c4e306bcfbaa92ca6109325895#r72735773

validate/validate.go
```diff
@@ -0,0 +1,574 @@
...
+		"CAP_NET_BIND_SERVICE",
+		"CAP_SYS_CHROOT",
+		"CAP_KILL",
+		"CAP_AUDIT_WRITE",
+	}
+)
+
+type Validator struct {
+	spec         rspec.Spec
+	rootfs       string
```
wking: This isn't the right property for `Validator`, since it includes information from the spec's `Root.Path`.  We should have `Validator.BundlePath` instead.

validate/validate.go
```diff
@@ -0,0 +1,574 @@
...
+		"CAP_SETPCAP",
+		"CAP_NET_BIND_SERVICE",
+		"CAP_SYS_CHROOT",
+		"CAP_KILL",
+		"CAP_AUDIT_WRITE",
+	}
+)
+
+type Validator struct {
+	spec         rspec.Spec
```
wking: We should probably namespace this in case we want a [state JSON][1] validator later, since validating state JSON is going to be part of runtime certification.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/runtime.md#state

validate/validate.go
liangchenye: make sense

validate/validate.go
liangchenye: Do we need to see and modify these? I think the 'Check*' functions are enough.

cmd/ocitools/validate.go
```diff
@@ -98,531 +54,15 @@ var bundleValidateCommand = cli.Command{
 			return fmt.Errorf("The root path %q is not a directory.", rootfsPath)
```
wking: I think we want to move the UTF-8 checking and such into the library as well.  We'd have `NewValidator` ([or `NewConfigValidator`][1], or whatever):

    func NewValidator(spec rspec.Spec, bundlePath string, hostSpecific bool) Validator

and also:

    func NewValidatorFromPath(bundlePath string, hostSpecific bool) Validator

which did the read-in, UTF-8 check, etc., and then called `NewValidator`.

[1]: https://github.com/opencontainers/ocitools/pull/188/files#r77942970

wking: Alternatively, use:

    func NewValidator(spec *rspec.Spec, bundlePath string, hostSpecific bool) Validator

and load from `{bundlePath}/config.json` if `spec` is `nil`.

validate/validate.go
wking: I think we want to use a pointer here to match `Generator.spec` (which I think should be [public][1] too ;).

[1]: https://github.com/opencontainers/ocitools/pull/188/files#r77942243

validate/validate.go
wking: On Wed, Sep 07, 2016 at 09:04:44PM -0700, 梁辰晔 (Liang Chenye) wrote:
> > +type Validator struct {
> > +	spec         rspec.Spec
> > +	rootfs       string
> > +	hostSpecific bool
>
> Do we need to see and modify these? I think the 'Check*' functions
> are enough.

Making them public means you don't need getters/setters and you can
still support users who want to partially setup the instance when they
call NewValidator (or whatever [1]) and finish the setup later.  Or if
they want to peer inside to get more details on why their validation
is failing.  Or who knows?  I'd rather err on the side of “we're all
consenting adults” [2], because I don't see a way to make validation
go terribly wrong if a user gets or sets one of these values whenever
they like.

[1]: https://github.com/opencontainers/ocitools/pull/188/files#r77942970
[2]: https://github.com/opencontainers/ocitools/pull/138#discussion_r71917014

wking: A downside to making things public is that changing them breaks your
public API.  But this is going to be a stable, boring implementation
based on a stable, boring spec right ;)?  I don't expect lots of
property churn as we figure out what we need to validate
configurations, so I'm not worried about the slight increase in API
surface area dragging us down with backwards-compat issues.

validate/validate.go
```diff
@@ -0,0 +1,574 @@
...
+		for _, env := range hook.Env {
+			if !envValid(env) {
+				msgs = append(msgs, fmt.Sprintf("Env %q for hook %v is in the invalid form.", env, hook.Path))
+			}
+		}
+	}
+
+	return
+}
+func (v *Validator) CheckHooks() (msgs []string) {
```
wking: There's unnecessary churn in:

    $ diff -u <(git cat-file -p origin/pr/188^^:cmd/ocitools/validate.go) <(git cat-file -p origin/pr/188^:validate/validate.go)

because you shifted this from being before `checkEventHooks` to being after it.  Can we move it back up?

validate/validate.go
```diff
@@ -0,0 +1,574 @@
...
+	}
+
+	return
+}
+
+func supportedMountTypes(OS string, hostSpecific bool) (map[string]bool, error) {
+	supportedTypes := make(map[string]bool)
+
+	if OS != "linux" && OS != "windows" {
+		logrus.Warnf("%v is not supported to (v *Validator) Check mount type", OS)
```
wking: Rouge find/replace?  The old warning was:

> "%v is not supported to check mount type"

I don't see a need to mess with that in this PR.

validate/validate.go
```diff
@@ -0,0 +1,574 @@
...
+	case "shared":
+	case "rshared":
+	default:
+		msgs = append(msgs, "rootfsPropagation must be empty or one of \"private|rprivate|slave|rslave|shared|rshared\"")
+	}
+
+	return
+}
+
+func (v *Validator) checkLinuxResources() (msgs []string) {
```
wking: This might as well be public, so folks who want to can check it without having to run all of `CheckLinux`.

validate/validate.go
```diff
@@ -0,0 +1,574 @@
...
+		}
+		if r.Memory.Limit != nil && r.Memory.Reservation != nil && uint64(*r.Memory.Reservation) > uint64(*r.Memory.Limit) {
+			msgs = append(msgs, fmt.Sprintf("Minimum memory limit should be larger than memory reservation"))
+		}
+	}
+
+	return
+}
+
+func (v *Validator) checkSeccomp() (msgs []string) {
```
wking: I think this [should be public too][1], and likely a number of similar helpers as well.  Stuff like `checkEventHooks` that are called multiple times for a single spec can stay private.

[1]: https://github.com/opencontainers/ocitools/pull/188/files#r77944056

cmd/ocitools/validate.go
liangchenye: I'll use 
`func NewValidatorFromPath(bundlePath string, hostSpecific bool) Validator` .

validate/validate.go
liangchenye: I use RootfsPath.   RootfsPath = BundlePath  join Root.Path .

wking: On Wed, Sep 07, 2016 at 11:38:55PM -0700, 梁辰晔 (Liang Chenye) wrote:
> > +type Validator struct {
> > +	spec         rspec.Spec
> > +	rootfs       string
> 
> I use RootfsPath.   RootfsPath = BundlePath  join Root.Path .

You don't want to use RootfsPath because that's not DRY [1].  What
happens if you change v.spec.Root.Path?  Do you update RootfsPath too?
It's better to restrict the param to the purely-external-to-config
bundle path and then do that path-join inside any validation methods
that need it.

[1]: https://github.com/opencontainers/ocitools/pull/188#discussion_r77942370

liangchenye:
Hi @wking , I modified codes according to your suggestion.  
Just one difference:  using RootfsPath other than BundlePath in Validator{} .

validate/validate.go
```diff
@@ -0,0 +1,610 @@
...
+	if err = json.Unmarshal(content, &spec); err != nil {
+		return Validator{}, err
+	}
+
+	rootfsPath := path.Join(bundlePath, spec.Root.Path)
+	if fi, err := os.Stat(rootfsPath); err != nil {
+		return Validator{}, fmt.Errorf("Cannot find the root path %q", rootfsPath)
+	} else if !fi.IsDir() {
+		return Validator{}, fmt.Errorf("The root path %q is not a directory.", rootfsPath)
+	}
```
wking: This “does the rootfs exist” check should be a separate check in `Validator.CheckAll()`

liangchenye: done in new commit!

validate/validate.go
liangchenye: I changed `spec` and `bundlePath` to private members in the new commit.

wking: On Thu, Sep 08, 2016 at 12:09:20AM -0700, 梁辰晔 (Liang Chenye) wrote:
> > +type Validator struct {
> > +	spec         rspec.Spec
> > +	rootfs       string
>
> I'm think it will be tricky if a user changes `BundlePath`, `Spec`
> need to be updated accordingly.  So I think `Generator` is right, it
> only publics `HostSpecific`.

I don't see why you'd need to update Spec if you change BundlePath.
Can you elaborate?  Maybe the user wants to validate several rootfses
(or whatever) against the same config, so they:

  validator := &Validator{
    Spec: spec,
    HostSpecific: true
  }
  for _, bundlePath := range bundles {
    validator.BundlePath = bundlePath
    msg = append(msgs, validator.CheckAll())
  }

liangchenye: `Spec` (config.json) is under `BundlePath`,  if `BundlePath` changes, we need to reload `Spec`.

wking: On Thu, Sep 08, 2016 at 05:44:00PM -0700, 梁辰晔 (Liang Chenye) wrote:
> > +type Validator struct {
> > +	spec         rspec.Spec
> > +	rootfs       string
>
> `Spec` (config.json) is under `BundlePath`, if `BundlePath` changes,
> we need to reload `Spec`.

I will be so happy if/when we land something like
opencontainers/runtime-spec#423 ;).  I think having a “bundle” concept
adds more confusion than it's worth.  What we're really validating
here is the config.  BundlePath just tells you how to interpret
relative paths (as made explicit by opencontainers/runtime-spec#558).
So I don't think we *need* to reload the spec if BundlePath changes
(although a caller certainly could if they wanted to).  Not reloading
the spec allows you to do things like [1].

[1]: https://github.com/opencontainers/ocitools/pull/188#discussion_r78033132

liangchenye: Now I understand, you want to remove 'bundle' concept :)
But since 'BundlePath' is the entry to a user in current runtime-spec, you case is better in this way  (at least now).:
```
  for _, bundlePath := range bundles {
     v, _ := NewValidatorFromPath( bundlePath, true)
    msg = append(msgs, v.CheckAll())
  }
```

So I prefer my current commit#06163fe.

liangchenye: I think 'bundle' is sitting between 'image-spec' and 'runtime-spec'.
It is the entrance to a runtime, so to me, I +1 for not removing it from 'runtime-spec'.

wking: On Thu, Sep 08, 2016 at 10:39:13PM -0700, 梁辰晔 (Liang Chenye) wrote:
> But since 'BundlePath' is the entry to a user in current
> runtime-spec, you case is better in this way (at least now).:
> ```
>   for _, bundlePath := range bundles {
>      v, _ := NewValidatorFromPath( bundlePath, true)
>     msg = append(msgs, v.CheckAll())
>   }
> ```

This will load a new config from each bundle directory.  Maybe I don't
even have a bundle and am looking to validate some property of a
rootfs.  So I want to use *the same config* and just expand some of
it's relative paths from a different working directory.  BundlePath is
just "the working directory from which some config relative links are
expanded".

liangchenye: We can use `NewValidator(spec, bundlePath, hostSpecific)` in this case.

But I even want to make `NewValidator` private.
Because it might make user an impression that the `config file` and the `bundlePath` are two independent things.  It is not what current OCI bundle mean. BundlePath is "the working directory of config file and some config relative links".

wking: On Fri, Sep 09, 2016 at 05:41:31AM -0700, 梁辰晔 (Liang Chenye) wrote:
> Because it might make user an impression that the `config file` and
> the `bundlePath` are two independent things.

I think they *are* two independent things.  But I'm fine punting this
discussion until something like opencontainers/runtime-spec#423 lands,
which may be never.

liangchenye: @wking  So I'll keep current way now.

@mrunalp  PTAL

liangchenye:
Rebase by using 'runtime-tools'.
Merge two commits (one for restruct and one for unit test case) into one. The original change log can not reflect the real meaning, code changes a lot after discussions with @wking.

@mrunalp PTAL

validate/validate.go
```diff
@@ -0,0 +1,617 @@
...
+	spec         *rspec.Spec
+	bundlePath   string
+	HostSpecific bool
+}
+
+func NewValidator(spec *rspec.Spec, bundlePath string, hostSpecific bool) Validator {
+	return Validator{spec: spec, bundlePath: bundlePath, HostSpecific: hostSpecific}
+}
+
+func NewValidatorFromPath(bundlePath string, hostSpecific bool) (Validator, error) {
```
wking: I think these `NewValidator*` functions should return `(*Validator, error)`.

wking:
a8c02cf looks good enough to me (I think we've agreed to punt on
everything I'd still like to change).

liangchenye:
Travis fails because of:
https://github.com/opencontainers/runtime-tools/pull/230

liangchenye:
#7cd3aa1 rebased on top of current 'master'.
Only difference with '#a8c02cf ' is '7cd3aa1' includes Mashimiao's new patch:
'validate: add hard and soft limit check for rlimit'
https://github.com/opencontainers/runtime-tools/commit/e129a67abe52e8e8c7f0c343fc80bfa86645f05f

@Mashimiao @mrunalp PTAL

vbatts:
so the test fail is due to go1.5?

liangchenye:
@vbatts yes, I'll rebuild the CI.
```
The command "go get github.com/golang/lint/golint" failed and exited with 2 during
```

liangchenye:
@mrunalp @Mashimiao  rebased.
Merge new IsAbs and `const 'config.json'` changes.

It is a code struct change, please PTAL first :)

Mashimiao:
`bde9989` looks good to me

mrunalp:
LGTM

PR #192, opened 2016-08-17
validation: add check for namespace path

Mashimiao:

### DIFF
--- a/cmd/ocitools/validate.go
+++ b/cmd/ocitools/validate.go
@@ -337,6 +337,13 @@ func checkLinux(spec rspec.Spec, rootfs string, hostCheck bool) (msgs []string) 
 			} else if spec.Linux.Namespaces[index].Type == rspec.MountNamespace {
 				mountExists = true
 			}
+		} else if hostCheck {
+			_, err := os.Stat(spec.Linux.Namespaces[index].Path)
+			if os.IsNotExist(err) {
+				msgs = append(msgs, fmt.Sprintf("Path of %v not exist", spec.Linux.Namespaces[index].Type))
+			} else if err != nil {
+				msgs = append(msgs, fmt.Sprintf("%v is invalid: %s", spec.Linux.Namespaces[index].Type, err.Error()))
+			}
 		}
 	}
 


cmd/ocitools/validate.go
```diff
@@ -337,6 +337,17 @@ func checkLinux(spec rspec.Spec, rootfs string, hostCheck bool) (msgs []string)
 			} else if spec.Linux.Namespaces[index].Type == rspec.MountNamespace {
 				mountExists = true
 			}
+		} else {
+			if hostCheck {
```
wking: `} else if hostCheck {` to avoid a layer of nesting?

Mashimiao: On 08/17/2016 11:42 AM, W. Trevor King wrote:
> |} else if hostCheck {| to avoid a layer of nesting?
I ever thought about this.
But, it seems they should not in the seem level.
So, I did not write like this.

cmd/ocitools/validate.go
```diff
@@ -337,6 +337,17 @@ func checkLinux(spec rspec.Spec, rootfs string, hostCheck bool) (msgs []string)
...
+				_, err := os.Stat(spec.Linux.Namespaces[index].Path)
+				if err != nil {
+					if os.IsNotExist(err) {
+						msgs = append(msgs, fmt.Sprintf("Path of %v not exist", spec.Linux.Namespaces[index].Type))
+					} else {
```
wking: To avoid a layer of nesting:

    _, err := os.Stat(…)
    if os.IsNotExist(err) {
      …
    } else if err != nil {
      …
    }

cmd/ocitools/validate.go
wking: On Tue, Aug 16, 2016 at 08:43:42PM -0700, Ma Shimiao wrote:
> 08/17/2016 11:42 AM, W. Trevor King:
> > |} else if hostCheck {| to avoid a layer of nesting?
> I ever thought about this.
> But, it seems they should not in the seem level.

It doesn't really matter, so if it feels too strange go ahead and
leave it in there.  It just reads more clearly to me with fewer
levels.

cmd/ocitools/validate.go
Mashimiao: On 08/17/2016 11:46 AM, W. Trevor King wrote:
> To avoid a layer of nesting:
> 
> |_, err := os.Stat(…) if os.IsNotExist(err) { … } else if err != nil { … } |

fixed, thanks.

Mashimiao:
ping @mrunalp , @liangchenye

initlove:
LGTM

wking:
c8b8c73f looks good to me.

PR #194, opened 2016-08-17
Respect the host platform

wking:
Details in the commit messages, but the net effect is to respect
runtime.GOOS and runtime.GOARCH when --host-specific is set (for both
generate and validate).  Also add an argument to generate.New to set
the target OS (falling back to runtime.GOOS).

### DIFF
--- a/cmd/oci-runtime-tool/generate.go
+++ b/cmd/oci-runtime-tool/generate.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"runtime"
 	"strconv"
 	"strings"
 	"unicode"
@@ -90,6 +91,7 @@ var generateFlags = []cli.Flag{
 	cli.StringSliceFlag{Name: "mounts-add", Usage: "configures additional mounts inside container"},
 	cli.StringSliceFlag{Name: "mounts-remove", Usage: "remove destination mountpoints from inside container"},
 	cli.BoolFlag{Name: "mounts-remove-all", Usage: "remove all mounts inside container"},
+	cli.StringFlag{Name: "os", Value: runtime.GOOS, Usage: "operating system the container is created for"},
 	cli.StringFlag{Name: "output", Usage: "output file (defaults to stdout)"},
 	cli.BoolFlag{Name: "privileged", Usage: "enable privileged container settings"},
 	cli.StringSliceFlag{Name: "process-cap-add-ambient", Usage: "add Linux ambient capabilities"},
@@ -141,21 +143,23 @@ var generateCommand = cli.Command{
 	Before: before,
 	Action: func(context *cli.Context) error {
 		// Start from the default template.
-		specgen := generate.New()
+		specgen, err := generate.New(context.String("os"))
+		if err != nil {
+			return err
+		}
 
 		var template string
 		if context.IsSet("template") {
 			template = context.String("template")
 		}
 		if template != "" {
-			var err error
 			specgen, err = generate.NewFromFile(template)
 			if err != nil {
 				return err
 			}
 		}
 
-		err := setupSpec(&specgen, context)
+		err = setupSpec(&specgen, context)
 		if err != nil {
 			return err
 		}
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -36,9 +36,14 @@ type ExportOptions struct {
 	Seccomp bool // seccomp toggles if only seccomp should be exported
 }
 
-// New creates a spec Generator with the default spec.
-func New() Generator {
-	spec := rspec.Spec{
+// New creates a spec Generator with the default spec for the target
+// OS.
+func New(os string) (generator Generator, err error) {
+	if os != "linux" && os != "solaris" {
+		return generator, fmt.Errorf("no defaults configured for %s", os)
+	}
+
+	config := rspec.Spec{
 		Version: rspec.Version,
 		Root: &rspec.Root{
 			Path:     "rootfs",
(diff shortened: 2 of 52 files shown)


generate/generate.go
```diff
@@ -39,41 +46,45 @@ func New() Generator {
...
-					Soft: uint64(1024),
-				},
-			},
 		},
 		Hostname: "mrsdalloway",
-		Mounts: []rspec.Mount{
+	}
+
+	if *os == "linux" {
+		spec.Process.User = rspec.User{}
```
Mashimiao: User is not only used for Linux, it's also used by solaris.

wking: On Wed, Aug 17, 2016 at 08:38:58AM -0700, Ma Shimiao wrote:
> >  		},
> >  		Hostname: "mrsdalloway",
> > -		Mounts: []rspec.Mount{
> > +	}
> > +
> > +	if *os == "linux" {
> > +		spec.Process.User = rspec.User{}
> 
> User is not only used for Linux, it's also used by solaris.

We could handle that in two ways.  If there is significant overlap
between some Linux and Solaris defaults, we'd want:

  if *os == "linux" || *os == "solaris" {
    spec.Process.User = rspec.User{}
    …other common setup…
  }

  if *os == "linux" {
    spec.Process.Capabilities = …
    …other Linux-specific setup…
  } else if *os == "solaris" {
    spec.Solaris.Milestone = "svc:/milestone/container:default"
    …other Solaris-specific setup…
  } else {
    return generator, fmt.Errorf("no defaults configured for %s", *os)
  }

If there is not significant overlap, it's probably easier to have:

  if *os == "linux" {
    spec.Process.User = rspec.User{}
    spec.Process.Capabilities = …
    …other Linux-specific setup…
  } else if *os == "solaris" {
    spec.Process.User = rspec.User{}
    spec.Solaris.Milestone = "svc:/milestone/container:default"
    …other Solaris-specific setup…
  } else {
    return generator, fmt.Errorf("no defaults configured for %s", *os)
  }

I think it will be easier to tell how much overlap there is once we
have a more detailed Solaris default.  I don't think setting up an
empty struct for spec.Process.User is enough to be worth claiming
support for Solaris, though, so I'd rather leave this commit as it
stands until we do have a detailed Solaris default.

mrunalp: I think we can go with second approach for now. It is easier to follow the code if there is one path we walk down per platform.

wking:
@cyphar had some concerns about this approach [1], but I'm not clear
on the details.  Just pointing that out in case we want to wait and
hear what the concerns are before thinking about merging.

[1]: https://groups.google.com/a/opencontainers.org/d/msg/dev/xxazermmxKk/eEY3Wo4IDAAJ
     Subject: Re: oci-image-tool create-runtime-bundle and Kernel Namespaces
     Date: Wed, 17 Aug 2016 15:25:26 +1000
     Message-ID: <CAOviyaienNi7B7zdv3A-3Ys4AFGgf4RHXjFhARPmYAOiV-H1eQ@mail.gmail.com>

cyphar:
Right. So my concerns are basically that I don't see why we're setting _any_ OS defaults with a plain `ocitool generate` invocation. I understand the argument (saner defaults) but the point of `ocitool` is that users have to configure things (and we generate a JSON blob that does what they asked for). I'd be fine with a `--platform=<>` flag which adds platform-specific defaults to the blob (you could even make it `--platforms=,linux,...` which added platform-specific defaults for multiple platforms in the same blob). But in general, if someone wanted to create a config that is platform-independent (so they could pass it to other tools to add platform-dependant stuff) then they wouldn't have a nice way of doing it. Also, if you're generating configs for a different platform then things get a bit odd (and I don't see why we need that).

wking:
On Thu, Aug 18, 2016 at 05:29:39PM -0700, Aleksa Sarai wrote:
> So my concerns are basically that I don't see why we're setting
> _any_ OS defaults with a plain `ocitool generate` invocation. I
> understand the argument (saner defaults) but the point of `ocitool`
> is that users have to configure things (and we generate a JSON blob
> that does what they asked for).

Callers that want to configure everything explicitly can use:

to bypass most defaults.  Without template, you get the fuller set of
ocitools-maintainer suggestions.

> I'd be fine with a `--platform=<>` flag which adds platform-specific
> defaults to the blob (you could even make it
> `--platforms=,linux,...` which added platform-specific defaults for
> multiple platforms in the same blob).

I'm ok with this, although I'd call it --defaults and have it take a
JSON path (e.g. ‘--defaults platform,linux.resources,…’).  But it
would be a fairly large shift from what we have now.

> But in general, if someone wanted to create a config that is
> platform-independent…

Do such configs exist?  There aren't many platform independent
settings, and some of those (e.g. mounts) have a platform-independent
schema but not platform independent values.  What would you put in a
platform-independent config?

> Also, if you're generating configs for a different platform then
> things get a bit odd (and I don't see why we need that).

One reason would be “I'm building an ocitools test suite and want to
make sure Windows default generation works even though I'll be running
the suite on Linux” ;).

cyphar:
> > Also, if you're generating configs for a different platform then things get a bit odd (and I don't see why we need that).
> One reason would be “I'm building an ocitools test suite and want to
make sure Windows default generation works even though I'll be running
the suite on Linux” ;).

That wording was incorrect. What I meant was that we **want** to be able to do that (create a Linux config on Solaris or whatever) but I'm not sure about the platform option being "current OS you're running on" by default -- how is that meant to interact with things like LX-branded zones (where you're running a Linux binary on illumos).

wking:
On Thu, Aug 18, 2016 at 10:32:55PM -0700, Aleksa Sarai wrote:
> What I meant was that we **want** to be able to do that (create a
> Linux config on Solaris or whatever) but I'm not sure about the
> platform option being "current OS you're running on" by default --
> how is that meant to interact with things like LX-branded zones
> (where you're running a Linux binary on illumos).

With this PR you can do that by setting --os, although we currently
only have defaults for ‘linux’.  runtime.GOOS is just the default if
--os is not set.

mrunalp:
I like the idea of defaulting to current OS or specifying alternative via --os. It covers both aspects of easy to use for default case as well as supporting cross platform spec generation.

Mashimiao:
@wking   any updates for this PR？

wking:
On Mon, Nov 14, 2016 at 12:38:38AM -0800, Ma Shimiao wrote:
> @wking   any updates for this PR？

I'm not sure what you were looking for, but I've rebased this onto
master with 217c769 → 1ae6388.

mrunalp:
Missing a signed-off?

wking:
On Mon, Nov 14, 2016 at 11:26:44AM -0800, Mrunal Patel wrote:
> Missing a signed-off?

My three commits are signed-off.  I think PullApprove is confused by
my re-basing from v1.0.0.rc1 to master.

Mashimiao:
> I like the idea of defaulting to current OS or specifying alternative via --os. It covers both aspects of easy to use for default case as well as supporting cross platform spec generation.

As @mrunalp said, I also want to keep default current OS or specifying alternative via --os 
What I'm looking for is `ea326f04c391035e653c93e10739971a6aeb5b1c` and `e3df0d57d3ead77a9abec382672f74960cdc7c73`

wking:
On Mon, Nov 14, 2016 at 05:21:26PM -0800, Ma Shimiao wrote:
> > I like the idea of defaulting to current OS or specifying
> > alternative via --os. It covers both aspects of easy to use for
> > default case as well as supporting cross platform spec generation.
>
> As @mrunalp said, I also want to keep default current OS or
> specifying alternative via --os What I'm looking for is
> `ea326f04c391035e653c93e10739971a6aeb5b1c` and
> `e3df0d57d3ead77a9abec382672f74960cdc7c73`

I'm not convinced that's what @mrunalp was saying, but if there's a
consensus around ea326f04 and e3df0d57 I'm happy to have those landed.
I've punted the tip commit off of this PR (into #270), so we can get
this one landed sooner.

validate/validate.go
```diff
@@ -164,6 +165,16 @@ func (v *Validator) CheckPlatform() (msgs []string) {
 	}
 	msgs = append(msgs, fmt.Sprintf("Operation system %q of the bundle is not supported yet.", platform.OS))
 
+	if hostCheck {
```
Mashimiao: hostCheck->v.HostSpecific

cmd/oci-runtime-tool/generate.go
```diff
@@ -92,21 +92,23 @@ var generateCommand = cli.Command{
 	Before: before,
 	Action: func(context *cli.Context) error {
 		// Start from the default template.
-		specgen := generate.New()
+		specgen, err := generate.New(nil)
```
Mashimiao: How about getting `--os` value before `generate.New()`?
Then we don't have to pass nil any more.

wking:
On Mon, Nov 14, 2016 at 10:43:03PM -0800, Ma Shimiao wrote:
> +	if hostCheck {
> 
> hostCheck->v.HostSpecific

Oops, fixed in ea326f0 → 2cb6c62.

> -		specgen := generate.New()
> +		specgen, err := generate.New(nil)
> 
> How about getting `--os` value before `generate.New()`?
> Then we don't have to pass nil any more.

I liked it better when the caller didn't have to have an opinion (what
happens if/when we extend platform.os or platform.arch beyond
GOOS/GOARCH?).  But whatever, also updated in ea326f0 → 2cb6c62.

Mashimiao:
@opencontainers/runtime-tools-maintainers , @wking 
I still have a question, which makes me a little confused. I think we need to make it clear.
If we used a template and also used some options and the values of options may conflict with values in template, should we respect values in template or let values of options to cover them?

wking:
On Tue, Nov 15, 2016 at 05:09:28PM -0800, Ma Shimiao wrote:
> If we used a template and also used some options and the values of
> options may conflict with values in template, should we respect
> values in template or let values of options to cover them?

Explicit command line options (e.g. ‘--os windows’) should always take
precedence.  The template *or* default config comes through when
conflicting command line options are not set.

As it stands in this PR, the template is getting clobbered by the
*implicit* default for the --os option.  That's what the tip commit
(now in #270) is fixing.  With the commits remaining in this PR, we
start respecting explicit --os to select non-conflicting defaults for
settings besides platform.os (e.g. not filling in ‘linux’ when asked
to generate a new config with ‘--os windows’).

generate/generate.go
```diff
@@ -143,12 +149,15 @@ func New() Generator {
 				},
 			},
 			Devices: []rspec.Device{},
-		},
+		}
+	} else {
+		return generator, fmt.Errorf("no defaults configured for %s", os)
```
q384566678: I think you need to generate the appropriate configuration for `windows` and `solaris` .

wking:
I see this is tagged v0.6.0, so I've rebased it onto master with 2cb6c62 → 984db97.  Hopefully we can land it before too many more `validation/*.go` land, since they'll probably need rebasing now that `generate.New(…)` can return an error.

generate/generate.go
wking: > I think you need to generate the appropriate configuration for `windows` and `solaris`.

I'm creating a very slim Solaris default with 984db97c.  I'm happy to add a Window default too if you'll tell me what it should be, but I don't know enough about Windows to be able to figure that out on my own.  Since we don't currently support generating defaults for Windows configs, I don't see a problem with leaving that unimplemented for this PR and just adjusting the function signatures to allow it to be added in the future.

wking:
I've pushed 984db97 → 4aecad2, restoring the `generate` `--os` option which had been removed by 597c7d44 (#409).  Otherwise there's no way to generate default configs for other platforms.

q384566678:
LGTM

PR #197, opened 2016-08-18
validate: Check configuration against JSON Schema

wking:
runtime-spec publishes [a JSON Schema covering the configuration format (and other JSON related to runtime-spec)](https://github.com/opencontainers/runtime-spec/tree/v1.0.0-rc1/schema).  Reduce duplication of effort by validating configurations against that schema.  For example this gives us lots of range checking:

```
$ ocitools validate
...
1. linux.resources.oomScoreAdj: Must be greater than or equal to -1000
...
```

without us having to duplicate all the range-input work that the runtime-spec folks have already done for us.

Only validating the JSON Schema is not sufficient, because `--host-specific` (e.g. you're running on a Linux box) and cross-property constraits (e.g. must create a new UTS namespace if you set hostname) are difficult/impossible to express in JSON Schema.

See related discussion in #195.


Mashimiao:
It seems you need to add `gojsonschema` to godeps

wking:
On Wed, Aug 17, 2016 at 11:43:25PM -0700, Ma Shimiao wrote:
> It seems you need to add `gojsonschema` to godeps

I think I've figured that out now, although it took a few tries ;).  I
don't write much Go :p.

Mashimiao:
I'm not sure what's wrong with your PR. I can't find the changed `validate.go` in https://github.com/opencontainers/ocitools/pull/197/files

wking:
On Wed, Aug 17, 2016 at 11:57:43PM -0700, Ma Shimiao wrote:
> I'm not sure what's wrong with your PR. I can't find the changed
> `validate.go` in
> https://github.com/opencontainers/ocitools/pull/197/files

That has:

  Sorry, we could not display the entire diff because too many files
  (392) changed.

I'd suggest trusting godep to have done the autogenerated commit
correctly, and then just reviewing [1].

[1]: https://github.com/opencontainers/ocitools/pull/197/commits/d4b604c3970cf5157e3905ecfcb3f305d2603c50

Mashimiao:
It looks good to me. 
Except, `schema.json` has been changed to `config-shema.json` in rumtime's upstream. It seems we have to change URL based on spec version in the later.

wking:
On Thu, Aug 18, 2016 at 12:18:41AM -0700, Ma Shimiao wrote:
> Except, `schema.json` has been changed to `config-shema.json` in
> rumtime's upstream. It seems we have to change URL based on spec
> version in the later.

Yeah, that's the sort of thing we'll switch on in JSONSchemaURL.

cmd/ocitools/validate.go
```diff
@@ -61,6 +62,8 @@ var (
 		"CAP_KILL",
 		"CAP_AUDIT_WRITE",
 	}
+
+	configSchemaTemplate = "https://raw.githubusercontent.com/opencontainers/runtime-spec/v%s/schema/schema.json"
```
mrunalp: We should vendor this file rather than downloading it.

wking: On Thu, Aug 18, 2016 at 11:41:17AM -0700, Mrunal Patel wrote:
> > @@ -61,6 +62,8 @@ var (
> >  		"CAP_KILL",
> >  		"CAP_AUDIT_WRITE",
> >  	}
> > +
> > +	configSchemaTemplate = "https://raw.githubusercontent.com/opencontainers/runtime-spec/v%s/schema/schema.json"
> 
> We should vendor this file rather than downloading it. 

If we vendor (and I don't see a huge need for that), I'd rather vendor
with the option to download (opencontainers/image-spec#150) ;).  The
drawback to *only* vendoring is that if we don't vendor (and
runtime-spec is consistent about the naming pattern) we can
automatically validate against *future* schema versions that weren't
available at compile time.

I am a fan of not requiring network activity, but I'd rather support
that with a ‘--schema $URL’ option (something like
opencontainers/runtime-spec#490) so users could run:

  $ oci validate --schema file:///path/to/config-schema.json …

wking:
I just pushed 9564b11 → bf09bc2 bumping the commit timestamp to get a
new hash, since GitHub seems to be confused by 9564b11 (never ran
Travis and think's @mrunalp is commenting on an outdated diff [1]).

[1]: https://github.com/opencontainers/ocitools/pull/197#discussion_r75365051

cmd/ocitools/validate.go
mrunalp: If we are compiling ocitools for a particular runtime-spec version, it makes sense to use the version from the same schema by default.

wking: On Thu, Sep 01, 2016 at 03:37:26PM -0700, Mrunal Patel wrote:
> > @@ -61,6 +62,8 @@ var (
> >  		"CAP_KILL",
> >  		"CAP_AUDIT_WRITE",
> >  	}
> > +
> > +	configSchemaTemplate = "https://raw.githubusercontent.com/opencontainers/runtime-spec/v%s/schema/schema.json"
> 
> If we are compiling ocitools for a particular runtime-spec version…

I'd rather *not* compile ocitools for a particular runtime-spec
version.  By staying flexible here we have one less place where we are
pinning ourselves down.

initlove:
+1 for this, I'll remove duplicated checks like 'mandatory',  or specific enums ('SeccompArch") after this.

can we have less files in vendor? For example 'testsuite' is not necessary.

wking:
On Sun, Sep 11, 2016 at 06:45:56AM -0700, David Liang wrote:
> can we have less files in vendor? For example 'testsuite' is not
> necessary.

I've put my godep invocation in the bf09bc24 commit message.  If I
should be running something else to update godeps, let me know.

wking:
Review here seems to have stalled out.  I'm happy to rebase this when folks have time to take another look; just ping me here.

Mashimiao:
any more comments?   @opencontainers/runtime-tools-maintainers

Mashimiao:
Please contribute to master branch.

wking:
> Please contribute to master branch.

I can rebase onto master and change the target branch without filing a new PR and fragmenting discussion.  Can you re-open this PR so I can do that?

wking:
Redirected the PR and rebased onto master with bf09bc2 → afd025f.

Mashimiao:
@wking   Can we use the schema file locally something as image-tools have done? I still don't think downloading is a good way.

wking:
On Mon, Jul 24, 2017 at 02:16:05AM +0000, Ma Shimiao wrote:
> Can we use the schema file locally something as image-tools have
> done?

Maintaining fs.go is a source of useless merge conflicts (e.g. see
@vbatts complaining about that in [1]).

> I still don't think downloading is a good way.

Previous discussion on that point starting with [2].  What are you
concerned about?  Bandwidth?  Network connectivity?  If you don't like
the explicit-local-directory approach [3], would you accept a local
cache (e.g.
$XDG_CACHE_HOME/oci-runtime-tools/runtime-spec/v1.0.0/config-schema.json
[4])?  Then you'd only need network connectivity for the first load
for a given version.

Also note that the JSON Schema we want to use for a v1.0.0 config is
probably going to be a v1.0.x patch release to pick up fixes like
opencontainers/runtime-spec#889.

[1]: https://github.com/opencontainers/image-spec/pull/544#issuecomment-279423767
[2]: https://github.com/opencontainers/runtime-tools/pull/197#r75365051
[3]: https://github.com/opencontainers/runtime-tools/pull/197#r75368764
[4]: https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html

wking:
I've pushed afd025f → cfb2f1c to shift the vendored content from `Godeps/_workspace/src` to `vendor`, although I'm not sure why my `godep` didn't put them there on its own.

Mashimiao:
I don't think it's a good idea that whether you can validate bundle depends on if your network is OK or not.

Mashimiao:
But if do as image-spec is not a good way either, I can't think out any better ideas. Maybe we can try to let this in first? @opencontainers/runtime-tools-maintainers

wking:
On Fri, Jul 28, 2017 at 03:33:14AM +0000, Ma Shimiao wrote:
> I don't think it's a good idea that whether you can validate bundle
> depends on if your network is OK or not.

If you use any sort of local caching or --schema override, then
whether you can validate only depends on whether your network was
working when you decided to populate that cache.  And we can provide a
nice tar of all the relevant JSON schemas (under opencontainers.org?)
if we want to make that easy for package managers and such.

liangchenye:
@q384566678    wking already works on this year ago +1: 
we can review and work on this.

Mashimiao:
@liangchenye  and I plan to focus on this. @wking  please rebase and update.

Mashimiao:
ping @wking

wking:
Rebased and added some tests with cfb2f1c → 17713b5.

validate/validate.go
```diff
@@ -114,6 +118,49 @@ func (v *Validator) CheckAll() error {
...
+// JSONSchemaURL returns the URL for the JSON Schema specifying the
+// configuration format.  It consumes configSchemaTemplate, but we
+// provide it as a function to isolate consumers from inconsistent
+// naming as runtime-spec evolves.
+func JSONSchemaURL(version string) (url string, err error) {
+	ver, err := semver.Parse(version)
+	if err != nil {
+		return "", err
+	}
+	configSchemaVersion, err := semver.Parse("1.0.0-rc2") // config.json became config-schema.json in 1.0.0-rc2
```
Mashimiao: should be 1.0.0?

Mashimiao:
And CI failed

validate/validate.go
wking: > should be 1.0.0?

No, rc2 is when `config.jdon` became `config-schema.json`.  `JSONSchemaURL` works with all of our releases, not just `1.0.0`.

wking:
> And CI failed

Looks like I forgot to format the Go.  Will do later today.

wking:
On Sat, Sep 09, 2017 at 01:52:07PM +0000, W. Trevor King wrote:
> Looks like I forgot to format the Go.

Fixed the formatting with 17713b5 → ea82fad.

validate/validate.go
Mashimiao: Sorry, I can't understand.
Our code now works based on runtime-spec v1.0.0.
Even though schema changes begin rc2,  why must be rc2?  It's not the latest released version.

wking: > Even though schema changes begin rc2, why must be rc2? It's not the latest released version.

I think the long-term goal is to be able to test against multiple spec versions.  See #98 for more on this.  @liangchenye at least seemed on board with that idea (if not with my Python approach to it) based on his comment [here][1].  I've pushed ea82fad → 9d07098 showing a example of using `CheckJSONSchema` on an older release (although I didn't go back before rc2) in case that helps clarify.  I understand that the rest of the Go-based validation is specific to 1.0.0, but don't think that is a reason to force `CheckJSONSchema` to be specific to 1.0.0.  Having a flexible `CheckJSONSchema` means we'll have less work to do when we eventually adjust `validate` to work with multiple versions.

The flexibility of `CheckJSONSchema` is somewhat limited in the current implementation because it happens after we read JSON into Go (and in some `validate` workflows there may never have been JSON to begin with).  That means we miss things like:

```json
{
  "ociVersion": "1.0.0-rc6",
  "linux": {
    "resources": {
      "disableOOMKiller": "I should be a boolean"
    }
  }
}
```

because since opencontainers/runtime-spec#896 there has been no way to represent `linux.resources.disableOOMKiller` in Go.  But I expect folks will mostly be interested in 1.0.0+, and the SemVer commitment will help us avoid that sort of issue for the 1.x line.

[1]: https://github.com/opencontainers/runtime-tools/issues/98#issuecomment-223878770

Mashimiao: Ah,  sorry.  I misunderstand the purpose of configSchemaVersion.

wking: > I misunderstand the purpose of configSchemaVersion.

Maybe I should rename it to `configRenamedToConfigSchemaVersion` or some such?

Mashimiao: I think it will be better

wking: > I think it will be better

Renamed in 9d07098 → 3935592.

Mashimiao:
LGTM

Mashimiao:
@liangchenye  @q384566678  @mrunalp  PTAL

q384566678:
LGTM

PR #199, opened 2016-08-23
add namespace check for uid/gid mappings

Mashimiao:

### DIFF
--- a/cmd/ocitools/validate.go
+++ b/cmd/ocitools/validate.go
@@ -316,13 +316,7 @@ func checkLinux(spec rspec.Spec, rootfs string, hostCheck bool) (msgs []string) 
 	ipcExists := false
 	mountExists := false
 	netExists := false
-
-	if len(spec.Linux.UIDMappings) > 5 {
-		msgs = append(msgs, "Only 5 UID mappings are allowed (linux kernel restriction).")
-	}
-	if len(spec.Linux.GIDMappings) > 5 {
-		msgs = append(msgs, "Only 5 GID mappings are allowed (linux kernel restriction).")
-	}
+	userExists := false
 
 	for index := 0; index < len(spec.Linux.Namespaces); index++ {
 		if !namespaceValid(spec.Linux.Namespaces[index]) {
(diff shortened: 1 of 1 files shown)


cmd/ocitools/validate.go
```diff
@@ -340,6 +336,14 @@ func checkLinux(spec rspec.Spec, rootfs string, hostCheck bool) (msgs []string)
 		}
 	}
 
+	if (len(spec.Linux.UIDMappings) > 0 || len(spec.Linux.GIDMappings) > 0) && !userExists {
+		msgs = append(msgs, "UID/GID mappings require User namespace exists")
```
wking: The phrasing used by existing similar warnings is:

> {setting} requires a new {type} namespace to be specified as well

Mashimiao: On 08/24/2016 12:50 AM, W. Trevor King wrote:
> The phrasing used by existing similar warnings is:
> 
>     {setting} requires a new {type} namespace to be specified as well
> 
I think container joins into an existing user namespace also can work.
Doesn't it?

wking: On Tue, Aug 23, 2016 at 05:58:31PM -0700, Ma Shimiao wrote:
> > +	if (len(spec.Linux.UIDMappings) > 0 || len(spec.Linux.GIDMappings) > 0) && !userExists {
> > +		msgs = append(msgs, "UID/GID mappings require User namespace exists")
> 
> 08/24/2016 12:50 AM, W. Trevor King:
> > The phrasing used by existing similar warnings is:
> > 
> >     {setting} requires a new {type} namespace to be specified as well
>
> I think container joins into an existing user namespace also can
> work.  Doesn't it?

The kernel has no problem with join-and-tweak, but the OCI spec does
not allow it at the moment [1,2].  Interestingly, the spec places no
such restriction on namespaces inherited from the host.  I've filed
[3] to figure out what the runtime-spec maintainers want to do about
that.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc1/config-linux.md#L41
[2]: https://github.com/opencontainers/runtime-spec/pull/158
[3]: https://github.com/opencontainers/runtime-spec/issues/537

Mashimiao: Thanks, got it. 
Since spec does not allow now, I fixed error message as you suggested.

wking:
bf50b73b looks good to me.  It's taking the (b) approach from [1],
even though the current spec wording doesn't quite match that
interpretation.  But it matches the ocitools existing namespace
validation logic, and we can revisit all of that if [1] ends up being
resolved in a different direction.

[1]: https://github.com/opencontainers/runtime-spec/issues/537

mrunalp:
LGTM

PR #202, opened 2016-08-31
man/ocitools.1: Replace "**...(1)**" with "**...**(1)"

wking:
Before this commit the latter was slightly more popular:

    $ git grep '\*\*(1)' | wc -l
    3
    $ git grep '(1)\*\*' | wc -l
    2

Change generated with:

    $ sed -i 's/(1)\*\*/**(1)/g' $(git grep -l '(1)\*\*')


mrunalp:
LGTM

PR #209, opened 2016-09-12
runtime: support IDMappings validation

Mashimiao:

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -409,6 +409,79 @@ func validateOOMScoreAdj(spec *rspec.Spec) error {
 	return nil
 }
 
+func getIDMappings(path string) ([]rspec.IDMapping, error) {
+	var idMaps []rspec.IDMapping
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+	s := bufio.NewScanner(f)
+	for s.Scan() {
+		if err := s.Err(); err != nil {
+			return nil, err
+		}
+
+		idMap := strings.Fields(strings.TrimSpace(s.Text()))
+		if len(idMap) == 3 {
+			hostID, err := strconv.ParseUint(idMap[0], 0, 32)
+			if err != nil {
+				return nil, err
+			}
+			containerID, err := strconv.ParseUint(idMap[1], 0, 32)
+			if err != nil {
+				return nil, err
+			}
+			mapSize, err := strconv.ParseUint(idMap[2], 0, 32)
+			if err != nil {
+				return nil, err
+			}
+			idMaps = append(idMaps, rspec.IDMapping{HostID: uint32(hostID), ContainerID: uint32(containerID), Size: uint32(mapSize)})
+		} else {
+			return nil, fmt.Errorf("invalid format in %v", path)
+		}
+	}
+
+	return idMaps, nil
+}
+
+func validateIDMappings(mappings []rspec.IDMapping, path string, property string) error {
+	idMaps, err := getIDMappings(path)
+	if err != nil {
+		return fmt.Errorf("can not get items: %v", err)
+	}
+	if len(mappings) != 0 && len(mappings) != len(idMaps) {
+		return fmt.Errorf("expected %d entries in %v, but acutal is %d", len(mappings), path, len(idMaps))
+	}
+	for _, v := range mappings {
+		exist := false
+		for _, cv := range idMaps {
+			if v.HostID == cv.HostID && v.ContainerID == cv.ContainerID && v.Size == cv.Size {
+				exist = true
+				break
+			}
+		}
+		if !exist {
+			return fmt.Errorf("%v is not applied as expected", property)
+		}
+	}
+
+	return nil
+}
+
+func validateUIDMappings(spec *rspec.Spec) error {
+	logrus.Debugf("validating uidMappings")
+
+	return validateIDMappings(spec.Linux.UIDMappings, "/proc/self/uid_map", "linux.uidMappings")
+}
+
+func validateGIDMappings(spec *rspec.Spec) error {
+	logrus.Debugf("validating gidMappings")
+
+	return validateIDMappings(spec.Linux.GIDMappings, "/proc/self/gid_map", "linux.gidMappings")
+}
+
 func mountMatch(specMount rspec.Mount, sysMount rspec.Mount) error {
 	if specMount.Destination != sysMount.Destination {
 		return fmt.Errorf("mount destination expected: %v, actual: %v", specMount.Destination, sysMount.Destination)
@@ -489,6 +562,8 @@ func validate(context *cli.Context) error {
 		validateMaskedPaths,
 		validateROPaths,
 		validateOOMScoreAdj,
+		validateUIDMappings,
+		validateGIDMappings,
 	}
 
 	for _, v := range defaultValidations {


cmd/runtimetest/main.go
```diff
@@ -320,6 +322,95 @@ func validateROPaths(spec *rspec.Spec) error {
...
 }
 
+func getIDMappings(idType string) ([]rspec.IDMapping, error) {
+	var mapFile string
+	var idMaps []rspec.IDMapping
+	switch idType {
+	case "uid_map":
+		mapFile = "/proc/1/uid_map"
+	case "gid_map":
+		mapFile = "/proc/1/gid_map"
```
wking: Can we look at `/proc/self/*_map` instead?  Then we don't have to think about “is it ok to look at PID 1's maps?”.  I can't think of a case where looking at PID 1 would be wrong, but I know looking at `/proc/self/*_map` will be right ;).

cmd/runtimetest/main.go
```diff
@@ -320,6 +322,95 @@ func validateROPaths(spec *rspec.Spec) error {
...
+	defer f.Close()
+
+	s := bufio.NewScanner(f)
+	for s.Scan() {
+		if err := s.Err(); err != nil {
+			return nil, err
+		}
+
+		text := strings.Replace(strings.TrimSpace(s.Text()), "          ", " ", -1)
+		idMap := strings.Split(text, " ")
```
wking: Better to use [`Fields`][1] here, which will let you drop the `Replace` line.

[1]: https://golang.org/pkg/strings/#Fields

cmd/runtimetest/main.go
wking: And I'd bump this choice up to the caller and have `getIDMappings` take a path:

    func getIDMappings(path string) ([]rspec.IDMapping, error)

cmd/runtimetest/main.go
```diff
@@ -320,6 +322,95 @@ func validateROPaths(spec *rspec.Spec) error {
...
+	}
+
+	return idMaps, nil
+}
+
+func validateUIDMappings(spec *rspec.Spec) error {
+	logrus.Debugf("validating uidMappings")
+	idMaps, err := getIDMappings("uid_map")
+	if err != nil {
+		return fmt.Errorf("can not get items from uid_map: %v", err)
```
wking: image-spec is using `errors.Wrapf` for this sort of thing, and I expect we probably should use that here too.  But this can happen in a follow-up PR if you like.

cmd/runtimetest/main.go
```diff
@@ -320,6 +322,95 @@ func validateROPaths(spec *rspec.Spec) error {
...
+		} else {
+			return nil, fmt.Errorf("invalid format in %v", mapFile)
+		}
+	}
+
+	return idMaps, nil
+}
+
+func validateUIDMappings(spec *rspec.Spec) error {
+	logrus.Debugf("validating uidMappings")
```
wking: I think we need a guard on this for when `spec.Linux.UIDMappings` is empty.  When you're creating a new user namespace, you expect `uid_map` to be empty if `UIDMappings` is empty.  When you're *not* creating a new user namespace, you expect the host's `uid_map` to be unchanged if `UIDMappings` is empty and the [config is invalid][1] if `UIDMappings` is not empty.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc1/config-linux.md#L41

cmd/runtimetest/main.go
```diff
@@ -320,6 +322,95 @@ func validateROPaths(spec *rspec.Spec) error {
...
+	idMaps, err := getIDMappings("uid_map")
+	if err != nil {
+		return fmt.Errorf("can not get items from uid_map: %v", err)
+	}
+	for _, v := range spec.Linux.UIDMappings {
```
wking: This ensures all `UIDMappings` entries are in `uid_map`, but does not check if all `uid_map` entries are in `UIDMappings`.  One way to do that would be to follow this loop with something like:

    if len(spec.Linux.UIDMappings) != len(idMaps) {
        return fmt.Errorf("unrecognized entries in /proc/self/uid_map: %d > %d", len(idMaps), len(spec.Linux.UIDMappings))
    }

but you'll get more informative error messages if you use a loop like your existing one for `idMaps`:
    
    for _, v := range spec.Linux.UIDMappings {
      …v in idMaps?…
    }
    for _, v := range idMaps {
      …v in spec.Linux.UIDMappings?…
    }

cmd/runtimetest/main.go
Mashimiao: fixed, thanks

cmd/runtimetest/main.go
Mashimiao: fixed, thanks.

cmd/runtimetest/main.go
Mashimiao: I think fix this in a follow-up PR is more suitable.

cmd/runtimetest/main.go
```diff
@@ -320,6 +322,95 @@ func validateROPaths(spec *rspec.Spec) error {
...
+		}
+		if !exist {
+			return fmt.Errorf("uidMappings is not applied as expected")
+		}
+	}
+
+	return nil
+}
+
+func validateGIDMappings(spec *rspec.Spec) error {
```
wking: This is very similar to `validateUIDMappings`  Can we pull out the common code into:

    validateIDMappings(mappings []IDMapping, path string, property string) error

and then use:

    func validateUIDMappings(spec *rspec.Spec) error {
        return validateIDMappings(spec.Linux.UIDMappings, "/proc/self/uid_map", "linux.uidMappings")
    }

    func validateGIDMappings(spec *rspec.Spec) error {
        return validateIDMappings(spec.Linux.GIDMappings, "/proc/self/gid_map", "linux.gidMappings")
    }

cmd/runtimetest/main.go
Mashimiao: I think If you creating a new user namespace, UIDMappings/GIDMappings must not be empty.
Am I right?

wking: On Wed, Sep 14, 2016 at 12:01:45AM -0700, Ma Shimiao wrote:
> > +func validateUIDMappings(spec *rspec.Spec) error {
> > +	logrus.Debugf("validating uidMappings")
> 
> I think If you creating a new user namespace,
> UIDMappings/GIDMappings must not be empty.  Am I right?

I don't see anything about that in [1].  Perhaps the user intended to
setup the mappings on their own before calling ‘start’?  I agree that
folks creating a new user namespace will usually put something in
these arrays, but I'd rather not bake in that assumption here without
being able to back it up in the spec.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/config-linux.md#user-namespace-mappings

cmd/runtimetest/main.go
Mashimiao: As you said above, if user could setup the mappings on their own before calling ‘start’, I think we can't check this. But in my opinion, user can't setup such thing before calling start.

cmd/runtimetest/main.go
Mashimiao: fixed, thanks

cmd/runtimetest/main.go
wking: On Wed, Sep 14, 2016 at 12:34:47AM -0700, Ma Shimiao wrote:
> As you said above, if user could setup the mappings on their own
> before calling ‘start’, I think we can't check this. But in my
> opinion, user can't setup such thing before calling start.

I think we want to make runtimetest too strict (with a caveat
explaining that [1]) and then not do the legal things it doesn't
allow.  The constraints on the validated runtime are much weaker if
you make runtimetest too permissive.  So where hitting “just right”
isn't possible, I'd err on the side of “too strict”.

[1]: https://github.com/opencontainers/ocitools/pull/211#discussion_r78610440

cmd/runtimetest/main.go
Mashimiao: On 09/14/2016 02:32 PM, W. Trevor King wrote:
> I think we need a guard on this for when |spec.Linux.UIDMappings| is empty. When you're creating a new user namespace, you expect |uid_map| to be empty if |UIDMappings| is empty. When you're /not/ creating a new user namespace, you expect the host's |uid_map| to be unchanged if |UIDMappings| is empty and the config is invalid <https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc1/config-linux.md#L41> if |UIDMappings| is not empty.

In my opinion:
This validation mainly intends to check if idMappings are applied as expected.
If idMappings is not empty, we should check if they are all in *_map and only them.
If not, we don't need to care if a new user namespace is created, just return error.

if idMappings is empty, this is a hard case.
As, I think we don't have ability to check if host *_map is unchanged or not.

wking: On Wed, Sep 14, 2016 at 01:27:41AM -0700, Ma Shimiao wrote:
> On 09/14/2016 02:32 PM, W. Trevor King wrote:
> > I think we need a guard on this for when |spec.Linux.UIDMappings|
> > is empty. When you're creating a new user namespace, you expect
> > |uid_map| to be empty if |UIDMappings| is empty. When you're /not/
> > creating a new user namespace, you expect the host's |uid_map| to
> > be unchanged if |UIDMappings| is empty and the config is invalid
> > <https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc1/config-linux.md#L41>
> > if |UIDMappings| is not empty.
> 
> In my opinion:
> This validation mainly intends to check if idMappings are applied as
> expected.  If idMappings is not empty, we should check if they are
> all in *_map and only them.  If not, we don't need to care if a new
> user namespace is created, just return error.

All of that sounds good.  You're currently missing the “and only them”
portion of that check.

> if idMappings is empty, this is a hard case.  As, I think we don't
> have ability to check if host *_map is unchanged or not.

You could pass it in via annotations or another auxiliary file if you
want to perform the comparison inside the container.  Or you could
test that by having the container process just print out the
information you want to inspect so you can carry out the comparison on
the host side.

With the sharness tests I use for ccon, I take the latter approach.
For example, the container process can print its capabilities and
host-side tooling can check them against an expected value [1].  This
lets you keep the container process leaner, because for a lot of tests
you could just use BusyBox (e.g. to cat /proc/self/uid_map).

[1]: https://github.com/wking/ccon/blob/v0.4.0/test/t1003-process-capabilities.t#L35-L46

Mashimiao: I update PR to cover "only them" check.
As checking host *_map is unchanged or not, after read you mentioned above, it seems clearly  we can't finish it just by runtimetest itself. Can we leave this in the follow-up PRs. 
Especially after test_runtimetest.sh is rebuilt, I think many cases test can't be finished by runtimetest alone, as cgroupPath, cgroups, etc.

wking: On Wed, Sep 14, 2016 at 01:51:29AM -0700, Ma Shimiao wrote:
> As checking host *_map is unchanged or not, after read you mentioned
> above, it seems clearly we can't finish it just by runtimetest
> itself. Can we leave this in the follow-up PRs.  Especially after
> test_runtimetest.sh is rebuilt, I think many cases test can't be
> finished by runtimetest alone, as cgroupPath, cgroups, etc.

Yeah, sounds good.

Mashimiao:
@mrunalp  @liangchenye   rebased.

mrunalp:
LGTM

PR #210, opened 2016-09-12
runtime: support oomScoreAdj validation

Mashimiao:

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bufio"
 	"bytes"
 	"encoding/json"
 	"fmt"
@@ -8,6 +9,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"syscall"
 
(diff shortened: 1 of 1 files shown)


cmd/runtimetest/main.go
```diff
@@ -320,6 +322,35 @@ func validateROPaths(spec *rspec.Spec) error {
...
+			text := strings.TrimSpace(s.Text())
+			actual, err := strconv.Atoi(text)
+			if err != nil {
+				return err
+			}
+			if actual != expected {
+				return fmt.Errorf("oomScoreAdj expected: %v, actual: %v", expected, actual)
+			}
+		}
+	}
```
wking: What happens if `oomScoreAdj` is `null` or unset?  Do we require particular runtime behavior in that case?  I'd expect we require the runtime to leave the setting alone, although the current spec doesn't spell that out.

wking:
I think 3d2cac7e makes it more obvious that we should specify what
happens when settings are null or unset.  But the spec doesn't
currently do that clearly, so having untested, unspecified behavior
matches the current spec and 3d2cac7e looks good to me.

mrunalp:
LGTM

PR #211, opened 2016-09-13
runtimetest: add linux devices validation

Mashimiao:

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -271,6 +271,65 @@ func validateDefaultFS(spec *rspec.Spec) error {
 	return nil
 }
 
+func validateLinuxDevices(spec *rspec.Spec) error {
+	logrus.Debugf("validating linux devices")
+
+	for _, device := range spec.Linux.Devices {
+		fi, err := os.Stat(device.Path)
+		if err != nil {
+			return err
+		}
+		fStat, ok := fi.Sys().(*syscall.Stat_t)
+		if !ok {
+			return fmt.Errorf("cannot determine state for device %s", device.Path)
+		}
+		var devType string
+		switch fStat.Mode & syscall.S_IFMT {
+		case syscall.S_IFCHR:
+			devType = "c"
+			break
+		case syscall.S_IFBLK:
+			devType = "b"
+			break
+		case syscall.S_IFIFO:
+			devType = "p"
+			break
+		default:
+			devType = "unmatched"
+		}
+		if devType != device.Type || (devType == "c" && device.Type == "u") {
+			return fmt.Errorf("device %v expected type is %v, actual is %v", device.Path, device.Type, devType)
+		}
+		if devType != "p" {
+			dev := fStat.Rdev
+			major := (dev >> 8) & 0xfff
+			minor := (dev & 0xff) | ((dev >> 12) & 0xfff00)
+			if int64(major) != device.Major || int64(minor) != device.Minor {
+				return fmt.Errorf("%v device number expected is %v:%v, actual is %v:%v", device.Path, device.Major, device.Minor, major, minor)
+			}
+		}
+		if device.FileMode != nil {
+			expected_perm := *device.FileMode & os.ModePerm
+			actual_perm := fi.Mode() & os.ModePerm
+			if expected_perm != actual_perm {
+				return fmt.Errorf("%v filemode expected is %v, actual is %v", device.Path, expected_perm, actual_perm)
+			}
+		}
+		if device.UID != nil {
+			if *device.UID != fStat.Uid {
+				return fmt.Errorf("%v uid expected is %v, actual is %v", device.Path, *device.UID, fStat.Uid)
+			}
+		}
+		if device.GID != nil {
+			if *device.GID != fStat.Gid {
+				return fmt.Errorf("%v uid expected is %v, actual is %v", device.Path, *device.GID, fStat.Gid)
+			}
+		}
+	}
+
+	return nil
+}
+
 func validateDefaultDevices(spec *rspec.Spec) error {
 	logrus.Debugf("validating linux default devices")
 
@@ -395,6 +454,7 @@ func validate(context *cli.Context) error {
 	linuxValidations := []validation{
 		validateDefaultFS,
 		validateDefaultDevices,
+		validateLinuxDevices,
 		validateSysctls,
 		validateMaskedPaths,
 		validateROPaths,


cmd/runtimetest/main.go
```diff
@@ -394,7 +450,7 @@ func validate(context *cli.Context) error {
 
 	linuxValidations := []validation{
 		validateDefaultFS,
-		validateDefaultDevices,
```
wking: We want to keep this (as long as the spec requires runtimes to supply those devices).

cmd/runtimetest/main.go
```diff
@@ -271,6 +271,62 @@ func validateDefaultFS(spec *rspec.Spec) error {
...
 }
 
+func validateLinuxDevices(spec *rspec.Spec) error {
+	logrus.Debugf("validating linux devices")
+
+	for _, device := range spec.Linux.Devices {
+		fi, err := os.Stat(device.Path)
+		if err != nil {
+			return err
+		}
```
wking: Perhaps not technically true (e.g. you could remove the device before calling `start`), but probably true for all the configurations we'll run in our validation tests.  Do we want to document that caveat somewhere?  Maybe add:

    app.Description = "runtimetest compares its current environment with an OCI runtime configuration read from `config.json` in its current working directory.  The tests are fairly generic and cover most configurations used by the runtime validation suite, but there are corner cases where a container launched by a valid runtime would not satisfy runtimetest."

to `main`?

cmd/runtimetest/main.go
Mashimiao: On 09/14/2016 01:38 AM, W. Trevor King wrote:
> We want to keep this (as long as the spec requires runtimes to supply those devices).

validateDefaultDevices is not removed.
I have just merged it into validateLinuxDevices

cmd/runtimetest/main.go
Mashimiao: On 09/14/2016 01:52 AM, W. Trevor King wrote:
> you could remove the device before calling |start|
Sorry, I'm not sure about this.
Could you tell me by how to remove the device?

cmd/runtimetest/main.go
Mashimiao: un... I think separate them more reasonable, will update.

cmd/runtimetest/main.go
wking: On Tue, Sep 13, 2016 at 06:22:39PM -0700, Ma Shimiao wrote:
> > +	for _, device := range spec.Linux.Devices {
> > +		fi, err := os.Stat(device.Path)
> > +		if err != nil {
> > +			return err
> > +		}
> 
> On 09/14/2016 01:52 AM, W. Trevor King wrote:
> > you could remove the device before calling |start|
>
> Sorry, I'm not sure about this.  Could you tell me by how to remove
> the device?

I'm not saying that you shuold remove anything.  I'm just saying that
a valid runtime launching a valid config might not pass this test.
For example:

  {
    …
    "devices": [
      {
        "path": "/dev/fuse",
        …
      }
    ],
    "hooks": {
      "prestart": [
        {
          "path": "/bin/rm",
          "args": ["rm", "/dev/fuse"]
        }
    }
  }

That's going to lead to a container without a /dev/fuse entry, which
would fail this check.  I think that's ok because we won't be calling
runtimetest with such a config.  But I'd like to put a caveat in the
--help description to make it clear that we're not attempting to
validate all possible configs with runtimetest.

Mashimiao: Ah,  this is a possible case, though it seems is a  meaningless setting.
Do you think it's possible to detect device related operations in prestart, and based on it to change device validation?

Mashimiao: As you know, almost all validation error out, if config did not match container.  Though we added a caveat in help message, the runtimetest still could not finish whole testcase.

wking: On Wed, Sep 14, 2016 at 01:39:06AM -0700, Ma Shimiao wrote:
> As you know, almost all validation error out, if config did not
> match container.  Though we added a caveat in help message, the
> runtimetest still could not finish whole testcase.

We can address that when we switch from the current “lots of tests run
from within one container” to a test harness that lets us run “lots of
little one-test containers”.  And I don't think it's worth reading the
config to detect these corner cases when we can say “this doesn't
cover all the corner cases” and then only call runtimetest in
situations where it works ;).

Mashimiao: > We can address that when we switch from the current “lots of tests run from within one container” to a test harness that lets us run “lots of little one-test containers”.  
That means we need to change the whole test structure

wking: On Wed, Sep 14, 2016 at 02:42:37AM -0700, Ma Shimiao wrote:
> > We can address that when we switch from the current “lots of tests
> > run from within one container” to a test harness that lets us run
> > “lots of little one-test containers”.
>
> That means we need to change the whole test structure

I think that is unavoidable [1,2].  But you can certainly rewrite
runtimetest to not carry on instead of dying on errors if you want to
finish the whole testcase [3].

[1]: https://github.com/opencontainers/ocitools/pull/61#issuecomment-235343447
[2]: https://github.com/opencontainers/ocitools/issues/66#issuecomment-235753741
[3]: https://github.com/opencontainers/ocitools/pull/211/files#r78707264

Mashimiao:
@wking  I think adding the caveat  and rewriting runtimetest to not carry on instead of dying on errors in a follow-up PR is more reasonable. 
Can we pass and add this validation first?

wking:
On Wed, Sep 14, 2016 at 03:07:57AM -0700, Ma Shimiao wrote:
> @wking I think adding the caveat and rewriting runtimetest to not
> carry on instead of dying on errors in a follow-up PR is more
> reasonable.  Can we pass and add this validation first?

Sure, although the one-liner I suggest in [1] seems easy enough to add
now ;).

[1]: https://github.com/opencontainers/ocitools/pull/211#discussion_r78610440

cmd/runtimetest/main.go
```diff
@@ -271,6 +271,62 @@ func validateDefaultFS(spec *rspec.Spec) error {
...
+		fStat := fi.Sys().(*syscall.Stat_t)
```
wking: This `Stat` with its type/major/minor checks (and probably the `Stat` existence check too) should get pulled out into it's own helper (`func validateDevice(path string, device *rspec.Device) error`?) so we can use it in `validateDefaultDevices` as well.

Mashimiao: On 09/14/2016 11:46 PM, W. Trevor King wrote:
> This |Stat| with its type/major/minor checks (and probably the |Stat| existence check too) should get pulled out into it's own helper (|func validateDevice(path string, device *rspec.Device) error|?) so we can use it in |validateDefaultDevices| as well.

added |Stat| existence check.
I think helper function is not needed yet, as default device validation does not need so complicated check

wking:
I still have some changes I'd like to make to this PR [1,2], but I
think f0ae1dc2 is a step forward for testing runtime device support at
all.  I'd be happy to see this landed so I can file my further
suggestions and we can decide if they're useful too ;).

[1]: https://github.com/opencontainers/runtime-tools/pull/211#r78610440
[2]: https://github.com/opencontainers/runtime-tools/pull/211#r78776381

mrunalp:
LGTM

PR #219, opened 2016-09-22
sanitize: Add a config-sanitization command

wking:
Based on image-spec, which [currently exposes](https://github.com/opencontainers/image-spec/blob/v0.5.0/serialization.md#container-runconfig-field-descriptions):
- `User` (user and group by ID or name).
  Represented in runtime-spec by `process.user.*`, although I'm clearing
  `additionalGids` for now because the old formats gave no way to
  represent that.
- `Memory` (limit).
  Represented in runtime-spec by `linux.resources.memory.limit` and
  `solaris.cappedMemory.physical`.
- `MemorySwap` (memory + swap limit).
  Represented in runtime-spec by `linux.resources.memory.swap` and
  `solaris.cappedMemory.physical` + `solaris.cappedMemory.swap`.
- `CpuShares` (relative weight vs. other containers).
  Represented in runtime-spec by `linux.resources.cpu.shares`.  Solaris
  has an `ncpus` property, but no shares analog.
- `ExposedPorts` (a set of port/protocol entries).
  This is not covered by runtime-spec, but image-spec was [planning on
  stuffing it into annotations](https://github.com/opencontainers/image-spec/issues/87#issuecomment-224185157).
- `Env` (array of environment variables).
  Represented in runtime-spec by `process.env`.
- `Entrypoint` (array of positional arguments).
  Represented in runtime-spec by `process.cmd`.
- `Cmd` (array of positional arguments).
  Represented in runtime-spec by `process.cmd`.
- `Volumes` (set of directories which should have data volumes mounted
  on them).
  Represented in runtime-spec by `mounts`, although `Volumes` doesn't
  include `source` locations.
- `WorkingDir` (initial working directory of container process).
  Represented in runtime-spec by `process.cwd`.

`Entrypoint` and `Cmd` are slightly different and have a few [different](https://docs.docker.com/engine/reference/builder/#entrypoint) [forms](https://docs.docker.com/engine/reference/builder/#cmd) each.  Both are represented in the runtime-spec config by `process.args`.

This commit sets all other config settings to their default values, and it clears `mounts[].source` to mimic the `Volumes` information.  image-spec currently [sets up `source`-less bind-mounts when translating `Volumes`](https://github.com/opencontainers/image-spec/blob/v0.5.0/image/config.go#L127-L136).

With the sanitization command, _any_ runtime configuration can be sanitized to be just as safe and limited as the current image-spec config.  That allows us to replace the [current](https://github.com/opencontainers/image-spec/blob/v0.5.0/cmd/oci-image-tool/man/oci-image-tool-create-runtime-bundle.1.md):
1. [Unpack](https://github.com/opencontainers/image-spec/blob/v0.5.0/cmd/oci-image-tool/man/oci-image-tool-unpack.1.md) the rootfs.
2. Fetch the config, translate it to runtime-spec's format, and save it as `config.json`.

with:
1. [Unpack](https://github.com/opencontainers/image-spec/blob/v0.5.0/cmd/oci-image-tool/man/oci-image-tool-unpack.1.md) the bundle.
2. Sanitize `config.json` if the file exists in the unpacked bundle.

That gets image-spec out of the business of cherry-picking portions of runtime-spec's format that are portable and/or extending it to support things like `ExposedPorts`.  Miss-matches made some images impossible (there was no way to say “containers launched from this image will need a terminal” without using annotations) and the translation was sometimes awkward (opencontainers/image-spec#87 suggested putting `Entrypoint` in annotations, and the translation [only supported integer IDs for users and groups](https://github.com/opencontainers/image-spec/blob/v0.5.0/image/config.go#L92-L100) while [claiming support for names as well](https://github.com/opencontainers/image-spec/blame/v0.5.0/serialization.md#L181)).

I'm sure there's room for improving the set of properties that are being sanitized, but I'd rather punt on that level of discussion until there's a consensus about whether this approach has legs at all ;).

### DIFF
--- a/cmd/oci-runtime-tool/main.go
+++ b/cmd/oci-runtime-tool/main.go
@@ -27,6 +27,7 @@ func main() {
 	app.Commands = []cli.Command{
 		generateCommand,
 		bundleValidateCommand,
+		sanitizeCommand,
 	}
 
 	if err := app.Run(os.Args); err != nil {
--- a/cmd/oci-runtime-tool/sanitize.go
+++ b/cmd/oci-runtime-tool/sanitize.go
@@ -1,0 +1,67 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+
+	rspec "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/opencontainers/runtime-tools/sanitize"
+	"github.com/urfave/cli"
+)
+
+var sanitizeFlags = []cli.Flag{
+	cli.StringFlag{Name: "output", Usage: "output file (defaults to stdout)"},
+}
+
+var sanitizeCommand = cli.Command{
+	Name:   "sanitize",
+	Usage:  "sanitize an OCI runtime configuration file",
+	Flags:  sanitizeFlags,
+	Before: before,
+	Action: func(context *cli.Context) (err error) {
+		var reader io.ReadCloser
+		if context.NArg() == 0 {
+			reader = os.Stdin
+		} else if context.NArg() == 1 {
+			reader, err = os.Open(context.Args().First())
+			if err != nil {
+				return err
+			}
+			defer reader.Close()
+		} else {
+			return fmt.Errorf("too many arguments (%d > 1)", context.NArg())
+		}
+
+		var config rspec.Spec
+		err = json.NewDecoder(reader).Decode(&config)
+		if err != nil {
+			return err
+		}
+
+		err = reader.Close()
+		if err != nil {
+			return err
+		}
+
+		err = sanitize.Sanitize(&config)
+		if err != nil {
+			return err
+		}
+
+		var writer io.WriteCloser
+		if context.IsSet("output") {
+			writer, err = os.OpenFile(context.String("output"), os.O_WRONLY | os.O_TRUNC, 0)
+			if err != nil {
+				return err
+			}
+			defer writer.Close()
+		} else {
+			writer = os.Stdout
+		}
+
+		encoder := json.NewEncoder(writer)
+		return encoder.Encode(&config)
+	},
+}
--- a/sanitize/sanitize.go
+++ b/sanitize/sanitize.go
@@ -1,0 +1,76 @@
+// Package sanitize removes dangerous and questionably-portable properties from container configurations.
+package sanitize
+
+import (
+	rspec "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+// Santize removes dangerous and questionably-portable properties from container configurations.
+func Sanitize(config *rspec.Spec) (err error) {
+	config.Process.Terminal = false
+	//config.Process.ConsoleSize = nil  // needs runtime-spec#581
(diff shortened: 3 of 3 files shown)


Mashimiao:
sorry,  I can't get the point.
You mean image should contain config.json and reset items to be null which also exposed by image-spec?

wking:
On Thu, Sep 22, 2016 at 01:45:10AM -0700, Ma Shimiao wrote:
> You mean image should contain config.json…

Yes.  And then it can drop ‘config’ from [1] as redundant.

> … and reset items to be null which also exposed by image-spec?

I think you have that turned around.  It's resetting (to the default)
everything which is *not* currently exposed by ‘config’.

Of course, replacing ‘config’ with a runtime-spec config.json doesn't
have to happen when this command lands.  This is just tooling that
would facilitate such a pivot.  Even without an image-spec pivot, this
command is useful for sanitizing third-party config.jsons (however you
acquire them) to get the same level of security currently provided by
application/vnd.oci.image.config.v1+json's ‘config’ property..

[1]: https://github.com/opencontainers/image-spec/blob/v0.5.0/serialization.md#image-json-field-descriptions

vbatts:
@wking this is more confusing in PR than it was on the call yesterday.

I get that having the image and runtime spec decoupled is something to think through, but this PR seems like left-field.

wking:
On Thu, Sep 22, 2016 at 11:59:29AM -0700, Vincent Batts wrote:
> @wking this is more confusing in PR than it was on the call yesterday.

This PR is just a mask over config.json that passes through fields
which image-spec currently sets from [1] and stuff that seemed very
tightly related (e.g. the Solaris memory properties).  I don't see
anything I can to to make that goal less confusing.

On the other hand, perhaps the implementation would be less confusing
if I replaced ccd3f6e9's blacklist approach to mutating the original
config instance with a whitelist approach that created a new config
instance and populated it from the original config.  Something like:

  // Santize removes dangerous and questionably-portable properties from container configurations.
  func Sanitize(config *rspec.Spec) (sanitizedConfig *rspec.Spec, err error) {
    sanitizedConfig.ociVersion = config.ociVersion
    sanitizedConfig.Platform = config.Platform
    …
    return santizedConfig, err
  }

If you feel like that would help, I'm happy to reroll.

[1]: https://github.com/opencontainers/image-spec/blob/v0.5.0/image/config.go#L67-L139

wking:
Anything I can do to help this along?

Mashimiao:
@wking  Do we still need this PR?  image-tools already has `unpack` command.

wking:
> Do we still need this PR? image-tools already has unpack command.

I still prefer this approach.  But given the maintainer resistance to any type of config sharing (e.g. discussion in opencontainers/runtime-spec#611), and the apparent ongoing enthusiasm for translating from the image-spec config (e.g. opencontainers/image-spec#711), I'm not holding my breath on this landiing here ;).  And folk who do like this approach can always do their own whitelisting or blacklisting (e.g. with `jq`).  So if a maintainer wants to close this PR, that's fine with me.

Mashimiao:
OK, I'm gonging to close this PR,

PR #220, opened 2016-09-23
validate: optimize capabilites check

Mashimiao:

By now, if we generate with option --priviliged, the validation of bundle will fail.
capabilities need to be more than original defaultCap has.

### DIFF
--- a/cmd/oci-runtime-tool/validate.go
+++ b/cmd/oci-runtime-tool/validate.go
@@ -4,9 +4,8 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/urfave/cli"
-
 	"github.com/opencontainers/runtime-tools/validate"
+	"github.com/urfave/cli"
 )
 
 var bundleValidateFlags = []cli.Flag{
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -11,6 +11,7 @@ import (
 
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/runtime-tools/generate/seccomp"
+	"github.com/opencontainers/runtime-tools/validate"
 	"github.com/syndtr/gocapability/capability"
 )
 
@@ -754,11 +755,10 @@ func (g *Generator) AddBindMount(source, dest string, options []string) {
 
 // SetupPrivileged sets up the privilege-related fields inside g.spec.
 func (g *Generator) SetupPrivileged(privileged bool) {
-	if privileged {
-		// Add all capabilities in privileged mode.
+	if privileged { // Add all capabilities in privileged mode.
 		var finalCapList []string
 		for _, cap := range capability.List() {
-			if g.HostSpecific && cap > lastCap() {
+			if g.HostSpecific && cap > validate.LastCap() {
 				continue
 			}
 			finalCapList = append(finalCapList, fmt.Sprintf("CAP_%s", strings.ToUpper(cap.String())))
@@ -771,36 +771,6 @@ func (g *Generator) SetupPrivileged(privileged bool) {
 	}
 }
 
-func lastCap() capability.Cap {
-	last := capability.CAP_LAST_CAP
-	// hack for RHEL6 which has no /proc/sys/kernel/cap_last_cap
-	if last == capability.Cap(63) {
-		last = capability.CAP_BLOCK_SUSPEND
-	}
-
-	return last
-}
-
-func checkCap(c string, hostSpecific bool) error {
-	isValid := false
-	cp := strings.ToUpper(c)
-
-	for _, cap := range capability.List() {
-		if cp == strings.ToUpper(cap.String()) {
-			if hostSpecific && cap > lastCap() {
-				return fmt.Errorf("CAP_%s is not supported on the current host", cp)
-			}
-			isValid = true
-			break
-		}
-	}
-
-	if !isValid {
-		return fmt.Errorf("Invalid value passed for adding capability")
-	}
-	return nil
-}
-
 // ClearProcessCapabilities clear g.spec.Process.Capabilities.
 func (g *Generator) ClearProcessCapabilities() {
 	if g.spec == nil {
@@ -811,11 +781,10 @@ func (g *Generator) ClearProcessCapabilities() {
 
 // AddProcessCapability adds a process capability into g.spec.Process.Capabilities.
 func (g *Generator) AddProcessCapability(c string) error {
-	if err := checkCap(c, g.HostSpecific); err != nil {
+	cp := strings.ToUpper(c)
+	if err := validate.CapValid(cp, g.HostSpecific); err != nil {
 		return err
 	}
-
-	cp := fmt.Sprintf("CAP_%s", strings.ToUpper(c))
 
 	g.initSpec()
 	for _, cap := range g.spec.Process.Capabilities {
@@ -830,11 +799,10 @@ func (g *Generator) AddProcessCapability(c string) error {
 
 // DropProcessCapability drops a process capability from g.spec.Process.Capabilities.
 func (g *Generator) DropProcessCapability(c string) error {
-	if err := checkCap(c, g.HostSpecific); err != nil {
+	cp := strings.ToUpper(c)
+	if err := validate.CapValid(cp, g.HostSpecific); err != nil {
 		return err
 	}
-
-	cp := fmt.Sprintf("CAP_%s", strings.ToUpper(c))
 
 	g.initSpec()
 	for i, cap := range g.spec.Process.Capabilities {
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -15,28 +15,29 @@ import (
 	"github.com/Sirupsen/logrus"
 	"github.com/blang/semver"
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/syndtr/gocapability/capability"
 )
 
 const specConfig = "config.json"
 
 var (
 	defaultRlimits = []string{
-		"RLIMIT_CPU",
-		"RLIMIT_FSIZE",
-		"RLIMIT_DATA",
-		"RLIMIT_STACK",
-		"RLIMIT_CORE",
-		"RLIMIT_RSS",
-		"RLIMIT_NPROC",
-		"RLIMIT_NOFILE",
-		"RLIMIT_MEMLOCK",
 		"RLIMIT_AS",
+		"RLIMIT_CORE",
+		"RLIMIT_CPU",
+		"RLIMIT_DATA",
+		"RLIMIT_FSIZE",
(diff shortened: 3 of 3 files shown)


cmd/oci-runtime-tool/validate.go
```diff
@@ -225,8 +210,8 @@ func checkProcess(spec rspec.Spec, rootfs string, hostCheck bool) (msgs []string
 
 	for index := 0; index < len(process.Capabilities); index++ {
 		capability := process.Capabilities[index]
-		if !capValid(capability) {
-			msgs = append(msgs, fmt.Sprintf("capability %q is not valid, man capabilities(7)", process.Capabilities[index]))
+		if !capValid(capability, hostCheck) {
+			msgs = append(msgs, fmt.Sprintf("capability %q is not valid, man capabilities(7)", capability))
```
wking: Complaining by capability (and not by index) makes sense to me.  But now that we're doing that, I think we should replace the current for loop with:

    for _, capability := range process.Capabilities {

cmd/oci-runtime-tool/validate.go
```diff
@@ -465,17 +450,33 @@ func envValid(env string) bool {
...
-		if val == capability {
-			return true
-		}
-	}
+func capValid(c string, hostSpecific bool) bool {
+    cp := strings.ToUpper(c)
+
+    for _, cap := range capability.List() {
+        if cp == fmt.Sprintf("CAP_%s", strings.ToUpper(cap.String())) {
+            if hostSpecific && cap > lastCap() {
```
wking: I'm not sure `List` is the right logic for host-agnostic checks.  The [docs say][1]:

> List returns list of all supported capabilities

What we want here is a list of all *possible* capabilities.  But maybe I'm just misreading the docs.  Does someone have an old kernel around to test this on?

[1]: https://godoc.org/github.com/syndtr/gocapability/capability#List

cmd/oci-runtime-tool/validate.go
```diff
@@ -465,17 +450,33 @@ func envValid(env string) bool {
...
+				return false
+            }
+            return true
+        }
+    }
+
 	return false
 }
 
+func lastCap() capability.Cap {
```
wking: I don't like duplicating code between the generate and validate libraries.  Validation seems more fundamental to me, so can we make a public `LastCap` function here and import that into the generate library?

wking: It looks like your new `capValid` code is also coming over [from generate][1].  Can we move all of that to a public validate function and import that into the generate library as well?

[1]: https://github.com/opencontainers/runtime-tools/blob/7ab77aa560f067c5a7cd4e67a37f6a73f117d07c/generate/generate.go#L930-L948

Mashimiao: I also don't like duplicating code. @liangchenye  is working on a new validate module at #188 .
Maybe we can made a cleanup after that merged or wait until it merged?

cmd/oci-runtime-tool/validate.go
Mashimiao: fixed, thanks.

cmd/oci-runtime-tool/validate.go
wking: I think #188 is ready to land, so I'd rather idle here.

Mashimiao:
@wking  @mrunalp  @liangchenye   rebased.

generate/generate.go
```diff
@@ -679,12 +679,12 @@ func lastCap() capability.Cap {
 	return last
 }
 
-func checkCap(c string, hostSpecific bool) error {
+func ValidCap(c string, hostSpecific bool) error {
 	isValid := false
 	cp := strings.ToUpper(c)
```
wking: I think we can drop this.  The spec [says nothing about case-insensitivity][1], and the values listed in the referenced man page are uppercase.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc2/config.md#process-configuration

generate/generate.go
```diff
@@ -694,7 +694,7 @@ func checkCap(c string, hostSpecific bool) error {
 	}
 
 	if !isValid {
-		return fmt.Errorf("Invalid value passed for adding capability")
+		return fmt.Errorf("Invalid capability value: %s", cp)
```
wking: Do we need “value”?  I'd rather have `Invalid capability: %q`.

generate/generate.go
```diff
@@ -709,12 +709,11 @@ func (g *Generator) ClearProcessCapabilities() {
 
 // AddProcessCapability adds a process capability into g.spec.Process.Capabilities.
 func (g *Generator) AddProcessCapability(c string) error {
-	if err := checkCap(c, g.HostSpecific); err != nil {
+	cp := fmt.Sprintf("CAP_%s", strings.ToUpper(c))
```
wking: Why are we adding the `CAP_` prefix here?  I'd rather require `AddProcessCapability` callers give us a valid string without massaging it internally.

generate/generate.go
```diff
@@ -679,12 +679,12 @@ func lastCap() capability.Cap {
 	return last
 }
 
-func checkCap(c string, hostSpecific bool) error {
+func ValidCap(c string, hostSpecific bool) error {
```
wking: I think `ValidCap` should live in `validate/` and `generate.go` should import the `validate` package.

validate/validate.go
```diff
@@ -217,19 +218,18 @@ func (v *Validator) CheckProcess() (msgs []string) {
...
-		if !rlimitValid(process.Rlimits[index].Type) {
-			msgs = append(msgs, fmt.Sprintf("rlimit type %q is invalid.", process.Rlimits[index].Type))
+	for _, rlimit := range process.Rlimits {
+		if !rlimitValid(rlimit.Type) {
+			msgs = append(msgs, fmt.Sprintf("rlimit type %q is invalid.", rlimit.Type))
 		}
-		if process.Rlimits[index].Hard < process.Rlimits[index].Soft {
-			msgs = append(msgs, fmt.Sprintf("hard limit of rlimit %s should not be less than soft limit.", process.Rlimits[index].Type))
+		if rlimit.Hard < rlimit.Soft {
+			msgs = append(msgs, fmt.Sprintf("hard limit of rlimit %s should not be less than soft limit.", rlimit.Type))
```
wking: `rlimitValid` is an overly-generic name if it only validates the type.  I'd recommend renaming it to `validRlimitType` or moving the hard/soft limit checks inside `validRlimit`.

generate/generate.go
Mashimiao: Fine, fixed.

generate/generate.go
Mashimiao: removed

validate/validate.go
Mashimiao: move hard/soft limit check into rlimitValid

validate/validate.go
```diff
@@ -123,6 +129,8 @@ func (v *Validator) CheckRootfsPath() (msgs []string) {
 	return
 
 }
+
+// CheckSemVer checks v.sepc.Version
```
wking: “sepc” → “spec”

validate/validate.go
```diff
@@ -203,6 +213,7 @@ func checkEventHooks(hookType string, hooks []rspec.Hook, hostSpecific bool) (ms
 	return
 }
 
+// CheckProcess checks v.spec.process
```
wking: “process” → “Process”.

validate/validate.go
```diff
@@ -457,22 +501,16 @@ func envValid(env string) bool {
...
-
-func rlimitValid(rlimit string) bool {
+func rlimitValid(rlimit rspec.Rlimit) error {
 	for _, val := range defaultRlimits {
-		if val == rlimit {
-			return true
+		if val == rlimit.Type {
+			if rlimit.Hard < rlimit.Soft {
+				return fmt.Errorf("hard limit of rlimit %s should not be less than soft limit", rlimit.Type)
+			}
```
wking: I'd rather not bury the hard/soft comparison inside the type check.  I'm fine checking this before the type, because “hard < soft” is *always* wrong, but “type not in `defaultRlimits`” could be “runtime-tools maintainers haven't taught runtime-tools about the new rlimit”.  If you really want the type check first, you could do something like:

    found := false
    for … {
      if val == rlimit.Type {
        found = true
        break
      }
    }
    if !found {
      return fmt.Errorf("rlimit type %q is invalid", rlimit.Type)
    }

Mashimiao:
On 10/22/2016 03:23 PM, W. Trevor King wrote:
> I'd rather not bury the hard/soft comparison inside the type check. I'm fine checking this before the type, because “hard < soft” is /always/ wrong, but “type not in |defaultRlimits|” could be “runtime-tools maintainers haven't taught runtime-tools about the new rlimit”. If you really want the type check first, you could do something like:
> 

I'd like do type check first. 
I think if type was invalid, hard/soft check doesn't make any sense.

> |found := false for … { if val == rlimit.Type { found = true break } } if !found { return fmt.Errorf("rlimit type %q is invalid", rlimit.Type) } |

Maybe I didn't understand your means.
I think the above code has no difference with current code except lack of hard/soft check.

wking:
On Tue, Oct 25, 2016 at 11:21:28AM -0700, Ma Shimiao wrote:
> I'd like do type check first.  I think if type was invalid,
> hard/soft check doesn't make any sense.

But runtime-tools may think the type is invalid when it really isn't
[1].

> I think the above code has no difference with current code except
> lack of hard/soft check.

Exactly.  That lets you handle the hard/soft check independently after
passing through the type-checking loop.  Separating the checks inside
`rlimitValid` makes it more clear that the type and hard/soft checks
are completely orthogonal.

[1]: https://github.com/opencontainers/runtime-tools/pull/220#discussion_r84572001

Mashimiao:
Fixed. @wking  PTAL

wking:
a73a5e8ffd looks good to me, although it would be better to have
squashed that change into the “optimize capabilites check” commit
instead of putting it in the “add comments for exported functions”
commit.

Mashimiao:

liangchenye:
LGTM

Mashimiao:
@hqhq  @mrunalp  PTAL

Mashimiao:
@mrunalp  @hqhq   PTAL

Mashimiao:
@mrunalp  @hqhq  PTAL

Mashimiao:
Rebased. @opencontainers/runtime-tools-maintainers  PTAL

Mashimiao:
@hqhq  @mrunalp @liangchenye   PTAL

hqhq:
LGTM

hqhq:
ping @liangchenye needs re-LGTM.

liangchenye:
LGTM

PR #223, opened 2016-09-27
validate: add hard and soft limit check for rlimit

Mashimiao:

### DIFF
--- a/cmd/oci-runtime-tool/validate.go
+++ b/cmd/oci-runtime-tool/validate.go
@@ -234,6 +234,9 @@ func checkProcess(spec rspec.Spec, rootfs string, hostCheck bool) (msgs []string
 		if !rlimitValid(process.Rlimits[index].Type) {
 			msgs = append(msgs, fmt.Sprintf("rlimit type %q is invalid.", process.Rlimits[index].Type))
 		}
+		if process.Rlimits[index].Hard < process.Rlimits[index].Soft {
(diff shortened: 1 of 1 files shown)


cmd/oci-runtime-tool/validate.go
```diff
@@ -234,6 +234,9 @@ func checkProcess(spec rspec.Spec, rootfs string, hostCheck bool) (msgs []string
 		if !rlimitValid(process.Rlimits[index].Type) {
 			msgs = append(msgs, fmt.Sprintf("rlimit type %q is invalid.", process.Rlimits[index].Type))
 		}
+		if process.Rlimits[index].Hard < process.Rlimits[index].Soft {
+			msgs = append(msgs, fmt.Sprintf("hard limit of rlimit %s should not be less than soft limit.", process.Rlimits[index].Type))
```
wking: f175fae0 looks fine to me vs. the current spec.

I'd rather not refer to array entries by a property of their value (resource `Type`).  I've filed opencontainers/runtime-spec#583, and if/when that lands referencing a particular entry by the resource will be more justified.  If the runtime-spec PR is rejected, I can file a follow-up PR here so we report errors by index.  Either way, I'm fine with this PR landing as it stands now, since checking with some ambiguity around the error message is better than no checking at all ;).

mrunalp:
LGTM

PR #230, opened 2016-10-06
use go > 1.5 since golint drop support for 1.5

liangchenye:
go 1.5.3 fails to run golint in 'travis' file.

Find the reason for this:
https://github.com/golang/lint/commit/a428635c58fe96360e83667e2e1a8343fc292bf0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go:
-  - 1.5.3
+  - 1.6
 
 sudo: false
 


vbatts:
I think I'm okay with this. @crosbymichael is there currently an expectation for golang version with runc, et al?

mrunalp:
@vbatts We are building runc with 1.7.1 so this is fine.

mrunalp:
LGTM

PR #231, opened 2016-10-07
update godeps and use by makefile

Mashimiao:
update dependencies to the latest version
and add directory vendor as _workspace will be deprecated in go1.8


liangchenye:
I'm fine to update this.
But I prefer to do it after most other PRs been merged.

mrunalp:
Yeah, let's get the other ones in.

Mashimiao:
@opencontainers/runtime-tools-maintainers 
Can we let this in first? I think there are many differences between spec v1.0.0-rc1 and v1.0.0-rc2.
The backward spec-go has affected usage. like mentioned in #268

.travis.yml
```diff
@@ -5,14 +5,15 @@ go:
 sudo: false
 
 before_install:
+  - go get github.com/tools/godep
```
wking: Why this addition?  Does Travis start using `godep` somewhere?

Mashimiao: godep will be used in Makefile

wking: Ah, I'd missed that between all the `Godeps` removals and `vendor` additions.

Mashimiao:
As #349 is landed, can we start to review this PR?
@liangchenye  @mrunalp

Mashimiao:
ping @liangchenye  @mrunalp  @hqhq

liangchenye:
`golint` in `make test` will check all the files under `vendor` directory, so the CI fails.

Can you skip this in Makefile? @Mashimiao

Mashimiao:
@liangchenye  yes, updated. I think there is no need to exit with output of golint

liangchenye:
LGTM

hqhq:
@Mashimiao What tool are you using? Just godep or tools like vndr?

Why there's no config file for dependencies?

Mashimiao:
Just use godep. sorry for leave the config behind.
@liangchenye  @hqhq  PTAL

Godeps/Godeps.json
```diff
@@ -49,6 +50,10 @@
 			"ImportPath": "github.com/urfave/cli",
 			"Comment": "v1.19.1",
 			"Rev": "0bdeddeeb0f650497d603c4ad7b20cfe685682f6"
+		},
+		{
+			"ImportPath": "golang.org/x/sys/unix",
+			"Rev": "f3918c30c5c2cb527c0b071a27c35120a6c0719a"
```
hqhq: Why do you need to add this in this PR?

Mashimiao: automatically added by godep, not sure if it should be here. But as unix also is called and depended by runtime-tools, I think may be it's OK

Mashimiao: I find runc or image-tools also add them into vendor directory

hqhq: They use `x/sys/unix` to replace `syscall` package, maybe it's used by the updated vendors, it's fine if is automatically added.

Makefile
```diff
@@ -31,25 +32,20 @@ uninstall:
...
 
 test: .gofmt .govet .golint
 
 .gofmt:
-	OUT=$$(go fmt ./...); if test -n "$${OUT}"; then echo "$${OUT}" && exit 1; fi
+	OUT=$$(go fmt ./... | grep -v vendor); if test -n "$${OUT}"; then echo "$${OUT}" && exit 1; fi
 
 .govet:
-	go vet -x ./...
+	OUT=$$(go vet ./... | grep -v "exit status" | grep -v vendor); if test -n "$${OUT}"; then echo "$${OUT}" && exit 1; fi
```
hqhq: I would do something like:
```
packages = $(shell go list ./... | grep -v vendor)
.govet:
    go vet $(packages)
```

Mashimiao: updated, PTAL

hqhq: Not exactly what I expected, but we can do that later.

hqhq:
LGTM

liangchenye:
LGTM

Issue #228, opened 2016-10-02
container does not  exist when I run test_runtime.sh


heartlock:
container "7df3579a-af8b-402f-b8d3-bc502521841d" does not exist
Runtime runc failed validation

I think we should create container first before start it.
https://github.com/opencontainers/runtime-tools/blob/master/test_runtime.sh#L82

mrunalp:
This is an issue because we are using start in test_runtime.sh
Try replacing `start` -> `runc`.

heartlock:
@mrunalp it works, and I make a pr to fix this issue #232

mrunalp:
@heartlock We need a wrapper process like discussed in #232

liangchenye:
fixed. close

PR #184, opened 2016-08-05
change start to run 

keloyang:
I Follow the instructions(README.md),  and "./test_runtime.sh -r docker-runc" execute failed. So I have this modify.
*    runc start failed if it has no runc create ,so change start to run (create+start), it will avoid "**/state.json: no such file or directory" error.
*    add --tty to command ,the generated config.json will has"terminal: true",  it will avoid the "/dev/console not found" error.
then I can see the validation.
```
root@ubuntu:****opencontainers/ocitools# ./test_runtime.sh -r docker-runc -l debug
-----------------------------------------------------------------------------------
VALIDATING RUNTIME: docker-runb
-----------------------------------------------------------------------------------
DEBU[0000] validating root filesystem                   
DEBU[0000] validating container process                 
DEBU[0000] validating capabilities                      
DEBU[0000] validating hostname                          
DEBU[0000] validating rlimits                           
DEBU[0000] validating mounts exist                      
DEBU[0000] validating linux default filesystem          
DEBU[0000] validating linux default devices             
DEBU[0000] validating sysctls                           
DEBU[0000] validating maskedPaths                       
DEBU[0000] validating readonlyPaths                     
Runtime docker-runb passed validation
```

Signed-off-by: Shukui Yang <yangshukui@huawei.com>

PR #232, opened 2016-10-07
fix issue #228 replace start to run

heartlock:
Signed-off-by: heartlock 21521209@zju.edu.cn

### DIFF
--- a/test_runtime.sh
+++ b/test_runtime.sh
@@ -79,8 +79,17 @@ cp runtimetest ${TESTDIR}
 
 oci-runtime-tool generate --output "${TESTDIR}/config.json" "${TEST_ARGS[@]}" --rootfs '.'
 
-TESTCMD="${RUNTIME} run $(uuidgen)"
+conID=$(uuidgen)
+
+CREATECMD="${RUNTIME} create ${conID}"
+TESTCMD="${RUNTIME} start ${conID}"
 pushd $TESTDIR > /dev/null
+if ! ${CREATECMD}; then
+	info "Runtime is ${RUNTIME}, no need to create before start"
+else
+	info "Runtime is ${RUNTIME}, need to create before start"
+fi
+
 if ! ${TESTCMD}; then
 	error "Runtime ${RUNTIME} failed validation"
 else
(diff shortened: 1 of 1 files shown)


Mashimiao:
This has been talked in #184.  I think you'd better separate it into create and start.

heartlock:
@Mashimiao good idea, I will update it. but I think when the work of issuse #66 is complete, test_runtime.sh will be removed. I don't know if  it is necessary to update it here.

Mashimiao:
I think it is, as test_runtime.sh is the only runtime test tool by now.

Mashimiao:
And there is no need two commits, just one commit will be fine.

test_runtime.sh
```diff
@@ -79,8 +79,17 @@ cp runtimetest ${TESTDIR}
...
+conID=$(uuidgen)
+
+CREATECMD="${RUNTIME} create ${conID}"
+TESTCMD="${RUNTIME} start ${conID}"
 pushd $TESTDIR > /dev/null
+if ! ${CREATECMD}; then
+	info "Runtime is ${RUNTIME}, no need to create before start"
+else
+	info "Runtime is ${RUNTIME}, need to create before start"
+fi
```
heartlock: when exitcode of command "${RUNTIME} create ${conID}" is 1, runtime has no command "create" continue to run command "start". If exitcode of command "${RUNTIME} create ${conID}" is -1, runtime create container fail and print error. but I don't know how to get exitcode of command "${RUNTIME} create ${conID}".

Mashimiao: when executed "${RUNTIME} create ${conID}", you can use `$?` to get the exit code.
just as:
`# sleep 1
# echo $?
0`

heartlock: but if runtime has no command "create", execute  "${RUNTIME} create ${conID}" will cause a crash.https://github.com/opencontainers/runtime-tools/blob/master/test_runtime.sh#L2

Mashimiao: If necessary, I think we can remove it.

Mashimiao: If necessary, I think we can remove it.

heartlock: I think it is unwise to remove it.

Mashimiao: In my opinion, just now it's not necessary to remove.  You can just error out if failed and continue when created succeed.

Mashimiao:
carried in #285

PR #233, opened 2016-10-09
check if Root.Path is Abs

liangchenye:
Add IsAbs check due to runtime spec changes:
https://github.com/opencontainers/runtime-spec/blob/master/config.md#root-configuration

### DIFF
--- a/cmd/oci-runtime-tool/validate.go
+++ b/cmd/oci-runtime-tool/validate.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"path"
 	"path/filepath"
 	"reflect"
 	"strings"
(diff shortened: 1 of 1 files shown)


liangchenye:
@mrunalp PTAL

cmd/oci-runtime-tool/validate.go
```diff
@@ -78,7 +78,7 @@ var bundleValidateCommand = cli.Command{
 			return err
 		}
 
-		configPath := path.Join(inputPath, "config.json")
+		configPath := filepath.Join(inputPath, "config.json")
```
Mashimiao: should we replace all path.* to filepath.* and remove `import "path"`?

liangchenye: I think so, I modified all path in 'validate.go' by updating the PR.

mrunalp: :+1:

mrunalp:
LGTM

PR #253, opened 2016-10-21
generate: add maskedPaths and readonlyPaths options

Mashimiao:


wking:
6afe0d8c looks fine to me.  But pull requests for every setting in runtime-spec seems like a lot of work when:

    $ echo '{}' |
    >   jq '.linux.maskedPaths = ["/proc/kcore", "/proc/latency_stats"]'
    {
      "linux": {
        "maskedPaths": [
          "/proc/kcore",
          "/proc/latency_stats"
        ]
      }
    }
    $ echo '{}' |
    >   jq '.linux.maskedPaths = ["/proc/kcore", "/proc/latency_stats"]' |
    >   jq '.linux.maskedPaths |= . + ["/proc/timer_stats"]'
    {
      "linux": {
        "maskedPaths": [
          "/proc/kcore",
          "/proc/latency_stats",
          "/proc/timer_stats"
        ]
      }
    }

and:

    g.spec.Linux.MaskedPaths = append(g.Spec.Linux.MaskedPaths, "/proc/timer_stats")

are so straightforward.  Are we sure we don't want to revisit [making `Generator.Spec` public][1]?  Then we could have the library just provide helpers for situations that are trickier than an `append(…)` call.

mrunalp:
@wking I think that we can make the field public but let's get this one in for now. LGTM.

PR #256, opened 2016-10-25
validate: add mounts whether nested check

Mashimiao:

### DIFF
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"regexp"
 	"runtime"
 	"strings"
 	"unicode"
@@ -401,15 +402,40 @@ func (v *Validator) CheckMounts() (msgs []string) {
 		return
 	}
 
-	for _, mount := range v.spec.Mounts {
-		if supportedTypes != nil {
-			if !supportedTypes[mount.Type] {
-				msgs = append(msgs, fmt.Sprintf("Unsupported mount type %q", mount.Type))
+	for i, mountA := range v.spec.Mounts {
+		if supportedTypes != nil && !supportedTypes[mountA.Type] {
+			msgs = append(msgs, fmt.Sprintf("Unsupported mount type %q", mountA.Type))
+		}
+		if v.platform == "windows" {
+			if err := pathValid(v.platform, mountA.Destination); err != nil {
+				msgs = append(msgs, err.Error())
+			}
+			if err := pathValid(v.platform, mountA.Source); err != nil {
+				msgs = append(msgs, err.Error())
+			}
+		} else {
+			if err := pathValid(v.platform, mountA.Destination); err != nil {
+				msgs = append(msgs, err.Error())
 			}
 		}
-
-		if !filepath.IsAbs(mount.Destination) {
-			msgs = append(msgs, fmt.Sprintf("destination %v is not an absolute path", mount.Destination))
+		for j, mountB := range v.spec.Mounts {
+			if i == j {
+				continue
+			}
+			// whether B.Desination is nested within A.Destination
+			nested, err := nestedValid(v.platform, mountA.Destination, mountB.Destination)
+			if err != nil {
+				msgs = append(msgs, err.Error())
+				continue
+			}
+			if nested {
+				if v.platform == "windows" && i < j {
+					msgs = append(msgs, fmt.Sprintf("On Windows, %v nested within %v is forbidden", mountB.Destination, mountA.Destination))
+				}
+				if i > j {
+					logrus.Warnf("%v will be covered by %v", mountB.Destination, mountA.Destination)
+				}
+			}
 		}
 	}
 
@@ -706,6 +732,65 @@ func namespaceValid(ns rspec.LinuxNamespace) bool {
 	}
 
 	return true
+}
+
+func pathValid(os, path string) error {
+	if os == "windows" {
+		matched, err := regexp.MatchString("^[a-zA-Z]:(\\\\[^\\\\/<>|:*?\"]+)+$", path)
+		if err != nil {
+			return err
+		}
+		if !matched {
+			return fmt.Errorf("invalid windows path %v", path)
+		}
+		return nil
+	}
+	if !filepath.IsAbs(path) {
+		return fmt.Errorf("%v is not an absolute path", path)
+	}
+	return nil
+}
+
+// Check whether pathB is nested whithin pathA
+func nestedValid(os, pathA, pathB string) (bool, error) {
+	if pathA == pathB {
+		return false, nil
+	}
+	if pathA == "/" && pathB != "" {
+		return true, nil
+	}
+
+	var sep string
+	if os == "windows" {
+		sep = "\\"
+	} else {
+		sep = "/"
+	}
+
+	splitedPathA := strings.Split(filepath.Clean(pathA), sep)
+	splitedPathB := strings.Split(filepath.Clean(pathB), sep)
+	lenA := len(splitedPathA)
+	lenB := len(splitedPathB)
+
+	if lenA > lenB {
+		if (lenA - lenB) == 1 {
+			// if pathA is longer but not end with separator
+			if splitedPathA[lenA-1] != "" {
+				return false, nil
+			}
+			splitedPathA = splitedPathA[:lenA-1]
+		} else {
+			return false, nil
+		}
+	}
+
+	for i, partA := range splitedPathA {
+		if partA != splitedPathB[i] {
+			return false, nil
+		}
+	}
+
+	return true, nil
 }
 
 func deviceValid(d rspec.LinuxDevice) bool {


validate/validate.go
```diff
@@ -294,10 +294,18 @@ func (v *Validator) CheckMounts() (msgs []string) {
...
 
-	if supportedTypes != nil {
-		for _, mount := range v.spec.Mounts {
-			if !supportedTypes[mount.Type] {
-				msgs = append(msgs, fmt.Sprintf("Unsupported mount type %q", mount.Type))
+	for i, fMount := range v.spec.Mounts {
+		if supportedTypes != nil && !supportedTypes[fMount.Type] {
+			msgs = append(msgs, fmt.Sprintf("Unsupported mount type %q", fMount.Type))
+		}
+		for j, bMount := range v.spec.Mounts {
```
wking: Why `fMount` and `bMount`?  Maybe `mount1` and `mount2`?

validate/validate.go
```diff
@@ -294,10 +294,18 @@ func (v *Validator) CheckMounts() (msgs []string) {
...
+			if nestedValid(bMount.Destination, fMount.Destination) {
+				if v.spec.Platform.OS == "windows" {
+					msgs = append(msgs, fmt.Sprintf("On Windows, %v nested within %v is forbidden", bMount.Destination, fMount.Destination))
```
wking: Looking over the spec language again, I'm not clear that it's forbidding covering mounts (e.g. mounting `c:\foo` *after* mounting `c:\foo\bar`).  If that's the intention, I think we need stronger/clearer spec language before forbidding them here (vs. just warning about them, like you're doing for Linux).

validate/validate.go
```diff
@@ -294,10 +294,18 @@ func (v *Validator) CheckMounts() (msgs []string) {
...
+				}
+				if v.spec.Platform.OS == "linux" && i > j {
+					logrus.Warnf("On Linux, %v will be covered by %v", bMount.Destination, fMount.Destination)
```
wking: This seems like something we should be handling in and `else` case so we also print warnings for Solaris and other OSes that don't forbid nesting/covering.

validate/validate.go
```diff
@@ -490,6 +498,14 @@ func namespaceValid(ns rspec.Namespace) bool {
 	return true
 }
 
+func nestedValid(bPath, fPath string) bool {
+	if fPath == bPath {
+		return false
+	}
+
+	return strings.HasPrefix(bPath, fPath)
```
wking: `strings.HasPrefix` is not what you want to use (e.g. `strings.HasPrefix("/ab", "/a")` is true, but `/ab` and `/a` are siblings).  What you want is something like [`filepath.Rel`][1], but using the path syntax appropriate for the `config.json`'s declared OS instead of the path syntax for which `oci-runtime-tool` was compiled.  So `windowsfilepath.Rel` or `posixfilepath.Rel`, depending on `platform.os`.

[1]: https://golang.org/pkg/path/filepath/#Rel

validate/validate.go
Mashimiao: The SPEC says "For the Windows operating system, one mount destination MUST NOT be nested within another mount (e.g., c:\foo and c:\foo\bar)."
I think it clearly says nested mounts are forbidden.

validate/validate.go
Mashimiao: fMount means formerMount, bMount means backMount.
Maybe it's not easy to understand, I will modify them to mountA, mountB

validate/validate.go
Mashimiao: I'm not familiar with other OSes, which ones should we cover?

validate/validate.go
wking: Nested yes.  What I'm not clear on is whether covering mounts are also forbidden.

validate/validate.go
wking: If warning for covering mounts is a good idea for Linux (and I think it is), I expect it is a good idea for all OSes where it's legal (and where it's illegal we should be erroring).

validate/validate.go
Mashimiao: That's really a problem. The SEPC doesn't say.
I will add limit, let's just check nested mounts first.

Mashimiao:
@wking  PTAL

validate/validate.go
```diff
@@ -490,6 +499,30 @@ func namespaceValid(ns rspec.Namespace) bool {
...
+
+	if purePathA == purePathB {
+		return false
+	} else if strings.HasPrefix(purePathB, purePathA) {
+		splitedPaths := strings.SplitN(purePathB, purePathA, 2)
+		if os == "windows" && strings.HasPrefix(splitedPaths[1], "\\") {
+			return true
+		} else if os != "windows" && strings.HasPrefix(splitedPaths[1], "/") {
+			return true
+		}
```
wking: Actually, the function should probably be `ancestorPath(os string, pathA string, pathB string, strict boolean) (isAncestor boolean)`, where when `strict` is false, `isAncestor` is true when `pathA` and `pathB` are equivalent or `pathB` is a strict ancestor of `pathA`.

validate/validate.go
```diff
@@ -490,6 +499,30 @@ func namespaceValid(ns rspec.Namespace) bool {
...
+	} else {
+		purePathA = strings.TrimSuffix(pathA, "/")
+		purePathB = strings.TrimSuffix(pathB, "/")
+	}
+
+	if purePathA == purePathB {
+		return false
+	} else if strings.HasPrefix(purePathB, purePathA) {
+		splitedPaths := strings.SplitN(purePathB, purePathA, 2)
+		if os == "windows" && strings.HasPrefix(splitedPaths[1], "\\") {
```
wking: You'll also want to catch the case where `splitedPaths` (`pathComponents`?) has a length of one (because the initial path was `/` or some such).

validate/validate.go
```diff
@@ -490,6 +499,30 @@ func namespaceValid(ns rspec.Namespace) bool {
...
 }
 
+func nestedValid(os, pathA, pathB string) bool {
+	var purePathA, purePathB string
+	if os == "windows" {
+		purePathA = strings.TrimSuffix(pathA, "\\")
+		purePathB = strings.TrimSuffix(pathB, "\\")
+	} else {
+		purePathA = strings.TrimSuffix(pathA, "/")
+		purePathB = strings.TrimSuffix(pathB, "/")
```
wking: The logic throughout this function would probably be simpler if you open with:

```go
var sep string
if os == "windows" {
  sep = "\\"
} else {
  sep = "/"
}
```

validate/validate.go
```diff
@@ -490,6 +499,30 @@ func namespaceValid(ns rspec.Namespace) bool {
 	return true
 }
 
+func nestedValid(os, pathA, pathB string) bool {
+	var purePathA, purePathB string
```
wking: The paths were passed by value, so you drop `purePath*` and just clobber `pathA` and `pathB`.

validate/validate.go
```diff
@@ -294,10 +294,19 @@ func (v *Validator) CheckMounts() (msgs []string) {
...
+		if supportedTypes != nil && !supportedTypes[mountA.Type] {
+			msgs = append(msgs, fmt.Sprintf("Unsupported mount type %q", mountA.Type))
+		}
+		for j, mountB := range v.spec.Mounts {
+			if nestedValid(v.spec.Platform.OS, mountA.Destination, mountB.Destination) {
+				if v.spec.Platform.OS == "windows" && i < j {
+					msgs = append(msgs, fmt.Sprintf("On Windows, %v nested within %v is forbidden", mountB.Destination, mountA.Destination))
+				}
+				if i > j {
+					// FIXME: add erroring for OS which not support covering mounts
```
wking: Not necessarily a fixme, since covering may be legal on all OSes.

validate/validate.go
```diff
@@ -294,10 +294,19 @@ func (v *Validator) CheckMounts() (msgs []string) {
...
+					logrus.Warnf("On %v, %v will be covered by %v", v.spec.Platform.OS, mountB.Destination, mountA.Destination)
```
wking: I'd drop “On %v,”.  Covering is surprising behavior worth warning about regardless of the target OS.  The nesting error a few lines up, on the other hand, is a Windows-specific error, so opening with “On Windows,” is fine.

validate/validate.go
Mashimiao: OK

validate/validate.go
Mashimiao: Fine.

validate/validate.go
```diff
@@ -490,6 +498,35 @@ func namespaceValid(ns rspec.Namespace) bool {
...
+	var purePathA, purePathB, sep string
+	if os == "windows" {
+		sep = "\\"
+	} else {
+		sep = "/"
+	}
+	purePathA = strings.TrimSuffix(pathA, sep)
+	purePathB = strings.TrimSuffix(pathB, sep)
+
+	if sep == "/" && pathA == sep && strings.HasPrefx(pathB, pathA) {
```
wking: [Missing a `strings` import][1]?

[1]: https://travis-ci.org/opencontainers/runtime-tools/builds/170737217#L220

validate/validate.go
```diff
@@ -490,6 +498,35 @@ func namespaceValid(ns rspec.Namespace) bool {
...
+	}
+
+	var purePathA, purePathB, sep string
+	if os == "windows" {
+		sep = "\\"
+	} else {
+		sep = "/"
+	}
+	purePathA = strings.TrimSuffix(pathA, sep)
+	purePathB = strings.TrimSuffix(pathB, sep)
```
wking: I [still][1] think these could be:

    pathA = strings.TrimSuffix(pathA, sep)

and I don't think you need the `sep == "/" && pathA == sep` qualifiers below.

[1]: https://github.com/opencontainers/runtime-tools/pull/256#discussion_r84948410

Mashimiao:
@opencontainers/runtime-tools-maintainers  @wking  PTAL
I have updated the PR. Maybe the code is not very elegant.
But with my own test, I believe not matter invalid paths windows's or linux's or paths like C:\\A, C:\A\B, /, /A, //A//B//C/, ///C//B, /A/C will be checked very well.

validate/validate.go
```diff
@@ -490,6 +504,89 @@ func namespaceValid(ns rspec.Namespace) bool {
 	return true
 }
 
+// Check whether pathB is nested whithin pathA
+func nestedValid(os, pathA, pathB string) (bool, error) {
```
hqhq: I didn't read through, according to your comments, isn't this something we can do with just `filepath.Clean()` and `filepath.Rel()`?

And I don't know how you handle with different oses other places in runtime-tools, but why not separate them in different files, distinguish with build tags or different command options. I'll be more extendable and make codes neat and clean.

Mashimiao:
On 11/04/2016 03:11 PM, Qiang Huang wrote:
> I didn't read through, according to your comments, isn't this something we can do with just |filepath.Clean()| and |filepath.Rel()|?

Well, with Clan() and Rel(), we still can't find out invalid path
and Rel() is not very good to deal with paths as /c and /b. it believes they are relative.

> 
> And I don't know how you handle with different oses other places in runtime-tools, but why not separate them in different files, distinguish with build tags or different command options. I'll be more extendable and make codes neat and clean.
> 

There are not too many places in runtime-tools different for different oses.
So, currently I prefer to use compatible code to handle them.

wking:
In case it helps make my explicit-OS filepath idea [1] clearer, I've
stubbed it out in [2].  I think that makes the “Is this an ancestor”
function a lot more readable, but even in my stub there are a lot of
FIXMEs.  On the other hand, I don't think either of us are properly
handling .. at the moment.

Instead of duplicating Go's (complicated and well-tested) stdlib
functionality, I'd rather fork it into a separate explicit-OS library
and consume that library here.  That would cover most (all?) of the
current FIXMEs in [2].

[1]: https://github.com/opencontainers/runtime-tools/pull/259#issuecomment-257213064
[2]: https://gist.github.com/wking/4ec91412fbd72f4bf7b22193b96b6bc0

Mashimiao:
Hi @wking  I don't separating explicit-OS library  and build with build tags can solve the question you raised about validating Windows container's config on Linux platform or validating Linux container's config on Windows platform.  As you know, Go just does separating explicit-OS library, but can't solve the above question

hqhq:
> Well, with Clan() and Rel(), we still can't find out invalid path 

Path validation is orthogonal in this case, I don't think you should consider that, otherwise, do you do path validation in every case that a file path is involved? Even you want to do that, you can do it in a separate function named like `validPath()` or something, and that's another topic.

> and Rel() is not very good to deal with paths as /c and /b. it believes they are relative.

What do you mean about this, I don't get it.

Mashimiao:
On 11/04/2016 05:32 PM, Qiang Huang wrote:
> Path validation is orthogonal in this case, I don't think you should consider that, otherwise, do you do path validation in every case that a file path is involved? Even you want to do that, you can do it in a separate function named like |validPath()| or something, and that's another topic.
> 

I think this is a good idea. I can separate it.

>     and Rel() is not very good to deal with paths as /c and /b. it believes they are relative.
> 
> What do you mean about this, I don't get it.
> 

I mean Rel() is not very good to judge whether paths like /c and /b is nested or not.
And as I reply to @wking above, separating based on OSes can't solve the problem we are facing,
validating Linux container's config on windows platform or validating windows container's config on Linux platform.

wking:
On Fri, Nov 04, 2016 at 01:52:23AM -0700, Ma Shimiao wrote:
> Hi @wking I don't separating explicit-OS library and build with
> build tags can solve the question you raised about validating
> Windows container's config on Linux platform or validating Linux
> container's config on Windows platform.  As you know, Go just does
> separating explicit-OS library, but can't solve the above question

Go has an *implicit OS* file/filepath using build tags to build it for
just the $GOOS.  My gist has an *explicit OS* API so you can call (for
example):

  IsAbs("windows", "C:\\a\\b\\c")

and get ‘true’ regardless of whether your $GOOS is windows or not.
I'm replacing Go's file/filepath's build-time OS decision with a
function-call-time OS decision.

Mashimiao:
ping @wking  @opencontainers/runtime-tools-maintainers

validate/validate.go
```diff
@@ -401,15 +402,40 @@ func (v *Validator) CheckMounts() (msgs []string) {
...
+		} else {
+			if err := pathValid(v.platform, mountA.Destination); err != nil {
+				msgs = append(msgs, err.Error())
 			}
 		}
-
-		if !filepath.IsAbs(mount.Destination) {
-			msgs = append(msgs, fmt.Sprintf("destination %v is not an absolute path", mount.Destination))
+		for j, mountB := range v.spec.Mounts {
+			if i == j {
```
q384566678: Change to  `i >= j` and make the following changes will be better ：

```
	if nested {
			if v.platform == "windows"  {
				msgs = append(msgs, fmt.Sprintf("On Windows, %v nested within %v is forbidden", mountB.Destination, mountA.Destination))
			} else {
				logrus.Warnf("%v will be covered by %v", mountB.Destination, mountA.Destination)
			}
		}
```

q384566678:
LGTM

liangchenye:
LGTM

PR #260, opened 2016-10-26
cmd/runtimetest: Add an app description

wking:
As I [recommended earlier](https://github.com/opencontainers/runtime-tools/pull/211#r78610440) in the context of checking for all configured devices.  An example would be:

```
{
  …
  "linux": {
    "devices": [
      {
        "path": "/dev/fuse",
        …
      }
    ],
  },
  "hooks": {
    "prestart": [
      {
        "path": "/bin/rm",
        "args": ["rm", "/dev/fuse"]
      }
    }
  }
}
```

where the resulting container (when created by a conformant runtime) would not have the `/dev/fuse` entry `runtimetest`'s `linux.devices` check is looking for.

This PR also bumps the urfave/cli dependency to pick up urfave/cli#543.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -1,7 +1,6 @@
 {
 	"ImportPath": "github.com/opencontainers/runtime-tools",
 	"GoVersion": "go1.4",
-	"GodepVersion": "v74",
 	"Deps": [
 		{
 			"ImportPath": "github.com/Sirupsen/logrus",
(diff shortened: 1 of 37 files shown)


Mashimiao:
Rejected
You should push this PR compared with master branch.

wking:
On Tue, Oct 25, 2016 at 10:50:04PM -0700, Ma Shimiao wrote:
> You should push this PR compared with master branch.

If it lands in v1.0.0.rc1, this commit will make it to master via #181
or a successor.  If/when #181 is rejected (hopefully with more clarity
around the plan for maintaining stable branches) I'll port this change
to the appropriate branch.  In the interim, it's Apache 2, so others
can port it around as they see fit.

Mashimiao:
@wking   would you mind to create new PR to contribute to master branch?

Mashimiao:
Moved to #277

PR #261, opened 2016-10-26
add pullapprove.yml as PullApprove enabled

Mashimiao:
Signed-off-by: Ma Shimiao mashimiao.fnst@cn.fujitsu.com

PR #262, opened 2016-10-29
Add RUNTIME_TOOLS_LINK in Makefile

hqhq:
So we can build runtime_tools project anyway.

Signed-off-by: Qiang Huang h.huangqiang@huawei.com

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,9 +2,10 @@ PREFIX ?= $(DESTDIR)/usr
 BINDIR ?= $(DESTDIR)/usr/bin
 
 BUILDTAGS=
+RUNTIME_TOOLS_LINK := $(CURDIR)/Godeps/_workspace/src/github.com/opencontainers/runtime-tools
 export GOPATH:=$(CURDIR)/Godeps/_workspace:$(GOPATH)
 
-all:
+all: $(RUNTIME_TOOLS_LINK)
 	go build -tags "$(BUILDTAGS)" -o oci-runtime-tool ./cmd/oci-runtime-tool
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -24,6 +25,10 @@ install: man
 
 clean:
 	rm -f oci-runtime-tool runtimetest *.1
+	rm -f $(RUNTIME_TOOLS_LINK)
+
+$(RUNTIME_TOOLS_LINK):
+	ln -sfn $(CURDIR) $(RUNTIME_TOOLS_LINK)
```
wking: `-n` is [not in POSIX][1].  Do we really need it to support this use-case?

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/ln.html

wking:
This feels sort of like opencontainers/image-tools#52.  runtime-tools
builds fine as it stands if you've checked it out to an appropriate
location in your $GOPATH.

hqhq:
@wking It's ok we don't use `-n`, I've updated. And I know runtime-tools builds fine if I checked it out to an appropriate location in my $GOPATH, the point is users don't have checked it out to an appropriate location in their $GOPATH and they can still build runtime-tools.

It makes it more convenient like what we did in runc.

mrunalp:
LGTM

caniszczyk:
LGTM

vbatts:
LGTM

vbatts:
@caniszczyk FYI @mrunalp's LGTM did not trigger pullapprove

wking:
> @mrunalp's LGTM did not trigger pullapprove

His LGTM predates #261 landing, so he wasn't cleared vs PullApprove when he made the comment.

PR #263, opened 2016-10-29
Fix runc command

hqhq:
Signed-off-by: Qiang Huang h.huangqiang@huawei.com


wking:
Dup of #184.  See also #232 and #228.

hqhq:
@wking OK, thanks.

PR #265, opened 2016-11-06
generate: make AddBindMount() options a slice

cyphar:
This allows for the set of options to be more useful than just "rw" or
"ro". It also makes it match the AddTmpfsMount far more.

We should also probably implement `AddMount` which just simply
adds a mount.

Signed-off-by: Aleksa Sarai <asarai@suse.com>

### DIFF
--- a/cmd/oci-runtime-tool/generate.go
+++ b/cmd/oci-runtime-tool/generate.go
@@ -17,7 +17,7 @@ var generateFlags = []cli.Flag{
 	cli.StringFlag{Name: "apparmor", Usage: "specifies the the apparmor profile for the container"},
 	cli.StringFlag{Name: "arch", Value: runtime.GOARCH, Usage: "architecture the container is created for"},
 	cli.StringSliceFlag{Name: "args", Usage: "command to run in the container"},
-	cli.StringSliceFlag{Name: "bind", Usage: "bind mount directories src:dest:(rw,ro)"},
+	cli.StringSliceFlag{Name: "bind", Usage: "bind mount directories src:dest[:options...]"},
 	cli.StringSliceFlag{Name: "cap-add", Usage: "add Linux capabilities"},
 	cli.StringSliceFlag{Name: "cap-drop", Usage: "drop Linux capabilities"},
 	cli.StringFlag{Name: "cgroup", Usage: "cgroup namespace"},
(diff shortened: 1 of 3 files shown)


wking:
Through 878097f looks good to me.

On Sun, Nov 06, 2016 at 04:58:24AM -0800, Aleksa Sarai wrote:
> We should also probably implement `AddMount` which just simply
> adds a mount.

+1.

cmd/oci-runtime-tool/generate.go
```diff
@@ -17,7 +17,7 @@ var generateFlags = []cli.Flag{
 	cli.StringFlag{Name: "apparmor", Usage: "specifies the the apparmor profile for the container"},
 	cli.StringFlag{Name: "arch", Value: runtime.GOARCH, Usage: "architecture the container is created for"},
 	cli.StringSliceFlag{Name: "args", Usage: "command to run in the container"},
-	cli.StringSliceFlag{Name: "bind", Usage: "bind mount directories src:dest:(rw,ro)"},
+	cli.StringSliceFlag{Name: "bind", Usage: "bind mount directories src:dest[:option1:option2:...]"},
```
Mashimiao: How about change it to `[:options]`?
`[***]` just means this can be set multiple times.

cmd/oci-runtime-tool/generate.go
```diff
@@ -491,18 +491,18 @@ func parseTmpfsMount(s string) (string, []string, error) {
...
 	bparts := strings.SplitN(s, ":", 3)
 	switch len(bparts) {
 	case 2:
 		source, dest = bparts[0], bparts[1]
 	case 3:
-		source, dest, options = bparts[0], bparts[1], bparts[2]
+		source, dest, options = bparts[0], bparts[1], strings.Split(bparts[2], ":")
 	default:
-		return source, dest, options, fmt.Errorf("--bind should have format src:dest:[options]")
+		return source, dest, options, fmt.Errorf("--bind should have format src:dest[:option1:option2:...]")
```
Mashimiao: comment as above

man/oci-runtime-tool-generate.1.md
```diff
@@ -29,15 +29,15 @@ read the configuration from `config.json`.
 
   --args "/usr/bin/httpd" --args "-D" --args "FOREGROUND"
 
-**--bind**=*[[HOST-DIR:CONTAINER-DIR][:OPTIONS]]*
+**--bind**=*[[HOST-DIR:CONTAINER-DIR][:OPTIONS:...]]*
```
Mashimiao: I think there is no need to add `...`

cyphar: I think most commands have the `...` suffix if you can specify multiple of a type. But I'll change it.

cmd/oci-runtime-tool/generate.go
wking: I agree with @cyphar, and there are docs for this in http://man7.org/linux/man-pages/man7/man-pages.7.html

Mashimiao: Well, I didn't find any rules about multiple options format in man7.
But from most of pages in man-pages, `[:options]` or `[:options...]` or `[:options]...` are all acceptable.

cyphar:
@Mashimiao I changed to `[:options...]` and squashed my commits. PTAL.

Mashimiao:
LGTM

cmd/oci-runtime-tool/generate.go
wking: From man-pages(7):
> Brackets (`[]`) surround optional arguments, vertical bars (`|`) separate choices, and ellipses (`...`) can be repeated.

Mashimiao:
On 11/07/2016 12:36 PM, W. Trevor King wrote:
> From man-pages(7):
> 
>     Brackets (|[]|) surround optional arguments, vertical bars (|||) separate choices, and ellipses (|...|) can be repeated.
> 


Oh, it really is,  thanks for your information.

mrunalp:
LGTM

PR #270, opened 2016-11-15
generate: Respect --os when generating default templates

wking:
Also add `--host-specific` checks to `SetPlatformOS` and `SetPlatformArch`.  The `SetPlatformArch` check is currently just a warning (details in eda1ac7, validate: With --host-specific, compare config platform vs. runtime, 2016-08-16) but I've added an error to its API so callers will be ready when we do actually raise errors (e.g. for `--arch arm` on an amd64 box).

Also add IsSet checks to `--os` and `--arch` so we don't clobber their template values.  Before this commit on my amd64 Linux box:

    $ echo '{"platform": {"os": "windows", "arch": "386"}}' >template.json
    $ ocitools generate --template template.json | jq .platform
    {
      "os": "linux",
      "arch": "amd64"
    }

After this commit:

    $ ocitools generate --template template.json | jq .platform
    {
      "os": "windows",
      "arch": "386"
    }

This commit was previously part of #194, but I've pulled it out into its own PR because @Mashimiao [isn't sold yet][1] ;).

[1]: https://github.com/opencontainers/runtime-tools/pull/194#issuecomment-260518219

### DIFF
--- a/cmd/oci-runtime-tool/generate.go
+++ b/cmd/oci-runtime-tool/generate.go
@@ -92,21 +92,23 @@ var generateCommand = cli.Command{
 	Before: before,
 	Action: func(context *cli.Context) error {
 		// Start from the default template.
-		specgen := generate.New()
+		specgen, err := generate.New(nil)
+		if err != nil {
+			return err
+		}
 
 		var template string
 		if context.IsSet("template") {
 			template = context.String("template")
 		}
 		if template != "" {
-			var err error
 			specgen, err = generate.NewFromFile(template)
 			if err != nil {
 				return err
 			}
 		}
 
-		err := setupSpec(&specgen, context)
+		err = setupSpec(&specgen, context)
 		if err != nil {
 			return err
 		}
--- a/cmd/oci-runtime-tool/generate.go
+++ b/cmd/oci-runtime-tool/generate.go
@@ -3,7 +3,6 @@ package main
 import (
 	"fmt"
 	"os"
-	"runtime"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


generate/generate.go
```diff
@@ -30,8 +31,14 @@ type ExportOptions struct {
 	Seccomp bool // seccomp toggles if only seccomp should be exported
 }
 
-// New creates a spec Generator with the default spec.
-func New() Generator {
+// New creates a spec Generator with the default spec for the target
+// OS (which defaults to runtime.GOOS).
+func New(os *string) (generator Generator, err error) {
+	var goos string
+	goos = runtime.GOOS
```
mrunalp:
```go
goos := runtime.GOOS
```

mrunalp:
make failed in the build.

cmd/oci-runtime-tool/generate.go
```diff
@@ -92,21 +91,28 @@ var generateCommand = cli.Command{
 	Before: before,
 	Action: func(context *cli.Context) error {
 		// Start from the default template.
-		specgen := generate.New()
+		var osPointer *string
```
mrunalp: Maybe rename the variable to `specOs`? I don't like that we have `Pointer` in the name. Go doesn't recommend hungarian notation.

wking:
On Tue, Nov 15, 2016 at 11:25:10AM -0800, Mrunal Patel wrote:
> make failed in the build. 

This PR is for the tip commit from #194.  Don't bother reviewing it
until #194 lands.

liangchenye:
Since platform is removed, close this PR.

PR #278, opened 2016-11-24
Generate: add hooks timeout and env options

Mashimiao:


liangchenye:
I think we should join 'timeout/args/env' together.
We might have a list of hooks with a same path.

Mashimiao:
> I think we should join 'timeout/args/env' together.

You mean just implement timeout and env adding with current --prestart,--poststart,--poststop?
Or add more options like --prestart-env?

q384566678:
> You mean just implement timeout and env adding with current --prestart,--poststart,--poststop?

I agree with this method.

cmd/oci-runtime-tool/generate.go
```diff
@@ -332,14 +338,36 @@ func setupSpec(g *generate.Generator, context *cli.Context) error {
...
+			g.AddPostStartHook(path, args)
+		}
+	}
+
+	if context.IsSet("hooks-poststart-env") {
+		postStartEnvs := context.StringSlice("hooks-poststart-env")
+		for _, postStartEnv := range postStartEnvs {
+			path, env, err := parseHookEnv(postStartEnv)
+			if err != nil {
+				return nil
```
q384566678: return err

Mashimiao: updated, thanks

q384566678:
LGTM

liangchenye:
LGTM

PR #273, opened 2016-11-17
add CheckSemVer, CheckRootfsPath, CheckPlatform unit test

liangchenye:
Signed-off-by: liangchenye <liangchenye@huawei.com>

PR #288, opened 2016-12-12
cmd: generate: add --linux-namespace-* family of flags

cyphar:
This allows users to modify what namespaces are used by a container.


mrunalp:
LGTM

mrunalp:
@Mashimiao @liangchenye @wking PTAL

Mashimiao:
we already have --user --memory --ipc --mount --network --pid --uts to add namespaces, I think `--linux-namespace-add` conflicts with them.

cyphar:
@Mashimiao Yeah, you're right. However, we should probably have something like `--linux-namespace-remove` and `--linux-namespace-remove-all`.

Mashimiao:
If the value of current namespace options is `host`, we will remove existing namespace. For example,
if we set --uts=host, we will remove existing uts namespace from config.
But I think that --linux-namespace-* is clear for users to understand and use.
How about removing namespace options like --uts, --user, --pid, etc? @mrunalp  @liangchenye  @wking

wking:
On Mon, Dec 12, 2016 at 10:33:19PM -0800, Ma Shimiao wrote:
> If the value of current namespace options is `host`, we will remove
> existing namespace.

Ah, right, I'd forgotten about that.

> But I think that --linux-namespace-* is clear for users to
> understand and use.  How about removing namespace options like
> --uts, --user, --pid, etc? @mrunalp @liangchenye @wking

Sounds good to me (it has better namespacing).  This PR still needs
the associated man-page changes.

cyphar:
I've updated this PR so that it removes the old namespace flags, and added documentation changes for them.

Mashimiao:
@cyphar  please also update completions/bash file

cyphar:
@Mashimiao Done. There were a few other flags not included or incorrectly classified, so I've fixed it up.

cmd/oci-runtime-tool/generate.go
```diff
@@ -608,6 +613,18 @@ func parseRlimit(rlimit string) (string, uint64, uint64, error) {
 	return parts[0], uint64(hard), uint64(soft), nil
 }
 
+func parseNamespace(ns string) (string, string, error) {
+	parts := strings.Split(ns, ":")
```
wking: `Split` → `SplitN`?  Linux allows paths to contain colons.

cmd/oci-runtime-tool/generate.go
```diff
@@ -296,12 +290,11 @@ func setupSpec(g *generate.Generator, context *cli.Context) error {
 		gidMaps = context.StringSlice("gidmappings")
 	}
 
+	// Add default user namespace.
 	if len(uidMaps) > 0 || len(gidMaps) > 0 {
-		needsNewUser = true
+		g.AddOrReplaceLinuxNamespace("user", "")
```
wking: This will silently create an invalid config with `--uidmappings=… --linux-namespace-add=user:foo`, because the resulting config with call for tweaking a joined namespace.  I'm pro-tweaking though, and we're already doing something silent about this case, so I'm ok with that ;).  Folks who want to ensure validity can pass their final config through to `validate` as a separate step.

cyphar: I don't think it'll generate an invalid config, because the `--linux-namespace-add` handling is done afterwards and so the `user` entry will be replaced.

cyphar: Well, sorry, it will generate an invalid config, but the user asked us to generate it...

cmd/oci-runtime-tool/generate.go
cyphar: Fixed.

wking:
a0f56f2 looks good to me.

cmd/oci-runtime-tool/generate.go
```diff
@@ -608,6 +613,18 @@ func parseRlimit(rlimit string) (string, uint64, uint64, error) {
 	return parts[0], uint64(hard), uint64(soft), nil
 }
 
+func parseNamespace(ns string) (string, string, error) {
+	parts := strings.SplitN(ns, ":", 2)
```
Mashimiao: If we don't check parts[0], then `:testpat` will also be a valid value. Though user set it, I think we should warn user it's invalid, as I did in #292

wking: You can validate it against the spec's whitelist.  But I'm fine punting validation to a post-generate call (for callers who care).  Maybe they intend to fix the values in post-processing.

cyphar: I've fixed it to check `parts[0]`. The whitelist checking should be done in the generate library, and I'll punt it for now. I mean, we don't actually have tests for `generate` at all -- so maybe we should do that first? :wink:

cmd/oci-runtime-tool/generate.go
```diff
@@ -608,6 +613,22 @@ func parseRlimit(rlimit string) (string, uint64, uint64, error) {
...
+	}
+
+	switch len(parts) {
+	case 1:
+		return parts[0], "", nil
+	case 2:
+		return parts[0], parts[1], nil
+	}
+
+	panic("should not be reached")
```
Mashimiao: Actually, I don't like this. the panic error message does not tell user what's wrong clear.
I think we should give user clear error message and they don't have to hack code to find out what's wrong

cmd/oci-runtime-tool/generate.go
Mashimiao: what kind of tests?

cmd/oci-runtime-tool/generate.go
cyphar: It's a `panic` because you literally cannot reach that part of the code. `len(parts)` can only have a length of `0`, `1`, and `2` and all cases have been handled. If I remove the `panic` you get errors about no return statement, even though you cannot reach that line.

cyphar: I can make the function longer to make you happy if you really want.

Mashimiao: All cases of `len(parts)` are handled, so we will never reach `panic`. In my opinion, `panic here` make no sense.

cyphar: So you want to return an error even though the actual line will never be hit? From my experience, people use `panic` in these situations because it's more clear that if you ever hit that line it's a programming error (not something that should happen normally).

Mashimiao: No, I mean do not need `panic` or `return` here. Just remove this `panic`, because it will never be reached

cyphar: Go won't compile that program, which is why I added `panic`... Go isn't clever enough to know that the line won't be hit.

cyphar: I mentioned this earlier in the thread.

```
go build -tags "" -o oci-runtime-tool ./cmd/oci-runtime-tool
# runtime-tools/cmd/oci-runtime-tool
../../.local/src/runtime-tools/cmd/oci-runtime-tool/generate.go:629: missing return at end of function
```

Mashimiao: How about this?
func parseNamespace(ns string) (string, string, error) {
	var nsType, nsPath string
	parts := strings.SplitN(ns, ":", 2)
	if len(parts) == 0 || parts[0] == "" {
		return "", "", fmt.Errorf("invalid namespace value: %s", ns)
	}

	switch len(parts) {
	case 1:
		nsType, nsPath = parts[0], ""
	case 2:
		nsType, nsPath = parts[0], parts[1]
	}

	return nsType, nsPath, nil
}

cyphar: Removed the panic. PTAL.

cmd/oci-runtime-tool/generate.go
cyphar: There aren't any unit tests or integration tests for `oci-runtime-tool`. We should probably add some. Also, this tool isn't versioned which makes [packaging annoying](https://build.opensuse.org/package/show/Virtualization:containers/oci-runtime-tools).

cyphar: More importantly, as far as I can tell there's no tests of the **validation** -- which is quite worrying.

Mashimiao: Yes, we should add unit tests.  unit tests for validation already in #273

Mashimiao:
LGTM

cmd/oci-runtime-tool/generate.go
wking: I'd floated some sharness stubs for testing in #180. Happy to rebase if there is renewed interest.

cmd/oci-runtime-tool/generate.go
```diff
@@ -461,6 +454,32 @@ func setupSpec(g *generate.Generator, context *cli.Context) error {
...
+	if context.IsSet("linux-namespace-remove") {
+		namespaces := context.StringSlice("linux-namespace-remove")
+		for _, name := range namespaces {
+			if err := g.RemoveLinuxNamespace(name); err != nil {
+				return err
+			}
+		}
+	}
+
+	if context.IsSet("linux-namespace-remove-all") {
```
Mashimiao: This should be `context.Bool("linux-namespace-remove-all")`. If not, when linux-namespace-remove-all is false, ClearLinuxNamespaces function will still be called.

cyphar: I was copying what `--rlimits-remove-all` and `--seccomp-remove-all` (the only other flags of this form) do. I don't disagree, it's just that the problem exists with other flags too.

Mashimiao: Yes, I just pushed a PR to fix it

Mashimiao:
LGTM

dqminh:
LGTM

Issue #290, opened 2016-12-14
environment generation causes duplicate entries


cyphar:
If you set an environment variable to a different value, you will get two entries in `process.env`. Apart from being an eyesore, POSIX doesn't actually specify what should happen if you have two variables set with the same name.

```
% oci-runtime-tool generate --env A=B --env A=C
...
                "env": [
                        "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
                        "TERM=xterm",
                        "A=B",
                        "A=C"
                ],
...
```

IMO we should change the environment setting signature to something like `AddProcessEnv(key, value string)`. Or we add parsing to the code so we replace entries if we see them.

wking:
On Wed, Dec 14, 2016 at 06:31:34AM -0800, Aleksa Sarai wrote:
> IMO we should change the environment setting signature to something
> like `AddProcessEnv(key, value string)`. Or we add parsing to the
> code so we replace entries if we see them.

Having the signature take explicit key/value args sounds good to me,
because it will make an eventual RemoveProcessEnv(key string) more
consistent.

There's an outside chance that someone would want a no-clobber version
of AddProcessEnv, but until we actually hear someone ask for it I
think we can go ahead with a signature/implementation that always
clobbers when a key already exists in the environment variable array.

PR #291, opened 2016-12-14
generate: replace duplicate env entries

cyphar:
If a given environment variable has already been set, replace the
variable rather than appending it to the environment list. This avoids
issues with POSIX (which doesn't define what to do with duplicate
values).

The method of determining whether a variable has been set is by checking
that the VAR_NAME= prefix of both strings is the same.

Fixes: #290

### DIFF
--- a/cmd/oci-runtime-tool/generate.go
+++ b/cmd/oci-runtime-tool/generate.go
@@ -208,7 +208,11 @@ func setupSpec(g *generate.Generator, context *cli.Context) error {
 		}
 
 		for _, env := range envs {
-			g.AddProcessEnv(env)
+			name, value, err := parseEnv(env)
+			if err != nil {
+				return err
+			}
+			g.AddProcessEnv(name, value)
 		}
 	}
 
@@ -738,6 +742,22 @@ func readKVStrings(files []string, override []string) ([]string, error) {
 	envVariables = append(envVariables, override...)
 
 	return envVariables, nil
+}
+
+// parseEnv splits a given environment variable (of the form name=value) into
+// (name, value). An error is returned if there is no "=" in the line or if the
+// name is empty.
+func parseEnv(env string) (string, string, error) {
+	parts := strings.SplitN(env, "=", 2)
+	if len(parts) != 2 {
+		return "", "", fmt.Errorf("environment variable must contain '=': %s", env)
+	}
+
+	name, value := parts[0], parts[1]
+	if name == "" {
+		return "", "", fmt.Errorf("environment variable must have non-empty name: %s", env)
+	}
+	return name, value, nil
 }
 
 // parseEnvFile reads a file with environment variables enumerated by lines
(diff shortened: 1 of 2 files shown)


generate/generate.go
```diff
@@ -334,6 +334,17 @@ func (g *Generator) ClearProcessEnv() {
 // AddProcessEnv adds env into g.spec.Process.Env.
 func (g *Generator) AddProcessEnv(env string) {
 	g.initSpec()
+	// I'm not sure what an environment variable without an = in it means, but
+	// better safe than sorry.
+	if strings.Contains(env, "=") {
```
wking: I'd rather error if the string does not contain an equal sign.  From [POSIX][1]:

> These strings have the form *name*=*value*…

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html

generate/generate.go
```diff
@@ -334,6 +334,17 @@ func (g *Generator) ClearProcessEnv() {
 // AddProcessEnv adds env into g.spec.Process.Env.
 func (g *Generator) AddProcessEnv(env string) {
```
wking: I'd rather [take explict key and value args][1], which also avoids the “what happens if `env` doesn't contain an equal sign?” case.

[1]: https://github.com/opencontainers/runtime-tools/issues/290#issuecomment-267103862

cyphar:
There, I've updated it.

/ping @opencontainers/runtime-tools-maintainers

liangchenye:
LGTM

wking:
c027ae7 looks good to me.

cmd/oci-runtime-tool/generate.go
```diff
@@ -740,6 +744,17 @@ func readKVStrings(files []string, override []string) ([]string, error) {
...
 
+// parseEnv splits a given environment variable (of the form name=value) into
+// (name, value). An error is returned if there is no "=" in the line.
+func parseEnv(env string) (string, string, error) {
+	if !strings.Contains(env, "=") {
+		return "", "", fmt.Errorf("invalid env: environment variable must contain '=': %s", env)
+	}
+
+	parts := strings.SplitN(env, "=", 2)
+	return parts[0], parts[1], nil
```
cyphar: Done. PTAL.

cmd/oci-runtime-tool/generate.go
```diff
@@ -740,6 +744,21 @@ func readKVStrings(files []string, override []string) ([]string, error) {
 	return envVariables, nil
 }
 
+// parseEnv splits a given environment variable (of the form name=value) into
+// (name, value). An error is returned if there is no "=" in the line.
```
Mashimiao: I think comment here is not suitable,  also need to be updated. But it's tiny problem.

cyphar: Fix'd.

wking:
e4317462 looks good to me.

Mashimiao:
LGTM

Mashimiao:
@liangchenye  @mrunalp  PTAL

hqhq:
LGTM

PR #294, opened 2016-12-21
generate: fix removing namespace bug

Mashimiao:
When there are duplicated namespace items, we should remove all of them.

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -906,12 +906,18 @@ func (g *Generator) RemoveLinuxNamespace(ns string) error {
 	if g.spec == nil || g.spec.Linux == nil {
 		return nil
 	}
+	index := 0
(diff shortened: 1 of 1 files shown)


wking:
On Tue, Dec 20, 2016 at 07:55:21PM -0800, Ma Shimiao wrote:
> When there are duplicated namespace items, we should remove all of
> them.

This is not a legal state (opencontainers/runtime-spec#597).  I think
a more robust approach is to fix the erroneous code that produces such
entries (which may be living outside this repository).  To protect
against buggy templates, I'm fine supporting a
--linux--namespace-dedup option (or some such), but I don't think we
want to tie it to RemoveLinuxNamespace.

Mashimiao:
Currently, prefer to give users warning by #390 to guarantee valid template. If we have valid template, this fix will not needed any more.

Issue #16, opened 2016-01-25
Reference spec when reporting violations


wking:
Validators (e.g. the [Nu Html Checker](https://validator.w3.org/)) often report errors with references to the specification requirement that was violated.  To get a good user experience and protect against checking unspecified behavior, I think ocitools checks should be setup to link to the relevant spec wording for any check they perform.  For example, a bundle with non-UTF-8 JSON would fail with a reference to [this](https://github.com/opencontainers/specs/blob/master/glossary.md#json).  If this sounds like a reasonable policy, we should write it up in the README or a CONTRIBUTING.md and open issues/PRs for adjusting any existing checks.

mrunalp:
@wking This is a good idea but maybe hard to maintain the links all the time. How about we target this for 1.0?

wking:
On Mon, Jan 25, 2016 at 10:19:31AM -0800, Mrunal Patel wrote:
> @wking This is a good idea but maybe hard to maintain the links all
> the time. How about we target this for 1.0?

There aren't that many releases left between now and 1.0, so I'd just
start now.  If we link to a particular version of the spec, it will
(a) help with stability and (b) mark stale checks that should be
audited updated after each spec release.

wking:
On Mon, Jan 25, 2016 at 10:19:31AM -0800, Mrunal Patel wrote:
> @wking This is a good idea but maybe hard to maintain the links all
> the time. How about we target this for 1.0?

We're into 1.0.0-rc1, so I'd guess it's time to do this?  I can start
filing PRs if we want to green-light the general idea.

wking:
With #354 merged yesterday, we now have a way to do this.  We still need to transition a lot of code to the new error model, but it's no longer blocking on designing the approach.

PR #295, opened 2016-12-22
validate: add namespace if is duplicated check

Mashimiao:

### DIFF
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -309,31 +309,38 @@ func (v *Validator) CheckMounts() (msgs []string) {
 func (v *Validator) CheckLinux() (msgs []string) {
 	logrus.Debugf("check linux")
 
-	utsExists := false
-	ipcExists := false
-	mountExists := false
-	netExists := false
-	userExists := false
-
-	for index := 0; index < len(v.spec.Linux.Namespaces); index++ {
-		if !namespaceValid(v.spec.Linux.Namespaces[index]) {
-			msgs = append(msgs, fmt.Sprintf("namespace %v is invalid.", v.spec.Linux.Namespaces[index]))
-		} else if len(v.spec.Linux.Namespaces[index].Path) == 0 {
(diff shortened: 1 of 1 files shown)


validate/validate.go
```diff
@@ -309,31 +309,40 @@ func (v *Validator) CheckMounts() (msgs []string) {
...
-				netExists = true
-			} else if v.spec.Linux.Namespaces[index].Type == rspec.MountNamespace {
-				mountExists = true
-			} else if v.spec.Linux.Namespaces[index].Type == rspec.UserNamespace {
-				userExists = true
-			}
 		}
+
+		nsType := v.spec.Linux.Namespaces[index].Type
+		nsPath := v.spec.Linux.Namespaces[index].Path
```
wking: nit: I'd rather have `ns := v.spec.Linux.Namespaces[index]`.  Only one variable (vs. the two here), and you only need an extra character when you use it (e.g. `ns.Type` vs your `nsType`).

wking:
de21837a looks good to me, although I'd slightly prefer [a single convenience variable][1].

Do we want to link [here][2] from the error message?  Or are we punting on that until we're ready for some future push on #16?

[1]: https://github.com/opencontainers/runtime-tools/pull/295#discussion_r93669755
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc3/config-linux.md#namespaces

validate/validate.go
Mashimiao: fixed.

Mashimiao:
@wking  I think we'd better don't link from error message in this PR. If we want to add links, we should find out all error message related with spec. part of messages have link, part of not, that does not look unified.  And maybe adding links will cause maintaining problem, that needs more discussion

Mashimiao:
@hqhq  @mrunalp  @liangchenye  PTAL

Mashimiao:

liangchenye:
LGTM

mrunalp:
LGTM

PR #297, opened 2016-12-30
validate: enhance linux devices validation

Mashimiao:
duplicated device path is invalid
duplicated major:minor is not recommended

### DIFF
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -386,9 +386,47 @@ func (v *Validator) CheckLinux() (msgs []string) {
 		msgs = append(msgs, fmt.Sprintf("On Linux, hostname requires a new UTS namespace to be specified as well"))
 	}
 
+	// Linux devices validation
+	devList := make(map[string]bool)
+	typeList := make(map[string]bool)
 	for index := 0; index < len(v.spec.Linux.Devices); index++ {
-		if !deviceValid(v.spec.Linux.Devices[index]) {
-			msgs = append(msgs, fmt.Sprintf("device %v is invalid.", v.spec.Linux.Devices[index]))
+		device := v.spec.Linux.Devices[index]
+		if !deviceValid(device) {
+			msgs = append(msgs, fmt.Sprintf("device %v is invalid.", device))
+		}
+
+		if _, exists := devList[device.Path]; exists {
+			msgs = append(msgs, fmt.Sprintf("device %s is duplicated", device.Path))
+		} else {
+			var rootfsPath string
+			if filepath.IsAbs(v.spec.Root.Path) {
+				rootfsPath = v.spec.Root.Path
+			} else {
+				rootfsPath = filepath.Join(v.bundlePath, v.spec.Root.Path)
+			}
+			absPath := filepath.Join(rootfsPath, device.Path)
+			_, err := os.Stat(absPath)
+			if os.IsNotExist(err) {
+				devList[device.Path] = true
+			} else if err != nil {
+				msgs = append(msgs, err.Error())
+			} else {
+				msgs = append(msgs, fmt.Sprintf("%s already exists in filesystem", device.Path))
+			}
+		}
+
+		// unify u->c when comparing, they are synonyms
+		var devID string
+		if device.Type == "u" {
+			devID = fmt.Sprintf("%s:%d:%d", "c", device.Major, device.Minor)
+		} else {
+			devID = fmt.Sprintf("%s:%d:%d", device.Type, device.Major, device.Minor)
+		}
+
+		if _, exists := typeList[devID]; exists {
+			logrus.Warnf("type:%s, major:%d and minor:%d for linux devices is duplicated", device.Type, device.Major, device.Minor)
+		} else {
+			typeList[devID] = true
 		}
 	}
 
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"strings"
+	"syscall"
 	"unicode"
 	"unicode/utf8"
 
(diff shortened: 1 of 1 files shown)


wking:
On Thu, Dec 29, 2016 at 08:09:33PM -0800, Ma Shimiao wrote:
> duplicated device path or major:minor is invalid

Is it?  I don't see any language to that effect in the spec [1].

I think duplicating device paths is redundant, but does not need to be
explicitly forbidden in the spec.

I think duplicating major:minor is questionable (use symlinks?  Fix
the config for whoever is consuming the less canonical path?), but
again, I don't see a need to explicitly forbid it in the spec.

[1]: https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#devices

Mashimiao:
On 01/04/2017 06:43 AM, W. Trevor King wrote:
> Is it? I don't see any language to that effect in the spec [1].

Yes, there is not in the spec.
> 
> I think duplicating device paths is redundant, but does not need to be
> explicitly forbidden in the spec.

But, I think this absolutely need to be forbidden.
Linux will not accept duplicated device path.
If only one can be applied, we should clearly tell user it's invalid.

> 
> I think duplicating major:minor is questionable (use symlinks? Fix
> the config for whoever is consuming the less canonical path?), but

buplicated major:minor is accepted in Linux, but I don't think it
make any sense. Let every device has unique major:minor is reasonable.

> again, I don't see a need to explicitly forbid it in the spec.

I think they are well known in Linux. But if you think it's necessary to
rule clearly in spec, I can submit PR to add it.

mrunalp:
@Mashimiao I think it should be clarified in the spec first.

wking:
On Tue, Jan 03, 2017 at 06:23:47PM -0800, Ma Shimiao wrote:
> Linux will not accept duplicated device path.  If only one can be
> applied, we should clearly tell user it's invalid.

The spec should clarify how runtimes should handle devices where a
file already exists at the target path (in which case mknod(2) would
return EEXIST).

Erroring out is one possibility, in which case we should check for
duplicated device paths (and for particularly enthusiastic validation,
compare the device path with the assembled container filesystem).

Attempting to (recursively?) remove the pre-existing file and then
call mknod again (or attempting a (recursive) remove before the first
mknod attempt) is another, in which case there's no need for
validation to require unique target paths, although warning about them
(“are you sure you want to do this?  It seems pointless”) would still
be nice.

> > I think duplicating major:minor is questionable (use symlinks? Fix
> > the config for whoever is consuming the less canonical path?), but
> 
> buplicated major:minor is accepted in Linux, but I don't think it
> make any sense. Let every device has unique major:minor is
> reasonable.

The kernel allows this, and I don't see a reason to be stricter than
the kernel.  Again, warning users about the possible inefficiency
would be nice, but I don't think we need a hard requirement around
this.

Mashimiao:
Try to fix runtime-spec at [647](https://github.com/opencontainers/runtime-spec/pull/647)

Mashimiao:
@wking  @mrunalp  @hqhq  @liangchenye 
As spec has been fixed, I think we can continue reviewing this PR.

validate/validate.go
```diff
@@ -356,9 +356,24 @@ func (v *Validator) CheckLinux() (msgs []string) {
...
+		if !deviceValid(device) {
+			msgs = append(msgs, fmt.Sprintf("device %v is invalid.", device))
+		}
+		if _, exists := devList[device.Path]; exists {
+			msgs = append(msgs, fmt.Sprintf("device %s is duplicated", device.Path))
+		} else {
+			devList[device.Path] = true
+		}
+		devId := fmt.Sprintf("%d:%d", device.Major, device.Minor)
+		if existId, exists := typeList[device.Type]; exists && existId == devId {
```
wking: `typeList` can also be a `map[string]bool` (used as a set, so the value doesn't matter).  `devId` should be the `typeList` key and contain the type as well (normalizing `u` -> `c`).

Mashimiao:
On 01/13/2017 11:31 AM, W. Trevor King wrote:
> |typeList| can also be a |map[string]bool| (used as a set, so the value doesn't matter). |devId| should be the |typeList| key and contain the type as well (normalizing |u| -> |c|).

Sorry, but what does `normalizing |u| -> |c|` mean?

wking:
`u` and `c` [are synonyms][1].

[1]: http://man7.org/linux/man-pages/man1/mknod.1.html

Mashimiao:
On 01/13/2017 11:31 AM, W. Trevor King wrote:
> |typeList| can also be a |map[string]bool| (used as a set, so the value doesn't matter). |devId| should be the |typeList| key and contain the type as well (normalizing |u| -> |c|).

updated, thanks.

validate/validate.go
```diff
@@ -356,9 +356,29 @@ func (v *Validator) CheckLinux() (msgs []string) {
...
+			devList[device.Path] = true
+		}
+		// unify u->c when comparing, they are synonyms
+		var devId string
+		if device.Type == "u" {
+			devId = fmt.Sprintf("%s:%d:%d", "c", device.Major, device.Minor)
+		} else {
+			devId = fmt.Sprintf("%s:%d:%d", device.Type, device.Major, device.Minor)
+		}
+		if _, exists := devList[devId]; exists {
```
wking: Now you're using `devList` with two types of keys: `device.Path` and `devId`.  While the risk of collision is low, I'd rather have two separate maps (`devPaths` and `devIds`?) just to be safe.

Mashimiao:
On 01/13/2017 01:27 PM, W. Trevor King wrote:
> Now you're using |devList| with two types of keys: |device.Path| and |devId|. While the risk of collision is low, I'd rather have two separate maps (|devPaths| and |devIds|?) just to be safe.

While I think it's impossible, but fine, it's not a big deal.
update.

validate/validate.go
```diff
@@ -356,9 +356,33 @@ func (v *Validator) CheckLinux() (msgs []string) {
...
+		device := v.spec.Linux.Devices[index]
+		if !deviceValid(device) {
+			msgs = append(msgs, fmt.Sprintf("device %v is invalid.", device))
+		}
+
+		if _, exists := devList[device.Path]; exists {
+			msgs = append(msgs, fmt.Sprintf("device %s is duplicated", device.Path))
+		} else {
+			devList[device.Path] = true
+		}
```
hqhq: I don't think it applies to

> If a [file][file.1] already exists at `path` that does not match the requested device, the runtime MUST generate an error.

correctly.

Mashimiao: As duplicated devices make no sense, unmatched device is not allowed, no matter what they are, if the path is duplicated, tell user this is not valid I think it's acceptable

hqhq: What if `path` is not duplicated, but the `path` already exists in container filesystem and does not match the requested device?

Mashimiao: That's really a problem. updated. PTAL

validate/validate.go
```diff
@@ -386,9 +386,47 @@ func (v *Validator) CheckLinux() (msgs []string) {
...
+			}
+			absPath := filepath.Join(rootfsPath, device.Path)
+			_, err := os.Stat(absPath)
+			if os.IsNotExist(err) {
+				devList[device.Path] = true
+			} else if err != nil {
+				msgs = append(msgs, err.Error())
+			} else {
+				msgs = append(msgs, fmt.Sprintf("%s already exists in filesystem", device.Path))
+			}
```
hqhq: > If a [file][file.1] already exists at path that does not match the requested device, the runtime MUST generate an error.

I think it means if the specified `path` exists in container filesystem, and it matches the requested device (same type, major and minor), that should be taken as valid case.

Mashimiao: @wking  @hqhq I think SPEC does not clearly rule Mounts should be applied first or Devices should be applied fist. And from runc's code, Mounts are applied first.
Then, there will be a problem. If we want to create device /dev/test, and though a unmatched /dev/test already exists in filesystem, mount may bind a valid /dev/test as we wanted.
I think we'd better just validate duplicate device problem and skip unmatched device problem
How do you think about it?

hqhq: > I think SPEC does not clearly rule Mounts should be applied first or Devices should be applied fist. And from runc's code, Mounts are applied first.

It's implementation details which spec should not define, as long as we define device path to be path in container, implementations would have to mount first.

> If we want to create device /dev/test, and though a unmatched /dev/test already exists in filesystem, mount may bind a valid /dev/test as we wanted.

Bind mount would cover what already in the filesystem, I think that's intentional, device check doesn't have to care about what mounts have done, but what we have at that moment in the filesystem when we do validation.

Mashimiao:
On 02/13/2017 10:27 AM, Qiang Huang wrote:
> Bind mount would cover what already in the filesystem, I think that's intentional, device check doesn't have to care about what mounts have done, but what we have at that moment in the filesystem when we do validation.

I can't agree with you.
If we don't care what mounts have done, This would give users wrong information.
With mounts, the bundle may can be run well and it's a valid bundle.
But we would say it's invalid. 
I think that's not what we really want.

hqhq:
> With mounts, the bundle may can be run well and it's a valid bundle.
But we would say it's invalid.

We really don't have much to do if mounts are gonna mess container filesystem, or are you suggesting we remove `If a [file][file.1] already exists at path that does not match the requested device, the runtime MUST generate an error.` from spec?

Mashimiao:
On 02/13/2017 09:06 PM, Qiang Huang wrote:
> or are you suggesting we remove |If a [file][file.1] already exists at path that does not match the requested device, the runtime MUST generate an error.| from spec?

That's not what I want to say.
As we can't reach an agreement, I will not validate with Mounts in this PR.
Updated, PTAL

Mashimiao:

validate/validate.go
```diff
@@ -386,9 +387,95 @@ func (v *Validator) CheckLinux() (msgs []string) {
...
+		// unify u->c when comparing, they are synonyms
+		var devID string
+		if device.Type == "u" {
+			devID = fmt.Sprintf("%s:%d:%d", "c", device.Major, device.Minor)
+		} else {
+			devID = fmt.Sprintf("%s:%d:%d", device.Type, device.Major, device.Minor)
+		}
+
+		if _, exists := typeList[devID]; exists {
+			logrus.Warnf("type:%s, major:%d and minor:%d for linux devices is duplicated", device.Type, device.Major, device.Minor)
```
wking: This is a good reason to bring in the RFC 2119 errorrs from #354.  But if you'd rather leave it almone until someone else adds it to the `validate` command, that's fine too.

Mashimiao: many other places need it, let's leave it for a follow-up PR.

liangchenye: Yes, RFC2119 erros could come in the following PRs.

liangchenye:
LGTM

q384566678:
LGTM

PR #300, opened 2017-01-09
Report all errors

mrunalp:
This makes runtime validation report all errors at the end.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -2,6 +2,9 @@
 	"ImportPath": "github.com/opencontainers/runtime-tools",
 	"GoVersion": "go1.4",
 	"GodepVersion": "v75",
+	"Packages": [
+		"./..."
+	],
 	"Deps": [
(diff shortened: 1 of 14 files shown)


mrunalp:
@Mashimiao @wking @hqhq @liangchenye PTAL

wking:
Producing TAP output (e.g. with [1]) gives you output that you can
feed into a number of TAP consumers [2].  Since runtimetest isn't a
library, there's no need for errors that are conveniently handled in
Go, so I'd prefer TAP output.

But 8e366c8 is a reasonably small change (which retooling the existing
checks to output TAP would be a bit more work), so I'm ok with it.
Either way (go-multierror or TAP) would be better than the quick death
we have now.

[1]: https://github.com/mndrix/tap-go
[2]: https://testanything.org/consumers.html

Mashimiao:
LGTM

hqhq:
LGTM

PR #306, opened 2017-01-13
validate:Increase the destination path judgment

q384566678:
According to the [specification](https://github.com/opencontainers/runtime-spec/blame/master/config.md#L49-L50).


mrunalp:
LGTM

Mashimiao:
LGTM

PR #309, opened 2017-01-18
fix bug for --arch and --seccomp-arch in bash completion

Kxuan:
Both `__oci-runtime-tool_complete_seccomp_arches` and `__oci-runtime-tool_complete_arches` just fill `$COMPREPLY` and output nothing.
We should not use their output to reset `$COMPREPLY` in `_oci-runtime-tool_generate`.

This patch prevent the "reset" operation, and fix the bug.


Mashimiao:
LGTM

mrunalp:
LGTM

PR #310, opened 2017-01-18
runtimetest: add verification

q384566678:
[specification](https://github.com/opencontainers/runtime-spec/blame/master/config-linux.md#L580)


Mashimiao:
We don't need this.
That's not what we should check in runtimetest. If they were not absolute path, they would not be applied in runtime.

wking:
On Tue, Jan 17, 2017 at 10:24:50PM -0800, Ma Shimiao wrote:
> We don't need this.

+1

> That's not what we should check in runtimetest. If they were not
> absolute path, they would not be applied in runtime.

And perhaps more importantly, we are *already* checking for these in
the validation code (since #259).

PR #247, opened 2016-10-14
generate: add hugepage limit related option

Mashimiao:
Signed-off-by: Ma Shimiao mashimiao.fnst@cn.fujitsu.com

PR #312, opened 2017-01-19
add linux-hugepage-limit option.

Kxuan:
add a hugepageLimits section in config.json.

Usage
oci-runtime-tool generate --linux-hugepage-limit 2MB:209715200

According to [Huge page limits](https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#huge-page-limits).


Mashimiao:
This is duplicate with #247

Kxuan:
ok

PR #315, opened 2017-01-23
bash-completion: add more architectures

Kxuan:
These architectures are added in "[2d0dbcf: Update runtime-spec dependency to v1.0.0-rc1"](https://github.com/opencontainers/runtime-tools/commit/2d0dbcf085e1e08fd05b16b6efb2b6e9336594e7#diff-ede289fca06f6891ce7eaa945805e086R425).

This patch adds them into bash-completion.

### DIFF
--- a/completions/bash/oci-runtime-tool
+++ b/completions/bash/oci-runtime-tool
@@ -144,17 +144,22 @@ __oci-runtime-tool_complete_arches() {
 # context of oci-runtime-tool containers.
 __oci-runtime-tool_complete_seccomp_arches() {
 	COMPREPLY=( $( compgen -W "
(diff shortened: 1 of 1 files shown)


Mashimiao:
Well, this is not a correct fix. 
--seccomp-arch  does not accept value as `SCMP_ARCH_*` any more.
Please fix this problem refer to fucntion `parseArch()` in file `generate/seccomp/parse_architecture.go`

Kxuan:
@Mashimiao Sorry for the mistake.
I updated the patch, and now it works correctly.

Mashimiao:
LGTM

mrunalp:
LGTM

PR #316, opened 2017-01-24
validate: add validation for OOMScoreAdj

Kxuan:
The acceptable values range from -1000 to +1000.

According to the description of "oom_score_adj" (https://www.kernel.org/doc/Documentation/filesystems/proc.txt)

### DIFF
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -417,6 +417,9 @@ func (v *Validator) CheckLinuxResources() (msgs []string) {
 		}
 	}
 
+	if r.OOMScoreAdj != nil && (*r.OOMScoreAdj < -1000 || *r.OOMScoreAdj > 1000) {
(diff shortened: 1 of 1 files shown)


Mashimiao:
I think this also is not needed, because we plan to use schema check value type and limit.
Please refer to https://github.com/opencontainers/runtime-spec/tree/master/schema

Kxuan:
Ok,I found it in the schema.
So is the validation deprecated?

Mashimiao:
No, schema and validation focus on different aspects

wking:
On Tue, Jan 24, 2017 at 07:57:04AM -0800, Ma Shimiao wrote:
> No, schema and validation focus on different aspects 

I don't think that's quite right, and would rather say JSON Schema
validation is a subset of full validation.  See the still-in-flight
#197 for applying this approach.

PR #319, opened 2017-02-07
generate.go: increase the success tips

q384566678:


Mashimiao:
I think we may don't need this, it's not necessary.
If generate suscceed, we can get the config's output.
And currently we can use the output with pipe, this change will break that.

PR #322, opened 2017-02-10
makefile: clean test

q384566678:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ clean:
 $(RUNTIME_TOOLS_LINK):
 	ln -sf $(CURDIR) $(RUNTIME_TOOLS_LINK)
 
-.PHONY: test .gofmt .govet .golint
+.PHONY: test
 
 test: .gofmt .govet .golint
 


Makefile
```diff
@@ -35,7 +35,7 @@ clean:
 $(RUNTIME_TOOLS_LINK):
 	ln -sf $(CURDIR) $(RUNTIME_TOOLS_LINK)
 
-.PHONY: test .gofmt .govet .golint
```
Mashimiao: Well, I think this is not a bug or wrong content. 
Putting them into .PHONY is a way to defend in case of there is `.gofmt`, `.govet` and so son which will break make process.
As there is not any one in project directory, removing them or not is got a big deal.
But noting `fix` in comment is not suitable, `cleanup` will be better.

q384566678: Because both cases can be implemented, and I did not find an accurate description of which is wrong. Just a personal feeling that would be better.

wking: > Putting them into .PHONY is a way to defend in case of there is .gofmt, .govet and so son which will break make process.

And it's also a way of saying “these target names are just for convenience.  We are not attempting to create files called `.gofmt`, etc.”.  I'd rather leave them `.PHONY` so folks don't get confused (“Why is this recipe not creating a `.gofmt` file?”).

Mashimiao:
@mrunalp  @hqhq  @liangchenye  How you think about it?

mrunalp:
Let's leave those targets as PHONY.

Mashimiao:
@q384566678  almost reviewers think we should keep it. So close it .

PR #325, opened 2017-02-14
add device-remove and device-remove-all option

q384566678:
Signed-off-by: zhouhao zhouhao@cn.fujitsu.com

### DIFF
--- a/cmd/oci-runtime-tool/generate.go
+++ b/cmd/oci-runtime-tool/generate.go
@@ -28,6 +28,7 @@ var generateFlags = []cli.Flag{
 	cli.StringFlag{Name: "cwd", Value: "/", Usage: "current working directory for the process"},
 	cli.StringSliceFlag{Name: "device-add", Usage: "add a device which must be made available in the container"},
 	cli.StringSliceFlag{Name: "device-remove", Usage: "remove a device which must be made available in the container"},
+	cli.BoolFlag{Name: "device-remove-all", Usage: "remove all devices which must be made available in the container"},
 	cli.BoolFlag{Name: "disable-oom-kill", Usage: "disable OOM Killer"},
 	cli.StringSliceFlag{Name: "env", Usage: "add environment variable e.g. key=value"},
 	cli.StringSliceFlag{Name: "env-file", Usage: "read in a file of environment variables"},
@@ -524,6 +525,10 @@ func setupSpec(g *generate.Generator, context *cli.Context) error {
 		}
 	}
 
+	if context.Bool("device-remove-all") {
+		g.ClearLinuxDevices()
+	}
+
 	err := addSeccomp(context, g)
 	return err
 }
@@ -655,8 +660,8 @@ var deviceType = map[string]bool{
 	"p": true, // a FIFO
 }
 
-// parseDevice takes the raw string passed with the --device flag
-func parseDevice(device string, g *generate.Generator) (rspec.Device, error) {
+// parseDevice takes the raw string passed with the --device-add flag
+func parseDevice(device string, g *generate.Generator) (rspec.LinuxDevice, error) {
 	dev := rspec.LinuxDevice{}
 
 	// The required part and optional part are separated by ":"
--- a/cmd/oci-runtime-tool/generate.go
+++ b/cmd/oci-runtime-tool/generate.go
@@ -27,6 +27,7 @@ var generateFlags = []cli.Flag{
 	cli.StringFlag{Name: "cgroups-path", Usage: "specify the path to the cgroups"},
 	cli.StringFlag{Name: "cwd", Value: "/", Usage: "current working directory for the process"},
 	cli.StringSliceFlag{Name: "device-add", Usage: "add a device which must be made available in the container"},
+	cli.StringSliceFlag{Name: "device-remove", Usage: "remove a device which must be made available in the container"},
 	cli.BoolFlag{Name: "disable-oom-kill", Usage: "disable OOM Killer"},
 	cli.StringSliceFlag{Name: "env", Usage: "add environment variable e.g. key=value"},
 	cli.StringSliceFlag{Name: "env-file", Usage: "read in a file of environment variables"},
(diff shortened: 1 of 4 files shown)


man/oci-runtime-tool-generate.1.md
```diff
@@ -60,6 +60,14 @@ read the configuration from `config.json`.
   The *fileMode*, *uid*, *gid* are optional.
     *fileMode* is the file mode of the device file.
     *uid*/*gid* is the user/group id of the device file.
+  This option can be specified multiple times.
+
+**--device-remove**=*TYPE:MAJOR:MINOR:PATH[:OPTIONS...]*
```
mrunalp: type:major:minor:path should be enough to identify the device. There is not need to parse the options.

q384566678: updated

q384566678: I think it will be better to add, if only these optional fields are not the same, then it will delete the wrong device?

q384566678: I use the whole structure of the comparison, so I think every field is needed.

generate/generate.go
```diff
@@ -142,7 +142,14 @@ func New() Generator {
 					Type: "mount",
 				},
 			},
-			Devices: []rspec.Device{},
+			Devices: []rspec.Device{
+				{
+					Path:  "/dev/fuse",
```
wking: `New()` is supposed to generate a config that the runtime-tools maintainers feel is a good default (without knowing anything about the container itself).  I don't think that config should contain`/dev/fuse` unless the user explicitly asks for it.

q384566678: Here I want to add a default value to the devices, so that you can better achieve devices-remove and other functions.
You mean to give the assignment value is not good, or value `/dev/fuse` is not good?

wking: There's no need to add a value to the default config just so you can test device removal.  Test that by using a template config (although we don't have a test suite for the generator at the moment, see #180 for a stub).  But the criterion for adding an entry to `New()` should be “is this a useful part of a generic config?”, and I don't think `/dev/fuse` passes that bar.

man/oci-runtime-tool-generate.1.md
q384566678: I added this[commit](https://github.com/opencontainers/runtime-tools/pull/325/commits/87bdf5c01e4abda795f240d045740c02327119c0?diff=unified) to fix the problem.

cmd/oci-runtime-tool/generate.go
```diff
@@ -26,7 +26,9 @@ var generateFlags = []cli.Flag{
 	cli.StringSliceFlag{Name: "cap-drop", Usage: "drop Linux capabilities"},
 	cli.StringFlag{Name: "cgroups-path", Usage: "specify the path to the cgroups"},
 	cli.StringFlag{Name: "cwd", Value: "/", Usage: "current working directory for the process"},
-	cli.StringSliceFlag{Name: "device", Usage: "specifies a device which must be made available in the container"},
+	cli.StringSliceFlag{Name: "device-add", Usage: "add a device which must be made available in the container"},
+	cli.StringSliceFlag{Name: "device-remove", Usage: "remove a device which must be made available in the container"},
+	cli.StringSliceFlag{Name: "device-remove-all", Usage: "remove all devices which must be made available in the container"},
```
Mashimiao: should be cli.BoolFlag

cmd/oci-runtime-tool/generate.go
```diff
@@ -502,15 +504,31 @@ func setupSpec(g *generate.Generator, context *cli.Context) error {
...
+		devices := context.StringSlice("device-remove")
+		for _, device := range devices {
+			err := g.RemoveDevice(device)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	if context.IsSet("device-remove-all") {
```
Mashimiao: change to `context.Bool("device-remove-all")`

cmd/oci-runtime-tool/generate.go
```diff
@@ -642,33 +660,33 @@ var deviceType = map[string]bool{
 	"p": true, // a FIFO
 }
 
-// addDevice takes the raw string passed with the --device flag, parses it, and add it
-func addDevice(device string, g *generate.Generator) error {
+// parseDevice takes the raw string passed with the --device flag
```
q384566678: updated

Mashimiao:
generate.go needs `gofmt`

generate/generate.go
```diff
@@ -915,21 +915,46 @@ func (g *Generator) RemoveLinuxNamespace(ns string) error {
...
-		FileMode: fileMode,
-		UID:      uid,
-		GID:      gid,
+	for i, dev := range g.spec.Linux.Devices {
+		if dev.Path == device.Path {
+			g.spec.Linux.Devices[i] = device
+			return
+		}
+		if dev.Type == device.Type && dev.Major == device.Major && dev.Minor == device.Minor {
+			fmt.Println("WARNING: The same type, major and minor should not be used for multiple devices.")
```
Mashimiao: I'm afraid this will break default output. Maybe fmt.Fprintln(os.Stderr, "") will be better

q384566678: updated, PTAL

Mashimiao:
LGTM

q384566678:
@mrunalp @liangchenye PTAL

q384566678:
@mrunalp @vbatts @hqhq @crosbymichael PTAL

q384566678:
@vbatts @mrunalp @liangchenye @Mashimiao PTAL

liangchenye:
LGTM

liangchenye:
https://github.com/opencontainers/runtime-tools/issues/343

I agree with @Kxuan , we need to have clear 'prefix' of 'Linux*' config/functions.
But I think we can do it after this commit and within one PR.

Mashimiao:
LGTM

PR #326, opened 2017-02-17
Add runtime-validate command to oci-runtime-tool

mrunalp:
This is a conversion of the script to golang. There are a couple of TODOs which will be implemented in follow on PRs. 
The intent here is to move away from the script to go and then start building on top of that.


cmd/oci-runtime-tool/runtime_validate.go
```diff
@@ -0,0 +1,88 @@
...
+	Before: before,
+	Action: func(context *cli.Context) error {
+		return runtimeValidate(context.String("runtime"))
+	},
+}
+
+func runtimeValidate(runtime string) error {
+	fmt.Println("-----------------------------------------------------------------------------------")
+	fmt.Printf("                  VALIDATING RUNTIME: %q\n", runtime)
+	fmt.Println("-----------------------------------------------------------------------------------")
```
wking: I don't think this header is really necessary ;).  Certainly not without a way to turn it off.  But if removing it is part of post-PR polishing, then whatever, we can keep it for now.

cmd/oci-runtime-tool/runtime_validate.go
```diff
@@ -0,0 +1,88 @@
...
+	// Untar the root fs
+	untarCmd := exec.Command("tar", "-xf", "rootfs.tar.gz", "-C", bundleDir)
+	output, err := untarCmd.CombinedOutput()
+	if err != nil {
+		fmt.Println(string(output))
+		return err
+	}
+
+	// Copy the runtimetest binary to the rootfs
+	err = fileutils.CopyFile("runtimetest", filepath.Join(bundleDir, "/runtimetest"))
```
wking: You probably don't need/want the leading slash (that's what `filepath.Join` is for).

wking:
I'm not sure how much you intend to handle with this PR, but you may want a `Makefile` rule to compile this.  And if the goal is to make it installable, it will have to compile-in (or install?) a copy of `runtimetest` and not rely on its being in the current working directory.

cmd/oci-runtime-tool/runtime_validate.go
mrunalp: Removed.

cmd/oci-runtime-tool/runtime_validate.go
mrunalp: Fixed.

mrunalp:
@wking Makefile already compiles oci-runtime-tool. This PR adds a new sub-command to it.
We could install the runtimetest, but given that it's a test binary, I am not too excited about that option. We could just make it an option to the command defaulting to looking in the current working directory.

wking:
On Fri, Feb 17, 2017 at 04:11:39PM -0800, Mrunal Patel wrote:
> @wking Makefile already compiles oci-runtime-tool. This PR adds a
> new sub-command to it.

Oh, right.  Too much switching between the *-tools projects :p.

> We could install the runtimetest, but given that it's a test binary,
> I am not too excited about that option.

Yeah, me neither.

> We could just make it an option to the command defaulting to looking
> in the current working directory.

The compiled-in approach would be to use http.FileSystem or similar
(which would certainly be a lot easier here than in image-spec,
because all of that could be setup at build time without checking the
compiled fs.go into version control).  I guess I like this approach
best if we're not comfortable installing runtimetest.

Short of that, an --runtimetest option (or whatever) with the same
semantics as your current --runtime option would work.  Defaulting to
filepath.Join(".", "runtimetest") and dropping the value string inno
exec.LookPath would give you the path to copy (or an error which we
could die on).  And while this would work, it's not a great user
experience to have to keep track of the location of an uninstalled
runtimetest.

Mashimiao:
Though --runtimetest option is not the best way, I think we can fix it later. I hope this can be landed as soon as possible.

liangchenye:
I agree with this. We can merge this and implement the remain 'TODO' later.
'runtime-tool' is more like a framework. It is hard to be done in one PR.

@Mashimiao @hqhq

liangchenye:
LGTM

Mashimiao:
LGTM

hqhq:
LGTM

PR #328, opened 2017-02-22
validate: completion capabilities

q384566678:
From [here](http://man7.org/linux/man-pages/man7/capabilities.7.html) we can see that the valid value of capabilitie is not written completely.


Mashimiao:
capabilities check has been fixed in #220

wking:
On Wed, Feb 22, 2017 at 01:27:06AM -0800, Ma Shimiao wrote:
> capabilities check has been fixed in #220 

I think we should close this PR (with its locally-hard-coded approach)
in favor of #220 (with its approach based on gocapability).  Would be
nice to get #220 (stable since October?) landed in the next few months
;).

Mashimiao:
I also hope #220  can be landed as soon as possible

PR #329, opened 2017-02-25
Use current process pid instead of assuming pid 1

vishvananda:
A process will not always be pid 1  if we are not unsharing pid namespace or we are entering an existing one.


Mashimiao:
LGTM

mrunalp:
LGTM

PR #321, opened 2017-02-10
docs/command-line-interface: Add Runtime CLI Spec

wking:
Transferred from opencontainers/runtime-spec#513 [as discussed in yesterday's meeting][1].  @mrunalp gave me some pointers on how he wanted this landed [here][2] and [here][3].

I'm not squashing this down into a single commit, because it's a lot easier for me to manage revisions if I keep a branch in the [stand-alone repository][4].  I would encourage reviewers (CC @RobDolinMS and @stevvooe, who expressed an interest in reviewing this stuff) to file issues/PRs against that repository so this PR doesn't get bogged down into the same morass as the runtime-spec PR; this PR would get updated to pull those in as I land the sub-PRs in the stand-alone repo.  @mrunalp thinks we might end up [squashing this just before merging][5], and how that goes is clearly up to the runtime-tools maintainers, but squashing (if it happens at all) should be after substantive review settles.

The WIP tag in the summary is because I'm still not sure where runtime-tools maintainers want [the JSON Schema][6] from the runtime-spec PR.

@mrunalp was also [open to some compliance language][7], but I haven't figured that out yet because the pattern [set in RFC 2616][8] (which runtime-spec followed) wasn't even repeated by [RFC 7230][9] (which obsoleted RFC 2616).  I need some time to compare the differences in the RFC wording before I can float appropriate wording for the command line API.

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2017/opencontainers.2017-02-08-22.00.log.html#l-125
[2]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2017-02-09.log.html#t2017-02-09T22:21:59
[3]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2017-02-10.log.html#t2017-02-10T00:02:44
[4]: https://github.com/wking/oci-command-line-api
[5]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2017-02-10.log.html#t2017-02-10T00:18:53
[6]: https://github.com/opencontainers/runtime-spec/pull/513/files#diff-751d571a4cba0b0155c3d3bf793e08e8R13
[7]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2017-02-10.log.html#t2017-02-10T00:18:25
[...]

PR #331, opened 2017-03-02
Bump version to 0.0.2

rhatdan:

### DIFF
--- a/cmd/oci-runtime-tool/main.go
+++ b/cmd/oci-runtime-tool/main.go
@@ -10,7 +10,7 @@ import (
 func main() {
 	app := cli.NewApp()
 	app.Name = "oci-runtime-tool"
-	app.Version = "0.0.1"
+	app.Version = "0.0.2"
 	app.Usage = "OCI (Open Container Initiative) runtime tools"
 	app.Flags = []cli.Flag{
 		cli.BoolFlag{


wking:
On Thu, Mar 02, 2017 at 01:45:15PM -0800, Daniel J Walsh wrote:
>   * Bump version to 0.0.2

We haven't actually tagged a release yet.  Is there a purpose to
bumping the runtime-tools version without cutting releases?

rhatdan:
I am fine with what ever procedure we can to update the release number.  The bottom line is this package is in Fedora with a very old version, I want to get it updated.

wking:
On Thu, Mar 02, 2017 at 02:43:26PM -0800, Daniel J Walsh wrote:
> The bottom line is this package is in Fedora with a very old
> version, I want to get it updated.

To me that sounds like “we need to start cutting runtime-tools
releases”, and I think that would be a great idea.

rhatdan:
:+1:

mrunalp:
Let's update with the runtime-spec rc5 and then get 1.0 rc for runtime-tools.

wking:
On Tue, Mar 07, 2017 at 11:42:28AM -0800, Mrunal Patel wrote:
> Let's update with the runtime-spec rc5 and then get 1.0 rc for
> runtime-tools.

Before we jump straight to a 1.0 RC, I think we should land #321.  I
think that is only blocked on “where runtime-tools maintainers want
the JSON Schema from the runtime-spec PR” (link to that JSON Schema in
[1]).  Until we document the API we're testing (and ideally, update
our implementation to use that API, e.g. splitting create/start in
[2]), I recommend sticking with more modest release numbers like
0.1.0, 0.2.0, etc.  Hopefully that will avoid the ¾-year period in 1.0
“RCs” which runtime-spec has gone through.

[1]: https://github.com/opencontainers/runtime-tools/pull/321#issue-206681629
[2]: https://github.com/opencontainers/runtime-tools/blob/8392e59c7c11b146462e20701c912242f83cf110/cmd/oci-runtime-tool/runtime_validate.go#L74

PR #332, opened 2017-03-03
validate.go add checkos

q384566678:
I think we should increase the judgment of several OS.
https://github.com/opencontainers/runtime-spec/blame/master/config.md#L324

### DIFF
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -103,6 +103,7 @@ func (v *Validator) CheckAll() (msgs []string) {
 	msgs = append(msgs, v.CheckMounts()...)
 	msgs = append(msgs, v.CheckPlatform()...)
 	msgs = append(msgs, v.CheckProcess()...)
+	msgs = append(msgs, v.CheckOS()...)
 	msgs = append(msgs, v.CheckLinux()...)
 	msgs = append(msgs, v.CheckHooks()...)
 
@@ -337,6 +338,24 @@ func (v *Validator) CheckMounts() (msgs []string) {
 			if !filepath.IsAbs(mount.Destination) {
 				msgs = append(msgs, fmt.Sprintf("destination %v is not an absolute path", mount.Destination))
 			}
+		}
+	}
+
+	return
+}
+
+func (v *Validator) CheckOS() (msgs []string) {
+	logrus.Debugf("check os")
+
+	if v.spec.Platform.OS != "linux" {
+		if v.spec.Linux != nil {
+			msgs = append(msgs, fmt.Sprintf("'linux' MUST NOT be set when platform.os is %q", v.spec.Platform.OS))
+		}
+	}
+
+	if v.spec.Platform.OS != "solaris" {
+		if v.spec.Solaris != nil {
+			msgs = append(msgs, fmt.Sprintf("'solaris' MUST NOT be set when platform.os is %q", v.spec.Platform.OS))
 		}
 	}
 


q384566678:
I will add the Windows type after updating the runtime-spec version (which I am going to do).

validate/validate.go
```diff
@@ -335,6 +336,24 @@ func (v *Validator) CheckMounts() (msgs []string) {
 	return
 }
 
+func (v *Validator) CheckOS() (msgs []string) {
+	logrus.Debugf("check os")
+
+	if v.spec.Platform.OS != "linux" {
+		if v.spec.Linux != nil {
+			msgs = append(msgs, fmt.Sprintf("When system is not linux, Linux field should be empty"))
```
wking: This is not supported by any spec wording, which [only SHOULDs the *setting* (and says nothing about unsetting) the platform-specific sections based on `platform.os`][1].

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc4/config.md#platform-specific-configuration

q384566678: But spec has changed in [this](https://github.com/opencontainers/runtime-spec/pull/673) PR.

wking: > But spec has changed in [this][1] PR.

Ah, I'd missed that (that PR changed a lot ;).  I'm not sure how the new approach will scale to a [`vm` section][2] if/when VM support lands in the spec, but I guess we'll see.

Anyhow, thanks for re-citing master.  I'm fine with your implementation landing as soon as runtime-tools switches from targeting v1.0.0-rc4 (which has the old language) to rc5 (or whichever spec release we target next which has the new language).

As a minor nit, I'd recommend wording that more closely follows the spec.  Something like:

    msgs = append(msgs, fmt.Sprintf("'linux' MUST NOT be set when platform.os is %q", v.spec.Platform.OS))

[1]: https://github.com/opencontainers/runtime-spec/pull/673
[2]: https://github.com/opencontainers/runtime-spec/pull/405#issuecomment-263702840

q384566678: updated,thanks for your advice.

Mashimiao:
LGTM

validate/validate.go
```diff
@@ -335,6 +336,24 @@ func (v *Validator) CheckMounts() (msgs []string) {
 	return
 }
 
+func (v *Validator) CheckOS() (msgs []string) {
+	logrus.Debugf("check os")
+
+	if v.spec.Platform.OS != "linux" {
+		if v.spec.Linux != nil {
+			msgs = append(msgs, fmt.Sprintf("'linux' MUST NOT set when platform.os is %q", v.spec.Platform.OS))
```
mrunalp: nit: MUST NOT be set

validate/validate.go
```diff
@@ -335,6 +336,24 @@ func (v *Validator) CheckMounts() (msgs []string) {
...
+		}
+	}
+
+	if v.spec.Platform.OS != "solaris" {
+		if v.spec.Solaris != nil {
+			msgs = append(msgs, fmt.Sprintf("'solaris' MUST NOT set when platform.os is %q", v.spec.Platform.OS))
```
mrunalp: Same here.

q384566678: updated, PTAL

Mashimiao:
LGTM

mrunalp:
LGTM

PR #349, opened 2017-03-21
update runtime-spec version

q384566678:
Update the `runtime-spec` version, some corresponding change questionable, what questions please suggest.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -1,7 +1,6 @@
 {
 	"ImportPath": "github.com/opencontainers/runtime-tools",
 	"GoVersion": "go1.4",
-	"GodepVersion": "v75",
 	"Packages": [
 		"./..."
 	],
@@ -34,8 +33,8 @@
 		},
 		{
 			"ImportPath": "github.com/opencontainers/runtime-spec/specs-go",
-			"Comment": "v1.0.0-rc1-31-gbb6925e",
-			"Rev": "bb6925ea99f0e366a3f7d1c975f6577475ca25f0"
+			"Comment": "v1.0.0-rc5",
+			"Rev": "035da1dca3dfbb00d752eb58b0b158d6129f3776"
 		},
 		{
 			"ImportPath": "github.com/satori/go.uuid",
@@ -44,7 +43,7 @@
 		},
 		{
 			"ImportPath": "github.com/syndtr/gocapability/capability",
-			"Rev": "2c00daeb6c3b45114c80ac44119e7b8801fdd852"
+			"Rev": "e7cb7fa329f456b3855136a2642b197bad7366ba"
 		},
 		{
 			"ImportPath": "github.com/urfave/cli",
--- a/cmd/oci-runtime-tool/generate.go
+++ b/cmd/oci-runtime-tool/generate.go
@@ -52,8 +52,8 @@ var generateFlags = []cli.Flag{
 	cli.IntFlag{Name: "linux-network-classid", Usage: "specifies class identifier tagged by container's network packets"},
 	cli.StringSliceFlag{Name: "linux-network-priorities", Usage: "specifies priorities of network traffic"},
 	cli.Int64Flag{Name: "linux-pids-limit", Usage: "maximum number of PIDs"},
-	cli.Uint64Flag{Name: "linux-realtime-period", Usage: "CPU period to be used for realtime scheduling (in usecs)"},
-	cli.Uint64Flag{Name: "linux-realtime-runtime", Usage: "the time realtime scheduling may use (in usecs)"},
+	cli.Int64Flag{Name: "linux-realtime-period", Usage: "CPU period to be used for realtime scheduling (in usecs)"},
+	cli.Int64Flag{Name: "linux-realtime-runtime", Usage: "the time realtime scheduling may use (in usecs)"},
 	cli.StringSliceFlag{Name: "masked-paths", Usage: "specifies paths can not be read inside container"},
 	cli.StringFlag{Name: "mount-cgroups", Value: "no", Usage: "mount cgroups (rw,ro,no)"},
 	cli.StringFlag{Name: "mount-label", Usage: "selinux mount context label"},
@@ -385,11 +385,11 @@ func setupSpec(g *generate.Generator, context *cli.Context) error {
 	}
 
 	if context.IsSet("linux-cpu-quota") {
-		g.SetLinuxResourcesCPUQuota(context.Uint64("linux-cpu-quota"))
+		g.SetLinuxResourcesCPUQuota(context.Int64("linux-cpu-quota"))
 	}
 
 	if context.IsSet("linux-realtime-runtime") {
-		g.SetLinuxResourcesCPURealtimeRuntime(context.Uint64("linux-realtime-runtime"))
+		g.SetLinuxResourcesCPURealtimeRuntime(context.Int64("linux-realtime-runtime"))
 	}
 
 	if context.IsSet("linux-pids-limit") {
@@ -644,7 +644,7 @@ var deviceType = map[string]bool{
 
 // addDevice takes the raw string passed with the --device flag, parses it, and add it
 func addDevice(device string, g *generate.Generator) error {
-	dev := rspec.Device{}
+	dev := rspec.LinuxDevice{}
 
 	// The required part and optional part are separated by ":"
 	argsParts := strings.Split(device, ":")
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -165,9 +165,25 @@ func validateCapabilities(spec *rspec.Spec) error {
 		return err
 	}
 
-	expectedCaps := make(map[string]bool)
-	for _, ec := range spec.Process.Capabilities {
-		expectedCaps[ec] = true
+	expectedCaps1 := make(map[string]bool)
+	expectedCaps2 := make(map[string]bool)
+	expectedCaps3 := make(map[string]bool)
+	expectedCaps4 := make(map[string]bool)
+	expectedCaps5 := make(map[string]bool)
+	for _, ec := range spec.Process.Capabilities.Bounding {
+		expectedCaps1[ec] = true
+	}
+	for _, ec := range spec.Process.Capabilities.Effective {
+		expectedCaps2[ec] = true
+	}
+	for _, ec := range spec.Process.Capabilities.Inheritable {
+		expectedCaps3[ec] = true
+	}
+	for _, ec := range spec.Process.Capabilities.Permitted {
+		expectedCaps4[ec] = true
+	}
+	for _, ec := range spec.Process.Capabilities.Ambient {
+		expectedCaps5[ec] = true
 	}
 
 	for _, cap := range capability.List() {
@@ -176,13 +192,45 @@ func validateCapabilities(spec *rspec.Spec) error {
 		}
 
 		capKey := fmt.Sprintf("CAP_%s", strings.ToUpper(cap.String()))
-		expectedSet := expectedCaps[capKey]
-		actuallySet := processCaps.Get(capability.EFFECTIVE, cap)
+		expectedSet := expectedCaps1[capKey]
+		actuallySet := processCaps.Get(capability.BOUNDING, cap)
 		if expectedSet != actuallySet {
 			if expectedSet {
-				return fmt.Errorf("Expected Capability %v not set for process", cap.String())
+				return fmt.Errorf("Expected bounding capability %v not set for process", cap.String())
 			}
-			return fmt.Errorf("Unexpected Capability %v set for process", cap.String())
+			return fmt.Errorf("Unexpected bounding capability %v set for process", cap.String())
+		}
+		expectedSet = expectedCaps2[capKey]
+		actuallySet = processCaps.Get(capability.EFFECTIVE, cap)
+		if expectedSet != actuallySet {
+			if expectedSet {
+				return fmt.Errorf("Expected effective capability %v not set for process", cap.String())
+			}
+			return fmt.Errorf("Unexpected effective capability %v set for process", cap.String())
+		}
+		expectedSet = expectedCaps3[capKey]
+		actuallySet = processCaps.Get(capability.INHERITABLE, cap)
+		if expectedSet != actuallySet {
+			if expectedSet {
+				return fmt.Errorf("Expected inheritable capability %v not set for process", cap.String())
+			}
+			return fmt.Errorf("Unexpected inheritable capability %v set for process", cap.String())
+		}
+		expectedSet = expectedCaps4[capKey]
+		actuallySet = processCaps.Get(capability.PERMITTED, cap)
+		if expectedSet != actuallySet {
+			if expectedSet {
+				return fmt.Errorf("Expected permitted capability %v not set for process", cap.String())
+			}
+			return fmt.Errorf("Unexpected permitted capability %v set for process", cap.String())
+		}
+		expectedSet = expectedCaps5[capKey]
+		actuallySet = processCaps.Get(capability.AMBIENT, cap)
+		if expectedSet != actuallySet {
+			if expectedSet {
+				return fmt.Errorf("Expected ambient capability %v not set for process", cap.String())
+			}
+			return fmt.Errorf("Unexpected ambient capability %v set for process", cap.String())
 		}
 	}
 
@@ -423,8 +471,8 @@ func validateOOMScoreAdj(spec *rspec.Spec) error {
 	return nil
 }
 
-func getIDMappings(path string) ([]rspec.IDMapping, error) {
-	var idMaps []rspec.IDMapping
+func getIDMappings(path string) ([]rspec.LinuxIDMapping, error) {
+	var idMaps []rspec.LinuxIDMapping
 	f, err := os.Open(path)
 	if err != nil {
 		return nil, err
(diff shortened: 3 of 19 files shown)


cmd/oci-runtime-tool/generate.go
```diff
@@ -22,8 +22,6 @@ var generateFlags = []cli.Flag{
 	cli.StringFlag{Name: "arch", Value: runtime.GOARCH, Usage: "architecture the container is created for"},
 	cli.StringSliceFlag{Name: "args", Usage: "command to run in the container"},
 	cli.StringSliceFlag{Name: "bind", Usage: "bind mount directories src:dest[:options...]"},
-	cli.StringSliceFlag{Name: "cap-add", Usage: "add Linux capabilities"},
-	cli.StringSliceFlag{Name: "cap-drop", Usage: "drop Linux capabilities"},
```
q384566678: Because the structure of  `capabilities` changes, I do not know how to deal with these two functions, so temporarily delete, there are good ways to solve.

cmd/oci-runtime-tool/generate.go
```diff
@@ -644,9 +624,9 @@ var deviceType = map[string]bool{
 
 // addDevice takes the raw string passed with the --device flag, parses it, and add it
 func addDevice(device string, g *generate.Generator) error {
-	dev := rspec.Device{}
+	dev := rspec.LinuxDevice{}
 
-	// The required part and optional part are separated by ":"
+	// The required part and optional part are seperated by ":"
```
wking: “separated” was right.

cmd/runtimetest/main.go
```diff
@@ -166,7 +166,19 @@ func validateCapabilities(spec *rspec.Spec) error {
 	}
 
 	expectedCaps := make(map[string]bool)
-	for _, ec := range spec.Process.Capabilities {
+	for _, ec := range spec.Process.Capabilities.Bounding {
+		expectedCaps[ec] = true
+	}
+	for _, ec := range spec.Process.Capabilities.Effective {
+		expectedCaps[ec] = true
```
wking: This isn't right.  You'll need different maps for each `spec.Process.Capabilities.*` and then different checks for each map below.  For example, the current `actuallySet := processCaps.Get(capability.EFFECTIVE, cap)` should be compared against an expected value from `spec.Process.Capabilities.Effective`.

generate/generate.go
```diff
@@ -54,23 +54,31 @@ func New() Generator {
...
-				"CAP_NET_BIND_SERVICE",
-				"CAP_SYS_CHROOT",
-				"CAP_KILL",
-				"CAP_AUDIT_WRITE",
+			Capabilities: &rspec.LinuxCapabilities{
+				Bounding: []string{
+					"CAP_AUDIT_WRITE",
+					"CAP_KILL",
+					"CAP_NET_BIND_SERVICE",
+				},
```
wking: What you're moving back towards here is close to what @mrunalp landed way back in b462307c9.  But in 3a2ba611, @mrunalp changed that default to more closely match values extracted from Docker.  In the absence of arguments for or against a given value, I expect we want to leave the values alone (and keep the fuller set from Docker as we move into the bounding/permitted/… capabilities world).

generate/generate.go
```diff
@@ -764,75 +772,35 @@ func (g *Generator) SetupPrivileged(privileged bool) {
...
+		g.spec.Process.Capabilities.Permitted = finalCapList
+		g.spec.Process.Capabilities.Ambient = finalCapList
 		g.spec.Process.SelinuxLabel = ""
 		g.spec.Process.ApparmorProfile = ""
 		g.spec.Linux.Seccomp = nil
 	}
 }
 
-// ClearProcessCapabilities clear g.spec.Process.Capabilities.
-func (g *Generator) ClearProcessCapabilities() {
```
wking: Clearing *all* of `process.capabilities` is still a well-defined operation.  We may need *new* operations to clear individual `process.capabilities.*`, but there's no need to drop this one.

cmd/oci-runtime-tool/generate.go
mrunalp: We can use these flags to add/drop from all the 5 sets. We can add other more specific flags like cap-bounding-add/cap-bounding-drop after this PR. Thanks!

q384566678: updated. In the subsequent PR can continue to improve.

cmd/runtimetest/main.go
```diff
@@ -176,8 +192,40 @@ func validateCapabilities(spec *rspec.Spec) error {
 		}
 
 		capKey := fmt.Sprintf("CAP_%s", strings.ToUpper(cap.String()))
-		expectedSet := expectedCaps[capKey]
-		actuallySet := processCaps.Get(capability.EFFECTIVE, cap)
+		expectedSet := expectedCaps1[capKey]
+		actuallySet := processCaps.Get(capability.BOUNDING, cap)
+		if expectedSet != actuallySet {
+			if expectedSet {
+				return fmt.Errorf("Expected Capability %v not set for process", cap.String())
```
mrunalp: Can we add the type of capability here?
```go
return fmt.Errorf("Expected bouinding capability %v not set for process", cap.String())`
```

cmd/runtimetest/main.go
```diff
@@ -176,8 +192,40 @@ func validateCapabilities(spec *rspec.Spec) error {
...
+			}
+			return fmt.Errorf("Unexpected Capability %v set for process", cap.String())
```
mrunalp: Same here.

cmd/runtimetest/main.go
```diff
@@ -176,8 +192,40 @@ func validateCapabilities(spec *rspec.Spec) error {
...
+		}
+		expectedSet = expectedCaps2[capKey]
+		actuallySet = processCaps.Get(capability.EFFECTIVE, cap)
+		if expectedSet != actuallySet {
+			if expectedSet {
+				return fmt.Errorf("Expected Capability %v not set for process", cap.String())
```
mrunalp: Can we add the type of capability here?
```go
return fmt.Errorf("Expected effective capability %v not set for process", cap.String())`
```

cmd/runtimetest/main.go
```diff
@@ -176,8 +192,40 @@ func validateCapabilities(spec *rspec.Spec) error {
...
+			}
+			return fmt.Errorf("Unexpected Capability %v set for process", cap.String())
```
mrunalp: Same here and for the other capability types below.

generate/seccomp/seccomp_default.go
```diff
@@ -32,829 +32,829 @@ func arches() []rspec.Arch {
...
 
 // DefaultProfile defines the whitelist for the default seccomp profile.
-func DefaultProfile(rs *specs.Spec) *rspec.Seccomp {
+func DefaultProfile(rs *specs.Spec) *rspec.LinuxSeccomp {
 
-	syscalls := []rspec.Syscall{
+	syscalls := []rspec.LinuxSyscall{
 		{
-			Name:   "accept",
+			Names:  []string{"accept"},
```
mrunalp: All of the system calls can now be added to a single Names list here.
For e.g.

```go
Names: []string{
              "accept",
              "accept4",
              "access",
}
```

mrunalp:
@GrantSeltzer Could you review the seccomp related changes?

GrantSeltzer:
Everything appears to be working properly, but there should be capability added to place a comment inside the rules.

generate/seccomp/seccomp_default.go
q384566678: updated, PTAL.

Godeps/Godeps.json
```diff
@@ -34,8 +33,8 @@
 		},
 		{
 			"ImportPath": "github.com/opencontainers/runtime-spec/specs-go",
-			"Comment": "v1.0.0-rc1-31-gbb6925e",
-			"Rev": "bb6925ea99f0e366a3f7d1c975f6577475ca25f0"
+			"Comment": "v1.0.0-rc5-30-g3adac26",
```
wking: Is this intentionally bumping to a commit that is 30 commits after the tagged rc5?  I don't think the `LinuxSyscall.Comment` removal or `Linux.IntelRdt` additions are so pressing that it's worth bringing in a non-release spec commit.

q384566678: Do you mean to only update to `v1.0.0-rc5`? Or like https://github.com/opencontainers/runtime-tools/pull/349/commits/237bca691ca6d2179701322961a67b34873b3ab8?

wking: > Do you mean to only update to `v1.0.0-rc5`?

That's what I mean.  I'd rather avoid:

    $ ./oci-runtime-tool validate
    1 Errors detected:
    internal error: validate currently only handles version 1.0.0-rc5-dev, but the supplied configuration targets 1.0.0-rc5

q384566678: I now use `godep update` to update the `runtime-spec`, after the update directly to the latest version (`v1.0.0-rc5-30`), what method can you only be updated to `1.0.0-rc5`? Thanks.

wking: > … what method can you only be updated to 1.0.0-rc5?

In whichever `GOPATH` you use, go to `src/github.com/opencontainers/runtime-spec`.  Then:

    $ git checkout v1.0.0-rc5

Now, back in runtime-tools:

    $ godep update github.com/opencontainers/runtime-spec/specs-go

and you should be good to go.

q384566678: updated. Thank you very much for your guidance.

vbatts:
LGTM

mrunalp:
There is an issue with generate specifically in the syscall section. 
runc run/ make localvalidation just hang.

cyphar:
This is currently blocking a wide variety of things (like being able to use the newest version of runC with anything generated by this library). So `umoci` can't work with the latest runC at the moment, neither can `cri-o` or `orca`.

Are there any updates?

mrunalp:
I'll look into this by tomorrow if there are no more updates to the PR.

vbatts:
yeah. the generated `config.json` does not currently work with `runc run`
something in capabilities is not marshalling correctly as a `[]string`, and the syscall names in seccomp are marshalling as empty strings.
I wish we could add a round-trip test for this (i realize that it's default to bootstrap that).

wking:
On Thu, Apr 06, 2017 at 11:03:07AM -0700, Vincent Batts wrote:
> I wish we could add a round-trip test for this (i realize that it's
> default to bootstrap that).

It's one step down the stack, but I've started stubbing in round-trip
tests in opencontainers/runtime-spec#759.

generate/seccomp/seccomp_default.go
```diff
@@ -32,1766 +32,482 @@ func arches() []rspec.Arch {
...
+				},
+				{
+					Index: 0,
+					Value: 0x0008,
+					Op:    rspec.OpEqualTo,
+				},
+				{
+					Index: 0,
+					Value: 0xffffffff,
+					Op:    rspec.OpEqualTo,
```
wking: This looks buggy to me (and may be the source of @mrunalp's [problem][1]).  In master, [only `personality` has an `0xffffffff` arg][2], but here you have a whole bunch of names with that arg.  I think you over-flattened here, and need to split this entry into separate entries for each `[]LinuxSeccompArg` value in master.

[1]: https://github.com/opencontainers/runtime-tools/pull/349#issuecomment-291261413
[2]: https://github.com/opencontainers/runtime-tools/blob/133b95395cd847ef951c237784b9863cdf1ed860/generate/seccomp/seccomp_default.go#L877-L885

mrunalp: Adding this diff makes this PR work. 

```
diff --git a/generate/seccomp/seccomp_default.go b/generate/seccomp/seccomp_default.go
index e057ca7..e9561ae 100644
--- a/generate/seccomp/seccomp_default.go
+++ b/generate/seccomp/seccomp_default.go
@@ -344,23 +344,6 @@ func DefaultProfile(rs *specs.Spec) *rspec.LinuxSeccomp {
                                "writev",
                        },
                        Action: rspec.ActAllow,
-                       Args: []rspec.LinuxSeccompArg{
-                               {
-                                       Index: 0,
-                                       Value: 0x0,
-                                       Op:    rspec.OpEqualTo,
-                               },
-                               {
-                                       Index: 0,
-                                       Value: 0x0008,
-                                       Op:    rspec.OpEqualTo,
-                               },
-                               {
-                                       Index: 0,
-                                       Value: 0xffffffff,
-                                       Op:    rspec.OpEqualTo,
-                               },
-                       },
                },
        }
        var sysCloneFlagsIndex uint
```

mrunalp: @q384566678 could you include the diff from my comment above in the PR?

wking: > … could you include the diff from my comment above in the PR?

I think that patch is too coarse.  For example `personality` should probably *keep* [its old entries][1].  The algorithm for the change should be:

> For all the old syscalls, group by `(action, args)` tuples, and create a new entry collecting all the old `Name`s into the new `Names`.

It looks like:

```
      Action: rspec.ActAllow,
      Args:   []rspec.Arg{},
```

matches everything except `personality` (based on a very quick skim, may have missed something).  And `personality` will still need three entries (one for each of its current `[]rspec.Arg` values.

[1]: https://github.com/opencontainers/runtime-tools/blob/133b95395cd847ef951c237784b9863cdf1ed860/generate/seccomp/seccomp_default.go#L855-L886

q384566678: I think  https://github.com/opencontainers/runtime-tools/pull/349/commits/dc38a7ddbecaef0d8529ad6e02d80f5f18573e0c is better.
@mrunalp @wking PTAL

mrunalp: @q384566678 Did you test latest changes with latest runc? (I'll test it tomorrow).

q384566678: @mrunalp I was doing it, but when I executed the `sudo make RUNTIME=runc localvalidation` command, the following error occurred：
> go: GOPATH entry is relative; must be absolute path: "".

I have already set up `GOPATH`, I'm looking for the solution, may want to make trouble you first test, thanks a lot.

mrunalp: @q384566678 I cannot find dc38a7ddbecaef0d8529ad6e02d80f5f18573e0c. Did you forget to push it?

wking: > I cannot find dc38a7d. Did you forget to push it?

The current tip (afc8d3588) includes the change I think we want, and was committed a few hours after @q384566678 mentioned dc38a7d.  I expect he just fixed up some minor thing in the interim, and that afc8d3588 is ready for testing/merging.

mrunalp:
Much better now. runc start fine and tests also run. The one problem I see in the generated config is `chroot` being added to the seccomp section 5 times:
```
                               {
                                        "names": [
                                                "chroot"
                                        ],
                                        "action": "SCMP_ACT_ALLOW",
                                        "args": [],
                                        "comment": ""
                                },
                                {
                                        "names": [
                                                "chroot"
                                        ],
                                        "action": "SCMP_ACT_ALLOW",
                                        "args": [],
                                        "comment": ""
                                },
                                {
                                        "names": [
                                                "chroot"
                                        ],
                                        "action": "SCMP_ACT_ALLOW",
                                        "args": [],
                                        "comment": ""
                                },
                                {
                                        "names": [
                                                "chroot"
                                        ],
                                        "action": "SCMP_ACT_ALLOW",
                                        "args": [],
                                        "comment": ""
                                },
                                {
                                        "names": [
                                                "chroot"
                                        ],
                                        "action": "SCMP_ACT_ALLOW",
                                        "args": [],
                                        "comment": ""
                                },
```

generate/seccomp/seccomp_default.go
```diff
@@ -32,1766 +32,486 @@ func arches() []rspec.Arch {
...
+				},
+			},
 		},
 	}
 	var sysCloneFlagsIndex uint
 
 	capSysAdmin := false
 	var cap string
+	var caps []string
```
mrunalp: The reason for repitition is because we are adding caps multiple times in this list. Practically the first 4 will be same for most containers and ambient could have extra capabilities. So, instead of adding all, can we just add the caps from permitted and ambient to a map and then iterate over the map to add the syscalls?

wking:
On Mon, Apr 10, 2017 at 02:20:04PM -0700, Mrunal Patel wrote:
> The one problem I see in the generated config is `chroot` being
> added to the seccomp section 5 times…

This is a pre-existing bug, because we're appending (e.g. [1]) instead
of using append-if-missing logic.  While it would be good to address
this, I don't think it needs to block this PR (it will work, and just
be less efficient than if we consolidated those duplicates).

[1]: https://github.com/opencontainers/runtime-tools/blob/133b95395cd847ef951c237784b9863cdf1ed860/generate/seccomp/seccomp_default.go#L1685-L1692

generate/seccomp/seccomp_default.go
wking: > So, instead of adding all, can we just add the caps from permitted and ambient to a map…

This is probably the way to go ([although I'm not sure it has to be part of this PR][1]).  However, the map is going to be complicated, because you want to [key by `(LinuxSeccompAction, LinuxSeccompArg)`][2] with values that are `[]*LinuxSyscall` (because the same backing `LinuxSyscall` may show up as the value for multiple keys in the map, and there may be multiple entries for each key).  And then once you find the appropriate group of `LinuxSyscall`s you only need to add the syscall if it's not already listed in `Names` for any `LinuxSyscall`.

[1]: https://github.com/opencontainers/runtime-tools/pull/349#issuecomment-293085845
[2]: https://github.com/opencontainers/runtime-tools/pull/349#discussion_r110487065

mrunalp:
@wking Yeah, we can fix that up in a follow on PR.

mrunalp:
LGTM
cc: @Mashimiao @liangchenye

Mashimiao:
LGTM

cyphar:
:+1:

vbatts:
:100:

PR #350, opened 2017-03-21
validate: increase OS validation for special cases

q384566678:
For some of the [linux system](https://github.com/opencontainers/runtime-spec/blame/master/config.md#L145) under the special value of the verification, should increase the decision on the `os`.

### DIFF
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -273,46 +273,71 @@ func (v *Validator) CheckProcess() (msgs []string) {
 		}
 	}
 
-	var caps []string
+	msgs = append(msgs, v.CheckCapablities()...)
+	msgs = append(msgs, v.CheckRlimits()...)
 
-	for _, cap := range process.Capabilities.Bounding {
-		caps = append(caps, cap)
-	}
-	for _, cap := range process.Capabilities.Effective {
-		caps = append(caps, cap)
-	}
-	for _, cap := range process.Capabilities.Inheritable {
-		caps = append(caps, cap)
-	}
-	for _, cap := range process.Capabilities.Permitted {
-		caps = append(caps, cap)
-	}
-	for _, cap := range process.Capabilities.Ambient {
-		caps = append(caps, cap)
-	}
+	if v.spec.Platform.OS == "linux" {
 
-	for _, capability := range caps {
-		if err := CapValid(capability, v.HostSpecific); err != nil {
-			msgs = append(msgs, fmt.Sprintf("capability %q is not valid, man capabilities(7)", capability))
+		if len(process.ApparmorProfile) > 0 {
+			profilePath := filepath.Join(v.bundlePath, v.spec.Root.Path, "/etc/apparmor.d", process.ApparmorProfile)
+			_, err := os.Stat(profilePath)
+			if err != nil {
+				msgs = append(msgs, err.Error())
+			}
 		}
 	}
 
+	return
+}
+
+func (v *Validator) CheckCapablities() (msgs []string) {
+	process := v.spec.Process
+	if v.spec.Platform.OS == "linux" {
+		var caps []string
+
+		for _, cap := range process.Capabilities.Bounding {
+			caps = append(caps, cap)
+		}
+		for _, cap := range process.Capabilities.Effective {
+			caps = append(caps, cap)
+		}
+		for _, cap := range process.Capabilities.Inheritable {
+			caps = append(caps, cap)
+		}
+		for _, cap := range process.Capabilities.Permitted {
+			caps = append(caps, cap)
+		}
+		for _, cap := range process.Capabilities.Ambient {
+			caps = append(caps, cap)
+		}
+
+		for _, capability := range caps {
+			if err := CapValid(capability, v.HostSpecific); err != nil {
+				msgs = append(msgs, fmt.Sprintf("capability %q is not valid, man capabilities(7)", capability))
+			}
+		}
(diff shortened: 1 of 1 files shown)


validate/validate.go
```diff
@@ -256,28 +256,30 @@ func (v *Validator) CheckProcess() (msgs []string) {
 		}
 	}
 
-	for _, capability := range process.Capabilities {
-		if err := CapValid(capability, v.HostSpecific); err != nil {
-			msgs = append(msgs, fmt.Sprintf("capability %q is not valid, man capabilities(7)", capability))
+	if v.spec.Platform.OS == "linux" {
+		for _, capability := range process.Capabilities {
```
wking: `process.capabilities` is no longer Linux-specific since opencontainers/runtime-spec#673 (which [removed this line][1]).  I'm not clear on how the capabilities check is supposed to work on other OSes, but unless the spec changes I don't think we can have a Linux-only check here.

[1]: https://github.com/opencontainers/runtime-spec/pull/673/files#diff-c9c91c29b41257aea3a3403cc606ad99L132

validate/validate.go
```diff
@@ -256,28 +256,30 @@ func (v *Validator) CheckProcess() (msgs []string) {
...
-	}
 
-	for index, rlimit := range process.Rlimits {
-		if err := rlimitValid(rlimit); err != nil {
-			msgs = append(msgs, err.Error())
-		}
-		for i := index + 1; i < len(process.Rlimits); i++ {
-			if process.Rlimits[index].Type == process.Rlimits[i].Type {
-				msgs = append(msgs, fmt.Sprintf("rlimit can not contain the same type %q.", process.Rlimits[index].Type))
+		for index, rlimit := range process.Rlimits {
```
wking: `process.rlimits` is also no longer Linux-specific since opencontainers/runtie-spec#673, so this shouldn't get moved into a Linux-specific section.

validate/validate.go
q384566678: But the validation of `capabilities` is compared with `CAPABILITIES (7)`,  [CAPABILITIES(7)](http://man7.org/linux/man-pages/man7/capabilities.7.html) is the value under linux.

wking: So we should add checks for capabilities on other OSes.  Or move the Linux-specific cap validation into a `CheckLinuxCapabilities` and raise not-implemented-yet errors if someone tries to validate a non-Linux config.  But silently ignoring `process.capabilities` in non-Linux configs will give users a false sense of security.

q384566678: But it is not clear the value of other systems, it can not be verified.
My idea is to verify the linux system under the capabilities of the special value, not the entire capabilities field validation, so is not the integration of these tests into a function will be better?

wking: > But it is not clear the value of other systems, it can not be verified.

I agree, and think somebody with more weight than me should bring this up among the runtime-spec maintainers before they cut 1.0 (I [tried][1], but didn't have much success).

> My idea is to verify the linux system under the capabilities of the special value, not the entire capabilities field validation…

That's fine, but I don't think we should be silently passing on non-Linux OSes.  Does [moving the Linux-specific cap validation into `CheckLinuxCapabilities` and raising not-implemented-yet errors for non-Linux configs][2] sound like a bad idea to you?

[1]: https://github.com/opencontainers/runtime-spec/pull/673/files#r99353136
[2]: https://github.com/opencontainers/runtime-tools/pull/350#discussion_r107457088

q384566678:
@wking What do you think of https://github.com/opencontainers/runtime-tools/pull/350/commits/15804036a2282a41aaaafde2b9aea8b2af77dde1?

validate/validate.go
```diff
@@ -256,29 +256,52 @@ func (v *Validator) CheckProcess() (msgs []string) {
...
+
+func (v *Validator) CheckCapablities() (msgs []string) {
+	if v.spec.Platform.OS == "linux" {
+		for _, capability := range v.spec.Process.Capabilities {
+			if err := CapValid(capability, v.HostSpecific); err != nil {
+				msgs = append(msgs, fmt.Sprintf("capability %q is not valid, man capabilities(7)", capability))
 			}
 		}
+	} else {
+		logrus.Warnf("OS %q has not yet have a special value for capabilities", v.spec.Platform.OS)
```
wking: Maybe reword to:

> process.capabilities validation not yet implemented for OS %q

q384566678:
@mrunalp @liangchenye @Mashimiao PTAL

validate/validate.go
```diff
@@ -273,47 +273,71 @@ func (v *Validator) CheckProcess() (msgs []string) {
...
+	process := v.spec.Process
+	if v.spec.Platform.OS == "linux" {
+		for index, rlimit := range process.Rlimits {
+			if err := rlimitValid(rlimit); err != nil {
+				msgs = append(msgs, err.Error())
+			}
+			for i := index + 1; i < len(process.Rlimits); i++ {
+				if process.Rlimits[index].Type == process.Rlimits[i].Type {
+					msgs = append(msgs, fmt.Sprintf("rlimit can not contain the same type %q.", process.Rlimits[index].Type))
+				}
```
Mashimiao: duplicate rlimits check should work for all OSs, not just Linux

q384566678: updated, PTAL.

Mashimiao:
LGTM

liangchenye:
LGTM

PR #352, opened 2017-03-27
Add to the runtime-validate option *.1.md file

q384566678:


wking:
>     A man/oci-runtime-tool-runtime-validate.1.md (25)

I think this command is going to be removed, now that #336 has landed.
Perhaps #326, or at least part of it, should be reverted?

Mashimiao:
runtime-validate is removed, we don't need this any more

PR #353, opened 2017-03-28
validate: Test config.json and rootfs sibling-hood

wking:
Test the [spec's][1]:

> While these artifacts MUST all be present in a single directory on the local filesystem, …

which is a condition it imposes on `config.json` and the directory referenced by `root.path`.

I think we should drop that restriction from the spec, but [my attempt to remove the restriction was rejected][2].  If a future spec drops the restriction, we can revert this commit.

Using `path/filepath` for the path manipulation will break when validating cross-platform configs (e.g. trying to validate a Windows bundle on a Linux machine).  But that's a bigger issue than this commit (previous discussion [here][3] and [here][4]), so I've left it alone for now.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/bundle.md#container-format
[2]: https://github.com/opencontainers/runtime-spec/pull/469
[3]: https://github.com/opencontainers/runtime-tools/pull/256#issuecomment-258365559
[4]: https://github.com/opencontainers/runtime-tools/pull/259#discussion_r85171377

### DIFF
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -115,17 +115,34 @@ func (v *Validator) CheckAll() (msgs []string) {
 func (v *Validator) CheckRootfsPath() (msgs []string) {
 	logrus.Debugf("check rootfs path")
 
+	absBundlePath, err := filepath.Abs(v.bundlePath)
+	if err != nil {
+		msgs = append(msgs, fmt.Sprintf("unable to convert %q to an absolute path", v.bundlePath))
+	}
+
 	var rootfsPath string
+	var absRootPath string
 	if filepath.IsAbs(v.spec.Root.Path) {
 		rootfsPath = v.spec.Root.Path
+		absRootPath = filepath.Clean(rootfsPath)
 	} else {
+		var err error
 		rootfsPath = filepath.Join(v.bundlePath, v.spec.Root.Path)
+		absRootPath, err = filepath.Abs(rootfsPath)
+		if err != nil {
+			msgs = append(msgs, fmt.Sprintf("unable to convert %q to an absolute path", rootfsPath))
+		}
 	}
 
 	if fi, err := os.Stat(rootfsPath); err != nil {
 		msgs = append(msgs, fmt.Sprintf("Cannot find the root path %q", rootfsPath))
 	} else if !fi.IsDir() {
 		msgs = append(msgs, fmt.Sprintf("The root path %q is not a directory.", rootfsPath))
+	}
+
+	rootParent := filepath.Dir(absRootPath)
+	if absRootPath == string(filepath.Separator) || rootParent != absBundlePath {
+		msgs = append(msgs, fmt.Sprintf("root.path is %q, but it MUST be a child of %q", v.spec.Root.Path, absBundlePath))
 	}
 
 	return


validate/validate.go
```diff
@@ -128,6 +140,11 @@ func (v *Validator) CheckRootfsPath() (msgs []string) {
 		msgs = append(msgs, fmt.Sprintf("The root path %q is not a directory.", rootfsPath))
 	}
 
+	rootParent := filepath.Dir(absRootPath);
+	if rootParent == string(filepath.Separator) || rootParent != absBundlePath {
```
Mashimiao: You mean rootParent can't be "/"?  It seems there is no such limitation in SPEC.

wking: > You mean rootParent can't be "/"?

That should be fine.  The `Separator` comparison is intended to catch  `absRootPath` being a spelling of `/`, based on [the `Dir` docs][1]:

> The returned path does not end in a separator unless it is the root directory.

But I'll double check `Dir("/") ` later tonight.

[1]: https://golang.org/pkg/path/filepath/#Dir

wking: Ok, should be fixed with 078bb17 → 8d697e8.

validate/validate.go
```diff
@@ -115,19 +115,38 @@ func (v *Validator) CheckAll() (msgs []string) {
 func (v *Validator) CheckRootfsPath() (msgs []string) {
 	logrus.Debugf("check rootfs path")
 
+	absBundlePath, err := filepath.Abs(v.bundlePath)
+	if err != nil {
+		msgs = append(msgs, fmt.Sprintf("unable to convert %q to an absolute path", v.bundlePath))
+	}
+	absBundlePath = filepath.Clean(absBundlePath)
```
Mashimiao: Maybe we don't need this, Abs() calls Clean() on the result.

validate/validate.go
```diff
@@ -115,19 +115,38 @@ func (v *Validator) CheckAll() (msgs []string) {
...
+
 	var rootfsPath string
+	var absRootPath string
 	if filepath.IsAbs(v.spec.Root.Path) {
 		rootfsPath = v.spec.Root.Path
+		absRootPath = rootfsPath
```
Mashimiao: I think adding filepath.Clean() here will be fine

validate/validate.go
```diff
@@ -115,19 +115,38 @@ func (v *Validator) CheckAll() (msgs []string) {
...
 	} else {
+		var err error
 		rootfsPath = filepath.Join(v.bundlePath, v.spec.Root.Path)
+		absRootPath, err = filepath.Abs(rootfsPath)
+		if err != nil {
+			msgs = append(msgs, fmt.Sprintf("unable to convert %q to an absolute path", rootfsPath))
+		}
 	}
+	absRootPath = filepath.Clean(absRootPath)
```
Mashimiao: removing this from here will be fine.

validate/validate.go
wking: Thanks :).  Addressed this (and your other `Abs` / `Clean` points) with 8d697e8 → 899afea.

Mashimiao:
LGTM

liangchenye:
LGTM

liangchenye:
Interesting, CI fails in getting golint. 
I restart the build, it works...

PR #354, opened 2017-03-30
add OCIError to differentiate 'MUST/SHOULD' and add reference support

liangchenye:
this is an implementation of #353 .

PTAL
@wking @Mashimiao @mrunalp

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -32,6 +32,13 @@ const PrGetNoNewPrivs = 39
 const specConfig = "config.json"
 
 var (
+	defaultFS = map[string]string{
+		"/proc":    "proc",
+		"/sys":     "sysfs",
+		"/dev/pts": "devpts",
+		"/dev/shm": "tmpfs",
+	}
+
 	defaultSymlinks = map[string]string{
 		"/dev/fd":     "/proc/self/fd",
 		"/dev/stdin":  "/proc/self/fd/0",
@@ -304,6 +311,28 @@ func validateRootFS(spec *rspec.Spec) error {
 		err := testWriteAccess("/")
 		if err == nil {
 			return fmt.Errorf("Rootfs should be readonly")
+		}
+	}
+
+	return nil
+}
+
+func validateDefaultFS(spec *rspec.Spec) error {
+	logrus.Debugf("validating linux default filesystem")
+
+	mountInfos, err := mount.GetMounts()
+	if err != nil {
+		return ociErr.NewOCIError(ociErr.DefaultFilesystems, err.Error())
+	}
+
+	mountsMap := make(map[string]string)
+	for _, mountInfo := range mountInfos {
+		mountsMap[mountInfo.Mountpoint] = mountInfo.Fstype
+	}
+
+	for fs, fstype := range defaultFS {
+		if !(mountsMap[fs] == fstype) {
+			return ociErr.NewOCIError(ociErr.DefaultFilesystems, fmt.Sprintf("%v must exist and expected type is %v", fs, fstype))
 		}
 	}
 
@@ -616,6 +645,10 @@ func validate(context *cli.Context) error {
 		{
 			test:        validateDefaultSymlinks,
 			description: "default symlinks",
+		},
+		{
+			test:        validateDefaultFS,
+			description: "default file system",
 		},
 		{
 			test:        validateDefaultDevices,
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -18,9 +18,11 @@ import (
 	"github.com/hashicorp/go-multierror"
 	"github.com/mndrix/tap-go"
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
-	"github.com/opencontainers/runtime-tools/cmd/runtimetest/mount"
 	"github.com/syndtr/gocapability/capability"
 	"github.com/urfave/cli"
+
+	"github.com/opencontainers/runtime-tools/cmd/runtimetest/mount"
+	ociErr "github.com/opencontainers/runtime-tools/validate"
 )
 
 // PrGetNoNewPrivs isn't exposed in Golang so we define it ourselves copying the value from
@@ -660,11 +662,16 @@ func validate(context *cli.Context) error {
 	t := tap.New()
 	t.Header(0)
 
+	complianceLevelString := context.String("compliance-level")
+	complianceLevel := ociErr.ParseLevel(complianceLevelString)
 	var validationErrors error
 	for _, v := range defaultValidations {
 		err := v.test(spec)
 		t.Ok(err == nil, v.description)
 		if err != nil {
+			if e, ok := err.(*ociErr.OCIError); ok && e.Level < complianceLevel {
+				continue
+			}
 			validationErrors = multierror.Append(validationErrors, err)
 		}
 	}
@@ -674,6 +681,9 @@ func validate(context *cli.Context) error {
 			err := v.test(spec)
 			t.Ok(err == nil, v.description)
 			if err != nil {
+				if e, ok := err.(*ociErr.OCIError); ok && e.Level < complianceLevel {
+					continue
+				}
 				validationErrors = multierror.Append(validationErrors, err)
 			}
 		}
@@ -699,6 +709,11 @@ func main() {
 			Name:  "path",
 			Value: ".",
 			Usage: "Path to the configuration",
+		},
+		cli.StringFlag{
+			Name:  "compliance-level",
+			Value: "must",
+			Usage: "Compliance level (must or should)",
 		},
 	}
 
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -322,7 +322,7 @@ func validateDefaultFS(spec *rspec.Spec) error {
 
 	mountInfos, err := mount.GetMounts()
 	if err != nil {
-		return ociErr.NewOCIError(ociErr.DefaultFilesystems, err.Error())
+		return ociErr.NewError(ociErr.DefaultFilesystems, err.Error())
 	}
 
 	mountsMap := make(map[string]string)
@@ -332,7 +332,7 @@ func validateDefaultFS(spec *rspec.Spec) error {
 
 	for fs, fstype := range defaultFS {
 		if !(mountsMap[fs] == fstype) {
-			return ociErr.NewOCIError(ociErr.DefaultFilesystems, fmt.Sprintf("%v must exist and expected type is %v", fs, fstype))
+			return ociErr.NewError(ociErr.DefaultFilesystems, fmt.Sprintf("%v must exist and expected type is %v", fs, fstype))
 		}
 	}
 
@@ -702,7 +702,7 @@ func validate(context *cli.Context) error {
 		err := v.test(spec)
 		t.Ok(err == nil, v.description)
 		if err != nil {
-			if e, ok := err.(*ociErr.OCIError); ok && e.Level < complianceLevel {
+			if e, ok := err.(*ociErr.Error); ok && e.Level < complianceLevel {
 				continue
 			}
 			validationErrors = multierror.Append(validationErrors, err)
@@ -714,7 +714,7 @@ func validate(context *cli.Context) error {
 			err := v.test(spec)
 			t.Ok(err == nil, v.description)
 			if err != nil {
-				if e, ok := err.(*ociErr.OCIError); ok && e.Level < complianceLevel {
+				if e, ok := err.(*ociErr.Error); ok && e.Level < complianceLevel {
 					continue
 				}
 				validationErrors = multierror.Append(validationErrors, err)
@@ -746,7 +746,7 @@ func main() {
 		cli.StringFlag{
 			Name:  "compliance-level",
 			Value: "must",
-			Usage: "Compliance level (must or should)",
+			Usage: "Compliance level (may, should or must)",
 		},
 	}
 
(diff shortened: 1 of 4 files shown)


cmd/runtimetest/main.go
```diff
@@ -642,6 +644,12 @@ func validate(context *cli.Context) error {
 			err := v.test(spec)
 			t.Ok(err == nil, v.description)
 			if err != nil {
+				if logLevel >= logrus.ErrorLevel {
+					// Skip 'SHOULD', 'MAY' ...
```
wking: I don't like overloading `logrus.ErrorLevel` to control this, since “SHOULD violations are fatal” is not the same as “I want to see debug logging”.  Can we add a setting to the `Validator` structure to control this for config validation, and a command-line parameter to `runtimetest` to control this for runtime validation?

validate/error.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+// OCIError represents a error with OCI reference
+type OCIError struct {
+	Level     RFC_LEVEL
+	Reference string
+	Err       error
+}
+
+const ReferencePrefix = "https://github.com/opencontainers/runtime-spec/blob/master/"
```
wking: Instead of floating this with `master`, we should be using the `ociVersion` of the config we're checking against.  This is another reason that we [should only be supporting tagged spec releases][1].

[1]: https://github.com/opencontainers/runtime-tools/pull/349#discussion_r108730611

cmd/runtimetest/main.go
liangchenye: Yes, overloading logrus is a little bit tricky.
I'll use another flag like 'rfc-level'.

wking:
The OCIError structure looks reasonable to me.  I think we should use
it for MUST-level violations as well.

cmd/runtimetest/main.go
wking: Maybe `compliance-level` would be better than `rfc-level`?  There are many RFCs, and what we're interested in here is how well the tested-thing complies with the specification.

validate/error.go
liangchenye: The current version is: '-rc5-dev' and we don't have a matched release.
We can change it after v.1.0.0 release, for example: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config-linux.md

liangchenye:
All the 'MUST-level' errors need to be filled. We can add all ErrorCode. It also helps to track #66.

cmd/runtimetest/main.go
liangchenye: I'll update with all your good suggestions tomorrow. It is a little bit late here.

validate/error.go
wking: > The current version is: '-rc5-dev'…

The current runtime-spec version supported by runtime-tools' master [is `-rc1-dev`][1].  If we make [the adjustment I'm asking for][2] in the in-flight #349, we can target `-rc5`.  I see no purpose to targeting `-rc5-dev`. 

[1]: https://github.com/opencontainers/runtime-tools/blob/38e8560d8b49e55d9edc61a3270272a261dae238/Godeps/Godeps.json#L37
[2]: https://github.com/opencontainers/runtime-tools/pull/349#discussion_r108730611

liangchenye:
Updated: 
add a 'compliance-level' option to runtimetest.
add man page and complete bash

The tagged release and other 'MUST' error will be added in the following commits.

PTAL   @wking  @mrunalp @hqhq @Mashimiao

validate/error.go
```diff
@@ -0,0 +1,69 @@
...
+	DefaultFilesystems: OCIError{Level: ComplianceShould, Reference: "config-linux.md#default-filesystems"},
+}
+
+// ParseLevel takes a string level and returns the OCI compliance level constant
+func ParseLevel(level string) ComplianceLevel {
+	switch strings.ToUpper(level) {
+	case "SHOULD":
+		return ComplianceShould
+	case "MUST":
+		return ComplianceMust
```
wking: This doesn't seem quite right.  [RFC 2119][1] has the following levels:

* MUST-level (MUST, MUST NOT, REQUIRED, SHALL, SHALL NOT).
* SHOULD-level (SHOULD, SHOULD NOT, RECOMMENDED, NOT RECOMMENDED).
* MAY-level (MAY, OPTIONAL).

You've [listed yours in the inverse order from the RFC][2] (and left off [NOT RECOMMENDED][3]?), and that means that your `ComplianceMustNot` is less than your `ComplianceMust`, even though both should be in the MUST-level.  I recommend either:

a. Reorder your constants to put `ComplianceShould` at the bottom of the SHOULD-level constants and `ComplianceMust` at the bottom of the MUST-level constants.
b. Add a function that maps the constants to their level (e.g. `ComplianceLevel(ComplianceShallNot) → ComplianceMust`) and use that when deciding whether errors are fatal.
c. Adjust this case here to select whatever the lowest constant is for that level (e.g. `"MUST" → ComplianceShall`).


[1]: https://tools.ietf.org/html/rfc2119
[2]: https://github.com/opencontainers/runtime-tools/pull/354/files#diff-02e70c9c804a3b4108260b193fedf2d7R13
[3]: https://tools.ietf.org/html/rfc2119#section-4

validate/error.go
```diff
@@ -0,0 +1,69 @@
...
+
+// OCIErrorCode represents the compliance content
+type OCIErrorCode int
+
+const (
+	DefaultFilesystems OCIErrorCode = iota
+)
+
+// OCIError represents an error with compliance level and OCI reference
+type OCIError struct {
```
liangchenye: OK

validate/error.go
```diff
@@ -0,0 +1,69 @@
...
+	Level     ComplianceLevel
+	Reference string
+	Err       error
+}
+
+//FIXME: change to tagged spec releases
```
liangchenye: How about add a field (not in this PR) like this :
       SpecVersions []string

So each OCI Error will be associate with one or more spec versions.
In runtimetest, we can add another option:
       spec-version.
So if an OCI Error does not have the wanted spec-version,  we can ignore it.

validate/error.go
```diff
@@ -0,0 +1,69 @@
+package validate
```
liangchenye: I think image-tool could use the same OCIError struct. But OCIErrorCode can not be shared between runtime-tool and image-tool. I think we can keep it here.

validate/error.go
liangchenye: updated with solution C.

validate/error.go
wking: > So each OCI Error will be associate with one or more spec versions.

I think it's better to only test against the spec for the target version.  But if we're passing around that target version so we can decide what conditions to test, there's probably no need to stash it on `Error`.  I'll withdraw my suggestion, and we can leave the struct fields alone.

validate/error.go
wking: > But OCIErrorCode can not be shared between runtime-tool and image-tool. I think we can keep it here.

Yeah.  A stand-alone module with all the tooling and a public code-> string map, and then the validating consumer populates that map (and spec URL) for whatever it's testing.

liangchenye: The runtime-tool and image-tool are separated, where should we have such a 'stand-alone module', "opencontainers/certification"?

wking: > The runtime-tool and image-tool are separated, where should we have such a 'stand-alone module', ...

I was thinking a separate sub-package in this repo.  Or maybe in runtime-spec, since both image repos currently have some runtime-spec connections.  It seems too tiny for its own repo, and opencontainers/certification seems more focused on policy than implementation.

Mashimiao:
LGTM

liangchenye:
@hqhq @mrunalp  PTAL

completions/bash/oci-runtime-tool
```diff
@@ -119,6 +119,13 @@ __oci-runtime-tool_complete_log_level() {
 	" -- "$cur" ) )
 }
 
+__oci-runtime-tool_complete_compliance_level() {
+	COMPREPLY=( $( compgen -W "
+		must
+		should
```
wking: Missing `may` (which is documented [here][1]).  Although I'm not sure what MAY-level validation looks like, and would be fine dropping it from `--compliance-level`.

[1]: https://github.com/opencontainers/runtime-tools/pull/354/files#diff-6c88621051d8dcd99407582c2b76fd68R713

validate/error.go
wking: With #349 landed we can fix this FIXME.

cmd/runtimetest/main.go
```diff
@@ -264,6 +273,28 @@ func validateRootFS(spec *rspec.Spec) error {
...
+	}
+
+	mountsMap := make(map[string]string)
+	for _, mountInfo := range mountInfos {
+		mountsMap[mountInfo.Mountpoint] = mountInfo.Fstype
+	}
+
+	for fs, fstype := range defaultFS {
+		if !(mountsMap[fs] == fstype) {
+			return ociErr.NewError(ociErr.DefaultFilesystems, fmt.Sprintf("%v must exist and expected type is %v", fs, fstype))
```
wking: “must” → “SHOULD”.

cmd/runtimetest/main.go
```diff
@@ -628,11 +663,16 @@ func validate(context *cli.Context) error {
...
 
+	complianceLevelString := context.String("compliance-level")
+	complianceLevel := ociErr.ParseLevel(complianceLevelString)
 	var validationErrors error
 	for _, v := range defaultValidations {
 		err := v.test(spec)
 		t.Ok(err == nil, v.description)
 		if err != nil {
+			if e, ok := err.(*ociErr.Error); ok && e.Level < complianceLevel {
+				continue
```
wking: We may want to print these as warnings even if they are non-fatal.  With TAP, I'd recommend using [diagnostics][1].  I'm not sure what the best approach is with Go's testing framework, but we should probably figure that out ;).

[1]: https://testanything.org/tap-version-13-specification.html#diagnostics

validate/error.go
```diff
@@ -0,0 +1,91 @@
...
+	case "MUSTNOT":
+		fallthrough
+	case "SHALL":
+		fallthrough
+	case "SHALLNOT":
+		fallthrough
+	case "REQUIRED":
+		return ComplianceMust
+	default:
+		return ComplianceMust
```
wking: I think the default should be to raise an “unrecognized level” error.  Silently defaulting to MUST seems confusing.

validate/error.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+}
+
+// NewError creates an Error by ErrorCode and message
+func NewError(code ErrorCode, msg string) error {
+	err := ociErrors[code]
+	err.Err = errors.New(msg)
```
wking: Is this making a new `Error`?  It looks like it's attaching a new `error` to a pre-existing `Error` (possibly clobbering a previous `error`).  Or maybe we will only ever report one `error` for a given `ErrorCode` and we want that clobbering?  Or maybe I'm misreading this and we are creating a new `Error` ;).

completions/bash/oci-runtime-tool
liangchenye: Lost 'may' in previous PR. We have ['MAY' in spec](https://github.com/opencontainers/runtime-spec/blob/master/config.md#platform-specific-configuration).  In this case, we 'may'  print 'MAY' information  when `platform.os is linux` but the `linux` item is not set.

validate/error.go
liangchenye: I'm creating a new Error :) So the OCIError could be used as a normal Error.

validate/error.go
liangchenye: you are right. I should raise error here.

cmd/runtimetest/main.go
liangchenye: I opened #362 to keep track of this.

liangchenye:
Updated by 
1. add 'may' in completion file
2. reference with 'released' version 
3. return error when ParseLevel with unrecognized string

PTAL @wking @Mashimiao @mrunalp @hqhq

completions/bash/oci-runtime-tool
wking: > In this case, we 'may' print 'MAY' information when `platform.os` is `linux` but the `linux` item is not set.

So would this be all unset OPTIONAL child properties?  Or more case-by case at the runtime-tools maintainers descretion?  Either was, a fatal level seems odd.  If something deserves occasional fatal warnings, it should probably be SHOULD in the spec.

liangchenye:
rebased and fixed the glint error. PTAL  @Mashimiao @hqhq @mrunalp

Mashimiao:
LGTM

liangchenye:
@hqhq @mrunalp  PTAL

liangchenye:
@q384566678 PTAL

cmd/runtimetest/main.go
```diff
@@ -30,6 +32,13 @@ const PrGetNoNewPrivs = 39
 const specConfig = "config.json"
 
 var (
+	defaultFS = map[string]string{
+		"/proc":    "proc",
```
q384566678: proc -> procfs

Mashimiao: proc here is right, there is no procfs such type in Linux

q384566678: https://github.com/opencontainers/runtime-spec/blame/master/config-linux.md#L15
In the `spec` is defined.

Mashimiao: we call it procfs but the real type in Linux is proc. I will submit a PR to fix it in runtime-spec

q384566678: fine.

q384566678:
LGTM

Mashimiao:
LGTM

PR #364, opened 2017-04-14
validate: Increase the windows system mount verification

q384566678:
According to this [spec.](https://github.com/opencontainers/runtime-spec/blame/master/config.md#L58)


wking:
Dup of #256?

q384566678:
My mistakes, I did not notice it.

PR #365, opened 2017-04-17
runtimetest: fix readonly validation 

q384566678:
According to this [spec.](https://github.com/opencontainers/runtime-spec/blame/master/config.md#L36)

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -303,7 +303,7 @@ func testWriteAccess(path string) error {
 func validateRootFS(spec *rspec.Spec) error {
 	logrus.Debugf("validating root filesystem")
 	if spec.Root.Readonly {
-		err := testWriteAccess("/")
+		err := testWriteAccess(spec.Root.Path)
 		if err == nil {
 			return fmt.Errorf("Rootfs should be readonly")
 		}


validate/validate.go
```diff
@@ -138,6 +138,10 @@ func (v *Validator) CheckRootfsPath() (msgs []string) {
 		msgs = append(msgs, fmt.Sprintf("Cannot find the root path %q", rootfsPath))
 	} else if !fi.IsDir() {
 		msgs = append(msgs, fmt.Sprintf("The root path %q is not a directory.", rootfsPath))
+	} else if v.spec.Root.Readonly {
+		if fi.Mode()&0444 == 0 || fi.Mode()|0333 != 0 {
```
wking: This is not a condition on the bundle source.  It's supposed to be a condition on the root after the container mount namespace pivots, and that is checked [here][1].  So I don't think we need bundle-validation code checking anything for `root.readonly` besides “if set, is it a boolean?”, and that is already checked by the JSON parser.

[1]: https://github.com/opencontainers/runtime-tools/blob/bf28e72b0846f5fae890216f64e322fbd19a0041/cmd/runtimetest/main.go#L305-L309

q384566678: You are right, readonly has been verified.

q384566678: But there is a problem with validation in runtimetest, which I have changed.

cmd/runtimetest/main.go
```diff
@@ -303,7 +303,7 @@ func testWriteAccess(path string) error {
 func validateRootFS(spec *rspec.Spec) error {
 	logrus.Debugf("validating root filesystem")
 	if spec.Root.Readonly {
-		err := testWriteAccess("/")
+		err := testWriteAccess(spec.Root.Path)
```
Mashimiao: This is not a correct fix, spec.Root.Path is mounted to "/" in container

PR #366, opened 2017-04-17
validate: increase os verification to checklinux

q384566678:
If `os` is not `linux`, then do not have to implement the next code, improve efficiency.

### DIFF
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -441,6 +441,11 @@ func (v *Validator) CheckOS() (msgs []string) {
 func (v *Validator) CheckLinux() (msgs []string) {
 	logrus.Debugf("check linux")
 
+	if v.spec.Platform.OS != "linux" {
+		logrus.Warnf("%q is not supported to check linux", v.spec.Platform.OS)
+		return
+	}
+
 	var typeList = map[rspec.LinuxNamespaceType]struct {
 		num      int
 		newExist bool


Mashimiao:
OS limitation we have checked in CheckOS() before, here is not necessary.

q384566678:
> OS limitation we have checked in CheckOS() before, here is not necessary.

If the error in the `checkos`, will not immediately return to the error, but also [continue to execute `checklinux`,](https://github.com/opencontainers/runtime-tools/blob/master/validate/validate.go#L91-L92) so the implementation of `checklinux` those excess code, so I think it should increase this judgment.

Mashimiao:
Plaform has been removed from spec, so closing this.

q384566678:
@Mashimiao I think only need rebase on it. This judgment is still needed, as I said above, through this judgment can reduce the number of unnecessary functions to run.

Mashimiao:
reopened, but I think this fix is not suitable.  It will be better if we can add judge if target platform is linux to work with v.spec.Linux != nil

wking:
On Thu, Jul 06, 2017 at 07:35:42AM +0000, Ma Shimiao wrote:
> It will be better if we can add judge if target platform is linux to
> work with v.spec.Linux != nil

We could rebase this PR to use Validator.platform [1].

But with rc6 there's wrong with setting the ‘linux’ property in a
config targeting the ‘linux’ platform [2].  I expect we want to
continue calling CheckLinux whenever the ‘linux’ property is set
(which we currently do [3]).  We probably want to skip the
host-specific checks if the target platform is not Linux, but we can
check the host-independent parts of the ‘linux’ property even if the
target platform is ‘windows’ (for example).  So I'm in favor of
re-closing this PR.

[1]: https://github.com/opencontainers/runtime-tools/blob/0d35fc8437ec8292cd7065e332000c584f7024c5/validate/validate.go#L51
[2]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc6/config.md#L346-L347
[3]: https://github.com/opencontainers/runtime-tools/blob/0d35fc8437ec8292cd7065e332000c584f7024c5/validate/validate.go#L105-L106

q384566678:
I closing the PR, because of  @wking's advice and the change of #416.

PR #367, opened 2017-04-17
validate: perfect deviceValid

q384566678:


Mashimiao:
adding duplicate code is not a good fix, I think the following will be better:
`case "b", "c", "u":`

q384566678:
> adding duplicate code is not a good fix, I think the following will be better:

updated, thanks.

Mashimiao:
LGTM

mrunalp:
LGTM

PR #368, opened 2017-04-18
validate.go：increase the type of checkplatform

q384566678:
https://golang.org/doc/install/source#environment

### DIFF
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -158,7 +158,7 @@ func (v *Validator) CheckPlatform() (msgs []string) {
 		"darwin":    {"386", "amd64", "arm", "arm64"},
 		"dragonfly": {"amd64"},
 		"freebsd":   {"386", "amd64", "arm"},
-		"linux":     {"386", "amd64", "arm", "arm64", "ppc64", "ppc64le", "mips64", "mips64le", "s390x"},
+		"linux":     {"386", "amd64", "arm", "arm64", "mips", "mipsle", "mips64", "mips64le", "ppc64", "ppc64le", "s390x"},
 		"netbsd":    {"386", "amd64", "arm"},
 		"openbsd":   {"386", "amd64", "arm"},
 		"plan9":     {"386", "amd64"},


validate/validate.go
```diff
@@ -158,7 +158,7 @@ func (v *Validator) CheckPlatform() (msgs []string) {
 		"darwin":    {"386", "amd64", "arm", "arm64"},
 		"dragonfly": {"amd64"},
 		"freebsd":   {"386", "amd64", "arm"},
-		"linux":     {"386", "amd64", "arm", "arm64", "ppc64", "ppc64le", "mips64", "mips64le", "s390x"},
+		"linux":     {"386", "amd64", "arm", "arm64", "mips", "mipsle", "mips64", "mips64le", "ppc64", "ppc64le", "s390x"},
```
wking: If we're keeping these alphabetized (which makes sense to me), then `mips64` and `mips64le` should come before `mipsle`.

wking: Also, the fact that Go uses `mips64le` and libseccomp uses `SCMP_ARCH_MIPSEL64` is unfortunate.  Do we want to do anything about supporting either permutation for folks who assume that one value (e.g. `mips64le`) will work for both `--arch` and `--seccomp-arch`?

q384566678: For the things you said I don't have any idea, do you have any good idea?

wking: > For the things you said I don't have any idea, do you have any good idea?

For the alphabetization, [the order described here][1] is:

    … "arm64", "mips", "mips64", "mips64le", "mipsle", "ppc64", …

For the inconsistency between `--arch` and `--seccomp-arch`, I wouldn't do anything unless the maintainers feel like something needs to be done.  I was just bringing the issue to their attention in case they wanted to act.

[1]: https://github.com/opencontainers/runtime-tools/pull/368#discussion_r112497741

q384566678:
@Mashimiao @liangchenye @mrunalp PTAL

mrunalp:
LGTM

Mashimiao:
LGTM

PR #370, opened 2017-04-20
validate: modify the valid value of rootfsPropagation

q384566678:
According to the  [shared subtree](https://www.kernel.org/doc/Documentation/filesystems/sharedsubtree.txt).

### DIFF
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -503,8 +503,10 @@ func (v *Validator) CheckLinux() (msgs []string) {
 	case "rslave":
 	case "shared":
 	case "rshared":
+	case "unbindable":
+	case "runbindable":
 	default:
-		msgs = append(msgs, "rootfsPropagation must be empty or one of \"private|rprivate|slave|rslave|shared|rshared\"")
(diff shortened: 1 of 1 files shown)


validate/validate.go
```diff
@@ -503,8 +503,9 @@ func (v *Validator) CheckLinux() (msgs []string) {
 	case "rslave":
```
q384566678: On the ` `, `rprivate` ,`rslave`, `rshared` these values, I am not sure whether to delete, because these values are not in the `spec`.

wking: I'd rather have them added to the spec.

q384566678: I think I understand your idea, I will do in the `spec`.

q384566678:
Since https://github.com/opencontainers/runtime-spec/pull/770 has been merged，so here also need to change.
@opencontainers/runtime-tools-maintainers PTAL.

liangchenye:
LGTM

Mashimiao:
LGTM

PR #375, opened 2017-04-28
generate: Assign value to devices.type when the value is empty

q384566678:
https://github.com/opencontainers/runtime-spec/blame/master/config-linux.md#L219

### DIFF
--- a/cmd/oci-runtime-tool/generate.go
+++ b/cmd/oci-runtime-tool/generate.go
@@ -141,6 +141,12 @@ func setupSpec(g *generate.Generator, context *cli.Context) error {
 	}
 
 	spec := g.Spec()
+	devices := spec.Linux.Resources.Devices
+	for index := 0; index < len(devices); index++ {
+		if devices[index].Type == "" {
+			devices[index].Type = "a"
+		}
+	}
 
 	if len(spec.Version) == 0 {
 		g.SetVersion(rspec.Version)


liangchenye:
Why should we set this, if a 'type' is empty, it means the template file is invalid, we can drop or ignore it. 'a' is not valid https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#devices

wking:
On Fri, Apr 28, 2017 at 03:42:14AM -0700, 梁辰晔 (Liang Chenye) wrote:
> Why should we set this, if a 'type' is empty, it means the template
> file is invalid, we can drop or ignore it. 'a' is not valid
> https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#devices

That's the wrong section (you're linking linux.devices).
linux.resources.devices is [1], and explicitly says:

  null or unset values mean "all", mapping to a.

Although we should likely drop the ‘null’ from that part of the spec
to catch up with opencontainers/runtime-spec#662.  Still, leaving
linux.resources.devices[].type optional makes sense to me.

I'm less sure that we should be replacing it with ‘a’.  Why don't we
just leave it alone?  Config consumers will have to handle both the
unset case and an explicit ‘a’ anyway, so I don't think we're helping
anyone by filling in the default value here.

[1]: https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#device-whitelist

Mashimiao:
Currently, unset case is also valid, let's leave it alone. People may be confused about that his template is changed.

PR #376, opened 2017-05-09
validate: modify rlimit validation

q384566678:

### DIFF
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -319,14 +319,7 @@ func (v *Validator) CheckRlimits() (msgs []string) {
 				msgs = append(msgs, fmt.Sprintf("rlimit can not contain the same type %q.", process.Rlimits[index].Type))
 			}
 		}
-
-		if v.spec.Platform.OS == "linux" {
-			if err := rlimitValid(rlimit); err != nil {
-				msgs = append(msgs, err.Error())
-			}
-		} else {
-			logrus.Warnf("process.rlimits validation not yet implemented for OS %q", v.spec.Platform.OS)
-		}
+		msgs = append(msgs, v.rlimitValid(rlimit)...)
 	}
 
 	return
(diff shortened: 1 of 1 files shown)


validate/validate.go
```diff
@@ -653,9 +656,6 @@ func envValid(env string) bool {
 }
 
 func rlimitValid(rlimit rspec.LinuxRlimit) error {
-	if rlimit.Hard < rlimit.Soft {
-		return fmt.Errorf("hard limit of rlimit %s should not be less than soft limit", rlimit.Type)
-	}
```
wking: Why pull this out of `rlimitValid`?  My only guess is that you feel that this is not a Linux-specific check, and `rlimitValid` is [currently only run on `linux`][1].  I think a better approach would be to always all `rlimitValid` but only run the type check on Linux.  For example:

```
	for index, rlimit := range process.Rlimits {
		for i := index + 1; i < len(process.Rlimits); i++ {
			if process.Rlimits[index].Type == process.Rlimits[i].Type {
				msgs = append(msgs, fmt.Sprintf("rlimit can not contain the same type %q.", process.Rlimits[index].Type))
			}
		}
		msgs = append(msgs, v.rlimitValid(rlimit)...)
	}
```

and then:

```
 func (v *Validator) checkRlimit(rlimit rspec.LinuxRlimit) (msgs []string) {
	if rlimit.Hard < rlimit.Soft {
		msgs = append(msgs, fmt.Sprintf("hard limit of rlimit %s should not be less than soft limit", rlimit.Type))
	}
	if v.spec.Platform.OS == "linux" {
		…check the type…
	} else {
		logrus.Warnf("process.rlimits validation not yet implemented for OS %q", v.spec.Platform.OS)
	}
}
```

[1]: https://github.com/opencontainers/runtime-tools/blob/69626a4ecf2b91372b146d7f3023c98e3e1a51de/validate/validate.go#L323

mrunalp:
LGTM

Mashimiao:
LGTM

PR #379, opened 2017-05-11
generate: remove default capabilities

Mashimiao:
The runtime-spec doesn't say process in container
should have such default capabilities, so remove them.


wking:
On Wed, May 10, 2017 at 07:36:08PM -0700, Ma Shimiao wrote:
> The runtime-spec doesn't say process in container
> should have such default capabilities, so remove them.

‘generate’ is not about spec compliance (that's ‘validate’), it's
about creating a useful starting point for users.  So runtime-tools
maintainers are free to do whatever they like in ‘New()’ if they think
it makes life easier for folks who don't have existing opinions about
what they want in the spec.

Mashimiao:
On 05/11/2017 01:19 PM, W. Trevor King wrote:
> ‘generate’ is not about spec compliance (that's ‘validate’), it's
> about creating a useful starting point for users.

Agree.
But I think we can't make sure which capability is necessary for user,
just leave it alone without default value will be better.

wking:
On Wed, May 10, 2017 at 10:53:14PM -0700, Ma Shimiao wrote:
> But I think we can't make sure which capability is necessary for
> user, just leave it alone without default value will be better.

I agree, but with that approach the best default config is probably
the minimal one for the platform (e.g. [1]) with the additional
recommended mounts [2].  The current default is really somewhere
between “useful for newcomers out of the box” and “demonstrates a
bunch of things so you can trim it down”.  The latter goal is also
served by [3], so maybe this PR is the first stage of a pivot towards
a more minimal default?

[1]: https://github.com/opencontainers/runtime-spec/blob/1259a08e003658f8c4094a6bbc9f52aee29c95df/schema/test/config/good/minimal.json
[2]: https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#default-filesystems
[3]: https://github.com/opencontainers/runtime-spec/blob/master/config.md#configuration-schema-example

Mashimiao:
In my own opinion, `generate`  should output config info based on what runtime-spec minimally request in default. Any optional or additional fields in config should be specified by users which may be by options of command generate.

wking:
On Thu, May 11, 2017 at 12:44:47AM -0700, Ma Shimiao wrote:
> In my own opinion, `generate` should output config info based on
> what runtime-spec minimally request in default.

So why is this PR just trimming the capabilities?  If the goal is
minimal (including the SHOULD-level mounts [1]?), you could work up a
much bigger cut pretty quickly.  Are there future plans for that and
some reason that taking this one group of properties at a time?

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/config-linux.md#default-filesystems

Mashimiao:
Before I sent this PR, I haven't so much about minimal or not. I just found the default caps will cause runtimetest fails, so began to fix it.

mrunalp:
We can use templates for minimal.

Mashimiao:
@wking  @mrunalp  PTAL

generate/generate.go
```diff
@@ -39,180 +39,17 @@ func New() Generator {
 			OS:   runtime.GOOS,
 			Arch: runtime.GOARCH,
 		},
-		Root: rspec.Root{
-			Path:     "",
```
wking: This is a [REQUIRED property][1], so I think you should keep this.  If you don't like the empty-string value, I'd be ok with a `"FIXME"` default.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/config.md#root

generate/generate.go
```diff
@@ -39,180 +39,17 @@ func New() Generator {
...
-			Readonly: false,
-		},
 		Process: rspec.Process{
-			Terminal: false,
-			User:     rspec.User{},
+			User: rspec.User{
+				UID: 0,
+				GID: 0,
```
wking: These are the Go zero values, so you can stick with `User:     rspec.User{}`.  I'm not opposed to explicitly setting them (like you're currently doing in this PR) though.

generate/generate.go
```diff
@@ -39,180 +39,17 @@ func New() Generator {
...
-					Type: "RLIMIT_NOFILE",
-					Hard: uint64(1024),
-					Soft: uint64(1024),
-				},
-			},
-		},
-		Hostname: "mrsdalloway",
-		Mounts: []rspec.Mount{
-			{
-				Destination: "/proc",
```
wking: I think you want to keep this (and a few of the other entries) to follow [the runtime's SHOULD][1].

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/config-linux.md#default-filesystems

generate/generate.go
Mashimiao: Fine.

generate/generate.go
Mashimiao: Because option --rootfs-path has default value `rootfs`, then I think keep it or not doesn't matter , so removed it.

generate/generate.go
Mashimiao: I think that's a question.
Should we let spec should-level properties in minimal template? @wking  @mrunalp

wking: > Should we let spec should-level properties in minimal template? 

I think so, because there is no reason to assume all callers have "[carefully weighed (the full implications) before choosing a different course][1]".  Callers who have can always clear `mounts`.

[1]: https://tools.ietf.org/html/rfc2119#section-3

Mashimiao: OK, updated

Mashimiao:
@wking  @mrunalp  @hqhq  @liangchenye  PTAL

liangchenye:
With this title 'output minial tempalte in default', I prefer to provide template file for 'linux/windows/solaris' directly, just like:
https://github.com/opencontainers/runtime-spec/tree/master/schema/test/config/good

Maybe we can add them to `runtime-tools/contrib/templates`?

Mashimiao:
Agree, we can supply default good templates.  But I think without users' spcified options, generate outputs minimal template is also right.

PR #380, opened 2017-05-12
generate: drop caps in every cap set

runcom:
commit afc8d3588cbf9c5de0be4ab00d352077be78a0d3 updated the
runtime-spec version but introduced a bug when dropping capabilities.
If you drop a capability it will only be dropped in the Bounding
capabilities set while kept in the other sets.
This causes a bug when using the generated config.json in which
containers cannot be started at all.
This patch fixes the above by dropping `return nil` statements.

@mrunalp @Mashimiao PTAL asap

Signed-off-by: Antonio Murdaca <runcom@redhat.com>


Mashimiao:
LGTM 
though I think that's a not precise fix, but as an emergency fix it's OK to me.
We will need more work on capabilities adding and removing.

rhatdan:
LGTM

mrunalp:
LGTM

PR #386, opened 2017-05-22
Support SELinux Check

Mashimiao:
1. update godeps and add go-selinux deps
2. add selinux check in validate and runtimetest

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -24,6 +24,7 @@ import (
 	"github.com/opencontainers/runtime-tools/cmd/runtimetest/mount"
 	rfc2119 "github.com/opencontainers/runtime-tools/error"
 	"github.com/opencontainers/runtime-tools/specerror"
+	"github.com/opencontainers/selinux/go-selinux/label"
 
 	"golang.org/x/sys/unix"
 )
@@ -778,6 +779,20 @@ func validatePosixMounts(spec *rspec.Spec) error {
 	return mountErrs
 }
 
+func validateMountLabel(spec *rspec.Spec) error {
+	for _, mount := range spec.Mounts {
+		fileLabel, err := label.GetFileLabel(mount.Destination)
+		if err != nil {
+			return fmt.Errorf("Failed to get mountLabel of %v", mount.Destination)
+		}
+		if fileLabel != spec.Linux.MountLabel {
+			return fmt.Errorf("Expected mountLabel %v, actual %v", spec.Linux.MountLabel, fileLabel)
+		}
+	}
+
+	return nil
+}
+
 func run(context *cli.Context) error {
 	logLevelString := context.String("log-level")
 	logLevel, err := logrus.ParseLevel(logLevelString)
@@ -879,6 +894,10 @@ func run(context *cli.Context) error {
 		{
 			test:        validateGIDMappings,
 			description: "gid mappings",
+		},
+		{
+			test:        validateMountLabel,
+			description: "mountLabel",
 		},
 	}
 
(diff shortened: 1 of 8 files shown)


cmd/runtimetest/main.go
```diff
@@ -594,6 +595,20 @@ func validateMountsExist(spec *rspec.Spec) error {
...
 }
 
+func validateMountLabel(spec *rspec.Spec) error {
+	for _, mount := range spec.Mounts {
+		fileLabel, err := label.GetFileLabel(mount.Destination)
+		if err != nil {
+			return fmt.Errorf("Failed to get mountLabel of %v", mount.Destination)
+		}
+		if fileLabel != spec.Linux.MountLabel {
+			return fmt.Errorf("Expected mountLabel %v, actual %v", spec.Linux.MountLabel, fileLabel)
```
wking: There are no [RFC 2119 requirements for the runtime for `linux.mountLabel`][1].  I think we need to get spec language about what the runtime MUST do before we can fail a runtime for not doing it (opencontainers/runtime-spec#746).

[1]: https://github.com/opencontainers/runtime-spec/blob/fae0d09a24b78ef17e43d411583780b9f2a48d5b/config-linux.md#mount-label

validate/validate.go
```diff
@@ -519,6 +520,12 @@ func (v *Validator) CheckLinux() (msgs []string) {
 		}
 	}
 
+	if v.spec.Linux.MountLabel != "" {
+		if err := label.Validate(v.spec.Linux.MountLabel); err != nil {
```
wking: I'm not sure what `label.Validate` is checking, but the spec [places no limits on the `linux.mountLabel` value][1].  If we want to validate it (against a particular kernel API?), then I think we probably need clarity in the spec about what API the value will be passed to, and the check here should probably only happen when `HostSpecific` is set.

[1]: https://github.com/opencontainers/runtime-spec/blob/fae0d09a24b78ef17e43d411583780b9f2a48d5b/config-linux.md#mount-label

Mashimiao: The validation of mountLabel is not host-specific. It just checks whether label contains unexpected options. Please refer to [label.Validate()](https://github.com/opencontainers/selinux/blob/master/go-selinux/label/label_selinux.go#L189)

wking: > It just checks whether label contains unexpected options.

But there is nothing in the spec about “MUST be a valid SELinux label according to [some SELinux spec]” or “MUST NOT contain `z` or `Z`” or anything like that.  Currently spec-readers have to make that jump on their own.

And I was suggesting host-specific because even if `z` and `Z` are not legal labels for today's kernel, are we sure that they will be illegal labels for *all* kernels?  Maybe someone compiles a Linux kernel where they are legal, and the spec's kernel-punt-where-possible approach is designed to support that sort of thing.

q384566678:
LGTM

validate/validate.go
wking: I still don't see spec grounds for this check, but if the runtime-tools maintainers feel like it's grounded, then the `validate/` changes from this PR belong in [v0.3.0][1].  That could either happen by spinning them out into a new PR, or by reviewing them here in the context of the runtimtetest changes.

cmd/runtimetest/main.go
```diff
@@ -778,6 +779,20 @@ func validatePosixMounts(spec *rspec.Spec) error {
 	return mountErrs
 }
 
+func validateMountLabel(spec *rspec.Spec) error {
+	for _, mount := range spec.Mounts {
+		fileLabel, err := label.GetFileLabel(mount.Destination)
+		if err != nil {
+			return fmt.Errorf("Failed to get mountLabel of %v", mount.Destination)
+		}
+		if fileLabel != spec.Linux.MountLabel {
```
q384566678: Linux judgment should be increased to avoid panic.

Mashimiao: updated

q384566678:
LGTM

PR #387, opened 2017-05-23
use current pid instead of 1

Mashimiao:
If we don't use pid namespace, the process's pid will not be 1


liangchenye:
LGTM

Mashimiao:
ping @mrunalp  @hqhq

hqhq:
LGTM

PR #390, opened 2017-06-08
add warning for template option

Mashimiao:

### DIFF
--- a/man/oci-runtime-tool-generate.1.md
+++ b/man/oci-runtime-tool-generate.1.md
@@ -317,6 +317,7 @@ read the configuration from `config.json`.
 **--template**=PATH
   Override the default template with your own.
   Additional options will only adjust the relevant portions of your template.
+  Templates are not validated for correctness, so the user should ensure that they are correct.
 
 **--tmpfs**=[] Create a tmpfs mount
   Mount a temporary filesystem (`tmpfs`) mount into a container, for example:


Mashimiao:

man/oci-runtime-tool-generate.1.md
```diff
@@ -317,6 +317,7 @@ read the configuration from `config.json`.
 **--template**=PATH
   Override the default template with your own.
   Additional options will only adjust the relevant portions of your template.
+  whether template is valid or not will not be checked, this may lead to finally generated config file can't work directly. Currently, users should guarantee template is valid themselves.
```
mrunalp: what do you mean by can't work directly?

Mashimiao: For example, as we don't have template validation now, if there are duplicated namespace entries in user's template, the final config file will also have duplicated namespace. Then the config file is invalid and can't be used directly. 
We need to solve such problems, but not easy, should reform and reuse validate fucntions, currently adding warning will be better.

mrunalp: @Mashimiao Thanks! Maybe word this as
"Templates are not validated for correctness, so the user should ensure that they are correct."

Mashimiao: fine, updated

Mashimiao:
@opencontainers/runtime-tools-maintainers  PTAL

liangchenye:
@Mashimiao I think we should add 'template validation' .

mrunalp:
We can add that in a follow-on I think. LGTM.

Mashimiao:
@liangchenye If we want to add 'template validation', we need to make some modifications on Generation and validaiton. I think we can't finish them in short period, so I submitted this PR to give a warn to users.
Agree with @mrunalp,  we can add it in another PR in the future.

liangchenye:
OK, I add an issue to keep track of this.  https://github.com/opencontainers/runtime-tools/issues/403

liangchenye:
LGTM

hqhq:
LGTM

PR #391, opened 2017-06-09
Split runtime validate

Mashimiao:
If we want to totally validate a runtime, we do not only need to check container inside environment but also outside settings like cgroup, label, lifecycle. So I split runtimevalidate into to parts: inside and outside.
The following is an example output:
`# make localvalidation 
RUNTIME=runc go test -tags ""  -v github.com/opencontainers/runtime-tools/validation
=== RUN   TestValidateRuntimeInside
TAP version 13
ok 1 - root filesystem
ok 2 - hostname
not ok 3 - mounts
not ok 4 - capabilities
ok 5 - default symlinks
ok 6 - default devices
ok 7 - linux devices
ok 8 - linux process
ok 9 - masked paths
ok 10 - oom score adj
ok 11 - read only paths
ok 12 - rlimits
ok 13 - sysctls
ok 14 - uid mappings
ok 15 - gid mappings
1..15
2 errors occurred:

* Expected mount {/volume/testing bind /tmp [rw bind]} does not exist
* Expected ambient capability chown not set for process
--- FAIL: TestValidateRuntimeInside (0.27s)
	validation_test.go:42: runc failed validation: exit status 1
=== RUN   TestValidateRuntimeOutside
TAP version 13
ok 1 - labels
1..1
--- PASS: TestValidateRuntimeOutside (0.08s)
FAIL
exit status 1
FAIL	github.com/opencontainers/runtime-tools/validation	0.350s
Makefile:40: recipe for target 'localvalidation' failed
make: *** [localvalidation] Error 1
`

### DIFF
--- a/validation/validation_test.go
+++ b/validation/validation_test.go
@@ -1,14 +1,19 @@
 package validation
 
 import (
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 	"testing"
 
+	"github.com/hashicorp/go-multierror"
+	"github.com/mndrix/tap-go"
 	"github.com/mrunalp/fileutils"
+	rspec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/runtime-tools/generate"
 	"github.com/satori/go.uuid"
 )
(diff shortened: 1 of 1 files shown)


wking:
On Fri, Jun 09, 2017 at 01:03:08AM -0700, Ma Shimiao wrote:
>   * add error catch

This seems like an orthogonal, easy-to-review change.  Maybe split it
out into its own PR so it can get in more quickly than the other
commit in this PR (which is a bit more complicated ;)?

validation/validation_test.go
```diff
@@ -17,54 +22,90 @@ var (
...
+		t.Errorf("%s failed validation: %v", runtime, err)
+	}
+}
+
+func TestValidateRuntimeOutside(t *testing.T) {
+	g, err := getDefaultGenerator()
+	if err != nil {
+		t.Errorf("%s failed validation: %v", runtime, err)
+	}
+	g.SetProcessArgs([]string{"sleep", "50"})
```
wking: Why are we setting this at all, since `sleep` may not exist inside the container?  [`process` is optional][1] (since opencontainers/runtime-spec#701), so I'd rather just unset `process` completely and call `create` (but never `start`) during these outer tests.

[1]: https://github.com/opencontainers/runtime-spec/blob/b5017e22ec9592d9e4935812fdce8f9a6b71a365/config.md#process

Mashimiao:
On 06/15/2017 01:05 AM, W. Trevor King wrote:
> Why are we setting this at all, since |sleep| may not exist inside the container? |process| is optional <https://github.com/opencontainers/runtime-spec/blob/b5017e22ec9592d9e4935812fdce8f9a6b71a365/config.md#process> (since opencontainers/runtime-spec#701 <https://github.com/opencontainers/runtime-spec/pull/701>), so I'd rather just unset |process| completely and call |create| (but never |start|) during these outer tests.

Currently, it seems a good idea, so updated.

Mashimiao:

validation/validation_test.go
```diff
@@ -17,51 +22,90 @@ var (
...
+	}
+
+	bundleDir, _, err := prepareBundle(g)
 	if err != nil {
 		return err
 	}
+	//	defer os.RemoveAll(bundleDir)
 
 	// TODO: Use a library to split run into create/start
 	// Launch the OCI runtime
```
wking: You're using `create` below, so this comment is stale here.  You will want to run `kill` and `delete` to clean up though.

Mashimiao:
updated, PTAL

Mashimiao:
Rebased, PTAL

liangchenye:
I like this change. It makes lifecycle testing easier. 
It is the first step, we need to check the output result when 'create/start/stop'.

validation/validation_test.go
```diff
@@ -17,54 +22,92 @@ var (
...
 	}
 
-	// Setup a temporary test directory
-	tmpDir, err := ioutil.TempDir("", "ocitest")
+	bundleDir, rootfsDir, err := prepareBundle(g)
 	if err != nil {
 		return err
 	}
-	defer os.RemoveAll(tmpDir)
+	//	defer os.RemoveAll(bundleDir)
```
wking: I'm guessing this change was for local testing, and that we want to restore the deferred removal before this PR lands?

validation/validation_test.go
```diff
@@ -17,54 +22,92 @@ var (
...
 
-	// Create bundle directory for the test container
-	bundleDir := tmpDir + "/busybox"
-	if err := os.MkdirAll(bundleDir, 0755); err != nil {
+	// Copy the runtimetest binary to the rootfs
+	err = fileutils.CopyFile("../runtimetest", filepath.Join(rootfsDir, "runtimetest"))
```
wking: Does this actually work?  My local build of `runtimetest` is dynamically linked:

```
$ lddtree runtimetest 
runtimetest (interpreter => /lib64/ld-linux-x86-64.so.2)
    libpthread.so.0 => /lib64/libpthread.so.0
    libc.so.6 => /lib64/libc.so.6
```

and neither that linker nor those libraries aren't part of our Busybox `rootfs`.  For example, here is the statically-linked Busybox working fine in a chroot based on that `rootfs`:

```
$ unshare -rUmfp --mount-proc sh -c 'mount --rbind /proc proc && chroot . sh -c "pwd && ls"'
/
bin          etc          proc         runtimetest
```

And here is `runtimetest` failing to launch in the same situation:

```
$ unshare -rUmfp --mount-proc sh -c 'mount --rbind /proc proc && chroot . /runtimetest'
chroot: failed to run command ‘/runtimetest’: No such file or directory
```

I expect we're going to need to either compile `runtimetest` as a static binary (#442), copy the linker and required libraries into `rootfs`, or use `busybox` for our internal tests.

validation/validation_test.go
Mashimiao: Ah... yes, this is a mistake

validation/validation_test.go
Mashimiao: we need #442 merged first

liangchenye:
Hi @Mashimiao, 
One thing we miss is, we don't have test cases.
Inspired by this PR, I think we can add more Test** functions.

For example, we can add TestCreate function. In this function, we provides testing cases to cover
different scenarios, for example: 
1. success in create 
2. fail to create due to args missing
3. fail to create since the id is duplicated
... 

I used to add test like this:
https://github.com/liangchenye/ocitools/commit/b55cfcd789fa1e7299ea8b4f996f6a13834d3ceb

wking:
I think this PR is obsolete, or at least in need of a fairly involved rebase, now that #447 has landed.

PR #392, opened 2017-06-09
runtimetest: fix mounts check

Mashimiao:
Even if bind mount succeed, there will not any bind or rbind key word
in type of options of mount.
So, we can not make sure if bind mount succeed.
It'd better to skip bind mount check

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -557,6 +557,11 @@ func validateMountsExist(spec *rspec.Spec) error {
 	}
 
 	for _, specMount := range spec.Mounts {
+		if specMount.Type == "bind" || specMount.Type == "rbind" {
+			// TODO: add bind or rbind check.
+			continue
+		}
+
 		found := false
 		for _, sysMount := range mountsMap[filepath.Clean(specMount.Destination)] {
 			if err := mountMatch(specMount, sysMount); err == nil {


Mashimiao:

mrunalp:
I think I am okay with this. I will let other maintainers chime in as well.

wking:
bind and rbind are still not covered by the spec
(opencontainers/runtime-spec#771), but it would be good to figure out
how we plan on testing them once they are.  The “check after” approach
would probably involve a special test in validation_test.go where we
setup a custom config and source filesystem, run ‘create’, create new
files in the mounted filesystems, and check for the expected
propagation.  The “check during” approach would be to ptrace(2) the
`create` process and look for the expected mount(2) calls.  But I
don't see a good way to handle these checks in a generic runtimetest
way, since we don't have much information about the initial filesystem
in that case.  Unless we *do* have plans for a generic runtimetest
approach, I'd rather silently skip (r)bind mount entries in
runtimetest instead of warning about the skips.  Or maybe use Debugf
instead of Warnf if you don't want to be completely silent.

Also note that opencontainers/runtime-spec#771 (which I hope will get
revived before 1.0, and is currently blocked by
opencontainers/runc#1460) is not using ‘type’ for (r)bind [1].  I
think runtimetest would be more robust if it switched on the presence
of (r)bind options [2,3] for whatever logic we end up using.  Also
note that move [4], private [5], etc. are probably going to need the
same sort of handling that we give (r)bind.

[1]: https://github.com/opencontainers/runtime-spec/pull/771/files#diff-c9c91c29b41257aea3a3403cc606ad99R65
[2]: https://github.com/opencontainers/runtime-spec/pull/771/files#diff-c9c91c29b41257aea3a3403cc606ad99R81
[3]: https://github.com/opencontainers/runtime-spec/pull/771/files#diff-c9c91c29b41257aea3a3403cc606ad99R104
[4]: https://github.com/opencontainers/runtime-spec/pull/771/files#diff-c9c91c29b41257aea3a3403cc606ad99R91
[5]: https://github.com/opencontainers/runtime-spec/pull/771/files#diff-c9c91c29b41257aea3a3403cc606ad99R103

liangchenye:
I don't think it is necessary to add this warning.
I'd rather to add a 'TODO' annotation to this function.

Mashimiao:
@liangchenye  OK, I'm fine with adding TODO.  we may add it after #391  landed.

Mashimiao:
updated, @mrunalp  @liangchenye  @hqhq  PTAL

mrunalp:
LGTM

liangchenye:
LGTM

PR #397, opened 2017-06-16
validate: modify the validation of mounts.type

q384566678:
According to this [PR](https://github.com/opencontainers/runtime-spec/pull/854) revision.


validate/validate.go
```diff
@@ -337,11 +337,11 @@ func (v *Validator) CheckRlimits() (msgs []string) {
...
 	supportedTypes := make(map[string]bool)
 
-	if OS != "linux" && OS != "windows" {
+	if OS != "linux" && OS != "solaris" {
 		logrus.Warnf("%v is not supported to check mount type", OS)
 		return nil, nil
-	} else if OS == "windows" {
-		supportedTypes["ntfs"] = true
+	} else if OS == "solaris" {
+		supportedTypes["lofs"] = true
```
q384566678: On the `type` of `solaris` support, I found the following sentence in the relevant documents:
> Used to specify the FSType on which to operate. The FSType must be specified or must be determinable from /etc/vfstab, or by consulting /etc/default/fs or /etc/dfs/fstypes. 

 But I did not find these documents, only given examples given a support `type`.

Mashimiao: This is not correct modification.  You can refer to FSTYPE in [mount(1)](http://docs.oracle.com/cd/E86824_01/html/E54764/mount-1m.html#REFMAN1Mmount-1m) to determine solaris supports fs types.
Windows does not support type in mount, there is no need to output warning and we should accept this change after spec v1.0 released.

validate/validate.go
```diff
@@ -337,11 +337,31 @@ func (v *Validator) CheckRlimits() (msgs []string) {
 func supportedMountTypes(OS string, hostSpecific bool) (map[string]bool, error) {
 	supportedTypes := make(map[string]bool)
 
-	if OS != "linux" && OS != "windows" {
+	if OS != "linux" && OS != "solaris" {
 		logrus.Warnf("%v is not supported to check mount type", OS)
```
wking: Repeating @Mashimiao's [earlier comment][1], since it still applies:

> Windows does not support type in mount, there is no need to output warning and we should accept this change after spec v1.0 released.

`type` is not defined for Windows since opencontainers/runtime-spec#854 (which you site in the comment opening this PR).  So configs which set it are taking advantage of the [extensibility rules][2] that allow them to do so.  Individual runtimes may or may not act upon the extention, but it's still a valid config.

There may be users who are interested in knowing if a runtime MUST support a given config (e.g. if that config can be portably used with any OCI-compliant runtime).  In that case, we would want to warn/error when we found extentions like this.  More thoughts on that sort of thing in opencontainers/image-tools#66.  However, neither runtime-tools nor image-tools currently provides switches for that sort of thing, so for the time being I'd rather silently ignore the `type` value (if any) in Windows configs.

[1]: https://github.com/opencontainers/runtime-tools/pull/397#discussion_r122376942
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#extensibility

Mashimiao:
I‘m not familiar with solaris system, so I'm not sure if this is a right validation. But from [here](http://docs.oracle.com/cd/E86824_01/html/E54775/vfstab-4.html#REFMAN4vfstab-4), It seems vfstab stores default mounted file systems not all supported file systems.
Any comments from @opencontainers/runtime-tools-maintainers

wking:
Why is this in [v0.7.0][1]?  I don't see anything about “solve options conflict problem” here (I'd expect v0.7.0 to only contain #414, but it currently doesn't).  This looks like part of implementing config validation, so it belonged in [v0.2.0][2].  And with v0.2.0 cut, it should fall back to [v0.3.0][3].

[1]: https://github.com/opencontainers/runtime-tools/milestone/3
[2]: https://github.com/opencontainers/runtime-tools/milestone/5
[3]: https://github.com/opencontainers/runtime-tools/milestone/5

Mashimiao:
It's a validation about solaris but we are not familiar with it, at least I@liangchenye and I. We can't make sure whether it's right or not just from solaris's reference document. And I think we should perfect Linux related validations first. Other platform's may be later. 
So, moving it to v0.7, just want to put it into low priority.
If anyone's familiar with solaris, you can review this PR. Though it's in v0.7, we can merge this before v0.7.

PR #398, opened 2017-06-16
validate: increase the validation of windows fields

q384566678:
https://github.com/opencontainers/runtime-spec/blame/master/config.md#L349

### DIFF
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -424,7 +424,11 @@ func (v *Validator) CheckPlatform() (msgs []string) {
 		}
 	}
 
-	if v.platform != "windows" {
+	if v.platform == "windows" {
+		if v.spec.Windows == nil {
+			msgs = append(msgs, "'windows' MUST be set when platform is `windows`")
+		}
+	} else {
 		if v.spec.Windows != nil {
 			msgs = append(msgs, fmt.Sprintf("'windows' MUST NOT be set when target platform is %q", v.platform))
 		}


wking:
On Thu, Jun 15, 2017 at 11:53:51PM -0700, Zhou Hao wrote:
> https://github.com/opencontainers/runtime-spec/blame/master/config.md#L353

runtime-tools currently targets runtime-spec v1.0.0-rc5 [1,2], and the
requirement you're checking here doesn't exist there [3].  I recommend
we table this PR until runtime-spec cuts an rc6.

[1]: https://github.com/opencontainers/runtime-tools/blob/7150e8d99ea8ba04b8a763bc4108c252e2d5d1fc/validate/validate.go#L152-L153
[2]: https://github.com/opencontainers/runtime-tools/blob/7150e8d99ea8ba04b8a763bc4108c252e2d5d1fc/Godeps/Godeps.json#L35-L36
[3]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc5/config.md#L325-L326

Mashimiao:
need rebase and modify

q384566678:
updated.

Mashimiao:
LGTM

q384566678:
@hqhq @liangchenye PTAL

liangchenye:
I'm checking the spec, it seems the spec changes.

q384566678:
I updated the link.

liangchenye:
LGTM  . but the spec changes, some of the related code may not necessary, I'll add a following PR.

validate/validate.go
```diff
@@ -424,7 +424,11 @@ func (v *Validator) CheckPlatform() (msgs []string) {
 		}
 	}
 
-	if v.platform != "windows" {
+	if v.platform == "windows" {
+		if v.spec.Windows == nil {
+			msgs = append(msgs, "'windows' MUST be set when platform is `windows`")
```
wking: This is [not true in rc5][1].

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc5/config.md#L325-L326

validate/validate.go
```diff
@@ -424,7 +424,11 @@ func (v *Validator) CheckPlatform() (msgs []string) {
...
+		}
+	} else {
 		if v.spec.Windows != nil {
 			msgs = append(msgs, fmt.Sprintf("'windows' MUST NOT be set when target platform is %q", v.platform))
```
wking: This is [no longer true in rc6][1].

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc6/config.md#L348-L349

wking: > This is no longer true in rc6.

And that is getting fixed by #416.  Once #416 lands, the only remaining issue will be that we're using v6 logic but [requiring][1] [1.0.0rc5][2], and that's not a big deal in between runtime-tools releases.

[1]: https://github.com/opencontainers/runtime-tools/blob/0d35fc8437ec8292cd7065e332000c584f7024c5/validate/validate.go#L164-L165
[2]: https://github.com/opencontainers/runtime-tools/blob/0d35fc8437ec8292cd7065e332000c584f7024c5/vendor/github.com/opencontainers/runtime-spec/specs-go/version.go#L6-L18

PR #399, opened 2017-06-21
fix seccomp-default parse and remove problem

Mashimiao:
1. completion for --seccomp-default is not correct, need to fix it and manpage.
2. There is no need to call parseAction() twice, we can pass type rspec.LinuxSeccompAction directly.
3. There is no need to distinguish whether string contains ',' or not, split can handle them.


mrunalp:
Can you add some description to motivate the change?

Mashimiao:
updated, PTAL

mrunalp:
LGTM

PR #402, opened 2017-06-22
.travis.yml: Run 'make test' in Travis

wking:
To catch things like [gofmt errors][1] in CI testing (vs. leaving that to manual review).  Also a few other changes so that .gofmt and .golint errors will cause 'make test' to fail.

[1]: https://github.com/opencontainers/runtime-tools/pull/400#discussion_r123550759


wking:
Now that I've pushed a73b6e5, [Travis is happy][1].

[1]: https://travis-ci.org/opencontainers/runtime-tools/builds/245873908

.travis.yml
```diff
@@ -15,4 +15,5 @@ script:
   - $HOME/gopath/bin/golint ./...
   - $HOME/gopath/bin/git-validation -run DCO,short-subject -v -range ${TRAVIS_COMMIT_RANGE}
   - make
+  - make test
```
Mashimiao: `test` contians govet, golint. There is no need to call them twice, we should remove them from travis.yml

wking: > `test` contians govet, golint…

Good point.  Fixed in a73b6e5 → 3c3079d.

Mashimiao:
LGTM

liangchenye:
LGTM

Issue #407, opened 2017-07-04
TODO: remove 'platform' validate


liangchenye:
platform is removed from spec, so all the validation code should be removed.

hqhq:
#dibs

PR #409, opened 2017-07-05
Remove platform

hqhq:
Fixes: #407

### DIFF
--- a/cmd/oci-runtime-tool/generate.go
+++ b/cmd/oci-runtime-tool/generate.go
@@ -5,7 +5,6 @@ import (
 	"bytes"
 	"fmt"
 	"os"
-	"runtime"
 	"strconv"
 	"strings"
 	"unicode"
@@ -19,7 +18,6 @@ import (
 
 var generateFlags = []cli.Flag{
 	cli.StringFlag{Name: "apparmor", Usage: "specifies the the apparmor profile for the container"},
-	cli.StringFlag{Name: "arch", Value: runtime.GOARCH, Usage: "architecture the container is created for"},
 	cli.StringSliceFlag{Name: "args", Usage: "command to run in the container"},
 	cli.StringSliceFlag{Name: "bind", Usage: "bind mount directories src:dest[:options...]"},
 	cli.StringSliceFlag{Name: "cap-add", Usage: "add Linux capabilities"},
@@ -62,7 +60,6 @@ var generateFlags = []cli.Flag{
 	cli.StringFlag{Name: "mount-label", Usage: "selinux mount context label"},
 	cli.BoolFlag{Name: "no-new-privileges", Usage: "set no new privileges bit for the container process"},
 	cli.IntFlag{Name: "oom-score-adj", Usage: "oom_score_adj for the container"},
-	cli.StringFlag{Name: "os", Value: runtime.GOOS, Usage: "operating system the container is created for"},
 	cli.StringFlag{Name: "output", Usage: "output file (defaults to stdout)"},
 	cli.StringSliceFlag{Name: "poststart", Usage: "set command to run in poststart hooks"},
 	cli.StringSliceFlag{Name: "poststop", Usage: "set command to run in poststop hooks"},
@@ -150,9 +147,6 @@ func setupSpec(g *generate.Generator, context *cli.Context) error {
 	if context.IsSet("hostname") {
 		g.SetHostname(context.String("hostname"))
 	}
-
-	g.SetPlatformOS(context.String("os"))
-	g.SetPlatformArch(context.String("arch"))
 
 	if context.IsSet("label") {
 		annotations := context.StringSlice("label")
--- a/cmd/oci-runtime-tool/validate.go
+++ b/cmd/oci-runtime-tool/validate.go
@@ -20,7 +20,8 @@ var bundleValidateCommand = cli.Command{
 	Action: func(context *cli.Context) error {
 		inputPath := context.String("path")
 		hostSpecific := context.GlobalBool("host-specific")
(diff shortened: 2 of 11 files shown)


hqhq:
ping @Mashimiao @liangchenye @mrunalp 

I know we have to update runtime-spec to fully remove platform in runtime-tools, I haven't done this because I'm not sure how much amends we have to do to adapt the latest runtime-spec, I might not have time for this for now, and it's safe we just remove checking platform in runtime-tools.

hqhq:
CI failed, fixing.

cmd/runtimetest/main.go
```diff
@@ -695,7 +697,12 @@ func main() {
 		cli.StringFlag{
 			Name:  "path",
 			Value: ".",
-			Usage: "path to the configuration",
+			Usage: "Path to the configuration",
+		},
+		cli.StringFlag{
+			Name:  "platform",
+			Value: "Linux",
```
Mashimiao: should be linux

hqhq:
CI passed, PTAL.

cmd/oci-runtime-tool/generate.go
```diff
@@ -151,9 +151,6 @@ func setupSpec(g *generate.Generator, context *cli.Context) error {
 		g.SetHostname(context.String("hostname"))
 	}
 
-	g.SetPlatformOS(context.String("os"))
-	g.SetPlatformArch(context.String("arch"))
```
Mashimiao: I think you also need to remove options os and arch from cli.Flag and manpage

validate/validate.go
```diff
@@ -423,25 +396,30 @@ func (v *Validator) CheckMounts() (msgs []string) {
...
 }
 
-// CheckOS checks v.spec.Platform.OS
-func (v *Validator) CheckOS() (msgs []string) {
-	logrus.Debugf("check os")
+// CheckPlatform checks v.platform
+func (v *Validator) CheckPlatform() (msgs []string) {
+	logrus.Debugf("check platform")
+
+	if v.platform != "linux" && v.platform != "solaris" && v.platform != "windows" {
```
Mashimiao: So, spec just supports linux, solaris and windows, all other platforms are invalid?

hqhq: For now, I guess the answer is YES, see https://github.com/opencontainers/runtime-spec/blob/master/spec.md#platforms

hqhq:
@Mashimiao Fixed, thanks.

validate/validate.go
liangchenye: I think we should check
```
if v.platform != "linux" && v.platform != "solaris" && v.platform != "windows" {
```
right after 
```
	platform := context.GlobalString("platform")
```

hqhq: I'm fine with either, if check platform right after got it from context, it would be like a runtime-tools limitation, while it's actually can be an OCI validation, so check it in `Validator` methods seems reasonable to be.

Plus I moved `CheckPlatform` to the top in the second commit, so we don't do much unnecessary check it platform is wrong.

cmd/oci-runtime-tool/validate.go
```diff
@@ -20,7 +20,8 @@ var bundleValidateCommand = cli.Command{
 	Action: func(context *cli.Context) error {
 		inputPath := context.String("path")
 		hostSpecific := context.GlobalBool("host-specific")
-		v, err := validate.NewValidatorFromPath(inputPath, hostSpecific)
+		platform := context.GlobalString("platform")
```
liangchenye: Add 'platform' to man page as well.

cmd/runtimetest/main.go
```diff
@@ -588,6 +588,8 @@ func validate(context *cli.Context) error {
 		return err
 	}
 
+	platform := context.String("platform")
```
liangchenye: We also need to pass this to `validation_test.go#getDefaultGenerator` and change the Makefile (by adding a PLATFORM for example).

hqhq: Done, but I don't think we use it in generator.

liangchenye: I mean we should change:  
```
g.SetProcessArgs([]string{"/runtimetest"})
```
to
```
g.SetProcessArgs([]string{"/runtimetest", "platform", platform})
```
since you add a platform option to runtimetest command.

liangchenye: Yes, we don't use it anymore in  generator.

cmd/runtimetest/main.go
```diff
@@ -695,7 +697,12 @@ func main() {
 		cli.StringFlag{
 			Name:  "path",
 			Value: ".",
-			Usage: "path to the configuration",
+			Usage: "Path to the configuration",
+		},
+		cli.StringFlag{
+			Name:  "platform",
```
Mashimiao: Should the runtimetest automatically get which platform it is running on? I think it seems there is no need to specify it manunally. What do you think?

cmd/runtimetest/main.go
hqhq: But platform is not a process arg. It should be a generator arg as it's a validate arg.

cmd/runtimetest/main.go
liangchenye: @hqhq how does runc do?

hqhq: @Mashimiao We can, but can we do that in a follow up PR, what we did in this PR looks consistent and we have the default value.

@liangchenye runc doesn't check platform at all, the spec author should be responsible to give runc the right config.

liangchenye:
LGTM

Mashimiao:
LGTM

PR #415, opened 2017-07-06
overwrite the 'platform' value  when 'host-specific' was set

liangchenye:

### DIFF
--- a/man/oci-runtime-tool.1.md
+++ b/man/oci-runtime-tool.1.md
@@ -32,9 +32,6 @@ oci-runtime-tool is a collection of tools for working with the [OCI runtime spec
 **--log-level**=LEVEL
   Log level (panic, fatal, error, warn, info, or debug) (default: "error").
 
-**--platform**
-  Platform the tool targets.
-
 **-v**, **--version**
   Print version information.
 
--- a/cmd/oci-runtime-tool/main.go
+++ b/cmd/oci-runtime-tool/main.go
@@ -22,11 +22,6 @@ func main() {
 			Value: "error",
 			Usage: "Log level (panic, fatal, error, warn, info, or debug)",
 		},
-		cli.StringFlag{
-			Name:  "platform",
-			Value: "linux",
-			Usage: "Platform the tool targets",
-		},
 	}
 
 	app.Commands = []cli.Command{
--- a/cmd/oci-runtime-tool/validate.go
+++ b/cmd/oci-runtime-tool/validate.go
@@ -10,6 +10,7 @@ import (
 
 var bundleValidateFlags = []cli.Flag{
 	cli.StringFlag{Name: "path", Value: ".", Usage: "path to a bundle"},
+	cli.StringFlag{Name: "platform", Value: "linux", Usage: "platform of the target bundle (linux, windows, solaris)"},
 }
 
 var bundleValidateCommand = cli.Command{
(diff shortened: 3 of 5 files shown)


liangchenye:
I think 'platform' should be moved to 'validate' sub command, but not the global command.

man/oci-runtime-tool.1.md
```diff
@@ -33,7 +33,7 @@ oci-runtime-tool is a collection of tools for working with the [OCI runtime spec
   Log level (panic, fatal, error, warn, info, or debug) (default: "error").
 
 **--platform**
-  Platform the tool targets.
+  Platform the tool targets. It will be overwritten by the host platform if '--host-specific' was set.
```
Mashimiao: Should we clearly says valid values and default value in manpage?

liangchenye: You are right, updating on it.

liangchenye:
updated, PTAL @Mashimiao

Mashimiao:
LGTM

Mashimiao:
ping @hqhq  @mrunalp

hqhq:
LGTM

wking:
On Thu, Jul 06, 2017 at 03:37:31AM +0000, 梁辰晔 (Liang Chenye) wrote:
> I think 'platform' should be moved to 'validate' sub command, but
> not the global command.

I think we should move it back to being a global option.  generate.New
does not currently take a platform argument [1] and it assumes Linux
[2,3], but both of those seem like oversights to me.  I'd certainly
expect future users to be interested in generating new Solaris and
Windows configs, especially if config-generation is what we use for
runtime testing [4].

[1]: https://github.com/opencontainers/runtime-tools/blob/0d35fc8437ec8292cd7065e332000c584f7024c5/generate/generate.go#L33-L34
[2]: https://github.com/opencontainers/runtime-tools/blob/0d35fc8437ec8292cd7065e332000c584f7024c5/generate/generate.go#L52
[3]: https://github.com/opencontainers/runtime-tools/blob/0d35fc8437ec8292cd7065e332000c584f7024c5/generate/generate.go#L181
[4]: https://github.com/opencontainers/runtime-tools/blob/0d35fc8437ec8292cd7065e332000c584f7024c5/validation/validation_test.go#L80

liangchenye:
I removed it for 'generate command' for two reasons:
1. I think it is OK for a config file to have both 'Window specific config' or 'Linux specific config'.
    The runtime-spec does not forbid it.
2. I'm afraid it will be too complicated to have 'platform/template/other single option' in 'generate command'.
    If user want to have 'Windows' configs, I suggest they should:
   1) use 'Windows template'
   2) add 'Windows unique' options so the 'generate' will add 'Window Specific Config' automatically.

wking:
On Tue, Jul 11, 2017 at 01:26:25PM +0000, 梁辰晔 (Liang Chenye) wrote:
> 1. I think it is OK for a config file to have both 'Window specific
>    config' or 'Linux specific config'.  The runtime-spec does not
>    forbid it.

I agree that this is legal (with rc6).  I'm not convinced it would
make a useful default template.

> 2. I'm afraid it will be too complicated to have
>    'platform/template/other single option' in 'generate command'.
>    If user want to have 'Windows' configs, I suggest they should:
>    1) use 'Windows template'

Then why provide generate.New at all?  Just provide NewFromSpec,
NewFromFile, and NewFromTemplate, and let all users supply their own
template.

>    2) add 'Windows unique' options so the 'generate' will add 'Window
>       Specific Config' automatically.

The default template currently sets process.user [1], but that's
optional on Windows [2].  Would folks generating Windows configs that
did not supply their own template have to set a --process-user-remove
option or some such?

Would they have to explicitly set process.cwd to override the
POSIX-centric default value [3]?  Even in #379 (as it currently stands
with da5029b7), the process.cwd default is POSIX-centric [4].

[1]: https://github.com/opencontainers/runtime-tools/blob/15ec1df3f6607f4223ab3915547c184cf60a30dd/generate/generate.go#L43
[2]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc6/config.md#L311
[3]: https://github.com/opencontainers/runtime-tools/blob/15ec1df3f6607f4223ab3915547c184cf60a30dd/generate/generate.go#L51
[4]: https://github.com/opencontainers/runtime-tools/pull/379/files#diff-ed9e41f69ceb4f9d141e2fe7a9da68a4R42

PR #417, opened 2017-07-06
add version file

Mashimiao:
It will be more convenient to control version with version package.


liangchenye:
I think in the long term, we can just reuse the version in 'runtime-spec/specs-go'.

Mashimiao:
I don't think I can agree.  Though runtime-tools depends on runtime-spec, they are two separate projects.

version/version.go
```diff
@@ -0,0 +1,32 @@
...
+
+package version
+
+import "fmt"
+
+const (
+	// VersionMajor is for an API incompatible changes
+	VersionMajor = 0
+	// VersionMinor is for functionality in a backwards-compatible manner
+	VersionMinor = 5
```
wking: Why `0.5.0-dev`?  As far as I can tell, there are still no runtime-tools releases, and [SemVer suggests starting with 0.1.0][1].  If we want SemVer, we should also document which API we're using to measure compatibility.  Is it the command line API?  The Go package APIs?  Both?  Without docs about which API is being validated, I'd rather have this package just declare an opaque string to match the Git tag, possibly with a `-dev` suffix between tagged commits.  For example:

```go
var Version = "0.1.0"
```

for the commit tagged 0.1.0,

```go
var Version = "0.1.0-dev"
```

for the commit after the 0.1.0 tag, and something like

```go
var Version = "unreleased"
```

for the current master (since we have no tags or releases at the moment).

[1]: http://semver.org/spec/v2.0.0.html#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase

Mashimiao: The comment in version file seems not so good, try to use VERSION file

Makefile
```diff
@@ -4,11 +4,12 @@ BINDIR ?= $(DESTDIR)/usr/bin
 BUILDTAGS=
 RUNTIME ?= runc
 COMMIT=$(shell git rev-parse HEAD 2> /dev/null || true)
+VERSION := ${shell cat ./VERSION}
 
 all: tool runtimetest
 
 tool:
-	go build -tags "$(BUILDTAGS)" -ldflags "-X main.gitCommit=${COMMIT}" -o oci-runtime-tool ./cmd/oci-runtime-tool
+	go build -tags "$(BUILDTAGS)" -ldflags "-X main.gitCommit=${COMMIT} -X main.version=${VERSION}" -o oci-runtime-tool ./cmd/oci-runtime-tool
```
wking: With this approach there is no way for a Go consumer to figure out which version of the runtime-tools libraries they were built with.  But maybe all serious Go consumers will vendor runtime-tools, so they'll have access to that version via their vendor-pinning file?

q384566678:
LGTM
Convenient version update.

liangchenye:
LGTM

PR #420, opened 2017-07-06
Add go version requirement

hqhq:
According to our .travis.yml, we only support go 1.7 and
above, I think we should explicitly claim that.

As go don't use GO15VENDOREXPERIMENT since 1.7, we can
remove that from Makefile as well.

Signed-off-by: Qiang Huang <h.huangqiang@huawei.com>


liangchenye:
LGTM

Mashimiao:
LGTM

PR #422, opened 2017-07-10
cmd/oci-runtime-tool/main.go: DRYer app.Version

wking:
Instead of hard-coding the current version in two locations, set it once and append the commit information if we have it.  After this change:

    $ git grep 0.0.1 | grep -v 'vendor\|Godeps'
    cmd/oci-runtime-tool/main.go:   app.Version = "0.0.1"
    cmd/runtimetest/main.go:        app.Version = "0.0.1"
    validate/validate_test.go:              {"0.0.1", false},

which is the best we can get without setting a `gitTag` `ldflag` or some such.

### DIFF
--- a/cmd/oci-runtime-tool/main.go
+++ b/cmd/oci-runtime-tool/main.go
@@ -15,10 +15,9 @@ var gitCommit = ""
 func main() {
 	app := cli.NewApp()
 	app.Name = "oci-runtime-tool"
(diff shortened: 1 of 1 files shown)


wking:
Closing in favor of #417.  I'll re-open if #417 is rejected or ends up in some other crazy state ;).

PR #418, opened 2017-07-06
fix bug caused by PR #371

Mashimiao:
fix problem caused by merged #371 

Signed-off-by: Ma Shimiao <mashimiao.fnst@cn.fujitsu.com>

PR #371, opened 2017-04-24
generate: rectify the arguments

q384566678:
Since in https://github.com/opencontainers/runtime-tools/issues/343#issuecomment-286005046 @Kxuan  put forward the idea, I submitted this PR.

Signed-off-by: zhouhao <zhouhao@cn.fujitsu.com>

PR #423, opened 2017-07-10
man/oci-runtime-tool-generate.1: Fix --mount-label and --process-selinux-label

wking:
In #371, the old `--mount-label` and `--selinux-label` became `--linux-mount-label` and `--linux-selinux-label`.  However, the `--mount-label` use in this example wasn't updated and the `--selinux-label` use was incorrectly updated to `--process-selinux-label`.  This commit brings the example back in line with the program source.

I've also added backticks around `--linux-selinux-label` to get better formatting for the [inline code][1].

There are also #371 fixups in flight with #418, but this PR is touching a different file.

[1]: https://daringfireball.net/projects/markdown/syntax#code

### DIFF
--- a/man/oci-runtime-tool-generate.1.md
+++ b/man/oci-runtime-tool-generate.1.md
@@ -374,9 +374,9 @@ colon:
 
 ## Using SELinux
 
-You can use SELinux to add security to the container.  You must specify the process label to run the init process inside of the container using the --process-selinux-label.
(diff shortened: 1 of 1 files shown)


Mashimiao:
LGTM    @liangchenye  @mrunalp  @hqhq

liangchenye:
LGTM

PR #425, opened 2017-07-11
Update rc6

Mashimiao:
updated to spec v1.0.0-rc6 and fix changes based on it.

### DIFF
--- a/vendor/github.com/opencontainers/runtime-spec/specs-go/config.go
+++ b/vendor/github.com/opencontainers/runtime-spec/specs-go/config.go
@@ -6,26 +6,24 @@ import "os"
 type Spec struct {
 	// Version of the Open Container Runtime Specification with which the bundle complies.
 	Version string `json:"ociVersion"`
-	// Platform specifies the configuration's target platform.
-	Platform Platform `json:"platform"`
 	// Process configures the container process.
-	Process Process `json:"process"`
+	Process *Process `json:"process,omitempty"`
 	// Root configures the container's root filesystem.
-	Root Root `json:"root"`
+	Root *Root `json:"root,omitempty"`
 	// Hostname configures the container's hostname.
 	Hostname string `json:"hostname,omitempty"`
 	// Mounts configures additional mounts (on top of Root).
 	Mounts []Mount `json:"mounts,omitempty"`
 	// Hooks configures callbacks for container lifecycle events.
-	Hooks *Hooks `json:"hooks,omitempty"`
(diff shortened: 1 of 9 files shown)


vendor/github.com/opencontainers/runtime-spec/specs-go/config.go
```diff
@@ -6,26 +6,24 @@ import "os"
 type Spec struct {
 	// Version of the Open Container Runtime Specification with which the bundle complies.
 	Version string `json:"ociVersion"`
-	// Platform specifies the configuration's target platform.
-	Platform Platform `json:"platform"`
 	// Process configures the container process.
-	Process Process `json:"process"`
+	Process *Process `json:"process,omitempty"`
```
wking: I think this change will mean you'll need an `initProcess` function and some `nil` checks (e.g. [here][1]).  You'll also need similar changes for `Root` → `*Root`.

[1]: https://github.com/opencontainers/runtime-tools/blob/0d35fc8437ec8292cd7065e332000c584f7024c5/validate/validate.go#L214-L215

vendor/github.com/opencontainers/runtime-spec/specs-go/config.go
```diff
@@ -540,14 +550,21 @@ const (
 type LinuxSeccompArg struct {
 	Index    uint                 `json:"index"`
 	Value    uint64               `json:"value"`
-	ValueTwo uint64               `json:"valueTwo"`
+	ValueTwo uint64               `json:"valueTwo,omiempty"`
```
wking: You'll probably want to use a runtime-spec commit that includes opencontainers/runtime-spec#884 (which didn't make it into rc6).  We'll have to replace [the version comparison][1] and [generation default][2] with a hard-coded version, probably by defining our own target-spec-version variable in a local version package.  But because runtime-spec versions do not cover their Go library, I think we'd have to do that regardless.

[1]: https://github.com/opencontainers/runtime-tools/blob/0d35fc8437ec8292cd7065e332000c584f7024c5/validate/validate.go#L164-L165
[2]: https://github.com/opencontainers/runtime-tools/blob/0d35fc8437ec8292cd7065e332000c584f7024c5/generate/generate.go#L36

cyphar:
/ping @opencontainers/runtime-tools-maintainers? This is quite important for tools like `umoci` that require the generate library and want to match the OCI 1.0 release (which is next week).

vbatts:
LGTM

mrunalp:
LGTM

cyphar:
❤️  Thanks for the fast reponse. 😸

PR #254, opened 2016-10-21
generate: update rootfs releated options

Mashimiao:
Signed-off-by: Ma Shimiao mashimiao.fnst@cn.fujitsu.com

After updated, options of rootfs will be more human-readable and easy to understand

PR #426, opened 2017-07-11
*: Fix --bind -> --mount-bind, etc.

wking:
Which was missed in #371.

Also fix:

* `--cwd` → `--process-cwd` (missed in #371)
* `--rootfs` → `--rootfs-path` (missed in #254)


mrunalp:
LGTM

Mashimiao:
LGTM

PR #428, opened 2017-07-17
Rename CheckRootfsPath and add Root check for hyper-v

Mashimiao:
1. CheckRootfsPath does not only include Path check, we should rename it.
2. For Hyper-V container, Root must not be set.

### DIFF
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -96,7 +96,7 @@ func NewValidatorFromPath(bundlePath string, hostSpecific bool, platform string)
 // CheckAll checks all parts of runtime bundle
 func (v *Validator) CheckAll() (msgs []string) {
 	msgs = append(msgs, v.CheckPlatform()...)
-	msgs = append(msgs, v.CheckRootfs()...)
+	msgs = append(msgs, v.CheckRoot()...)
 	msgs = append(msgs, v.CheckMandatoryFields()...)
 	msgs = append(msgs, v.CheckSemVer()...)
 	msgs = append(msgs, v.CheckMounts()...)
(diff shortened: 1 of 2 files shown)


mrunalp:
LGTM

validate/validate.go
```diff
@@ -96,7 +96,11 @@ func NewValidatorFromPath(bundlePath string, hostSpecific bool, platform string)
 // CheckAll checks all parts of runtime bundle
 func (v *Validator) CheckAll() (msgs []string) {
 	msgs = append(msgs, v.CheckPlatform()...)
-	msgs = append(msgs, v.CheckRootfsPath()...)
+	if v.platform == "windows" && v.spec.Windows.HyperV != nil {
+		msgs = append(msgs, fmt.Sprintf("for Hyper-V containers, Root must not be set"))
```
wking: I think we're missing a check here, and should be using:

```go
if v.platform == "windows" && v.spec.Windows.HyperV != nil {
  if v.spec.Root != nil {
    msgs = append(msgs, fmt.Sprintf("for Hyper-V containers, Root must not be set"))
  }
} else { …
```

Otherwise we'll be erroring on all Hyper-V configs, not just those that set `root`.

validate/validate.go
```diff
@@ -96,7 +96,11 @@ func NewValidatorFromPath(bundlePath string, hostSpecific bool, platform string)
...
+	} else {
+		msgs = append(msgs, v.CheckRootfs()...)
```
wking: You probably also want to check for an unset `root`:

```go
} else if (v.spec.Root == nil) {
  msgs = append(msgs, fmt.Sprintf("for non-Hyper-V containers, root is required"))
} else {
  msgs = append(msgs, v.CheckRootfs()...)
}
```

which will give us handling similar to that proposed for `process` in #433.

validate/validate.go
Mashimiao: good catch! Thanks

Mashimiao:
updated. @opencontainers/runtime-tools-maintainers  PTAL

validate/validate.go
```diff
@@ -109,9 +109,20 @@ func (v *Validator) CheckAll() (msgs []string) {
 	return
 }
 
-// CheckRootfsPath checks status of v.spec.Root.Path
-func (v *Validator) CheckRootfsPath() (msgs []string) {
-	logrus.Debugf("check rootfs path")
+// CheckRootfs checks status of v.spec.Root
+func (v *Validator) CheckRootfs() (msgs []string) {
+	logrus.Debugf("check rootfs")
```
wking: `rootfs` is the value that the spec recommends for POSIX `root.path`.  But since this method is also checking `root.readonly`, I'd recommend using `CheckRoot` instead of `CheckRootfs`, `"check root"` instead of `"check rootfs"`, etc.

q384566678:
LGTM

liangchenye:
LGTM

PR #430, opened 2017-07-17
man: delete extra content

q384566678:
It has changed in https://github.com/opencontainers/runtime-spec/pull/783.

### DIFF
--- a/man/oci-runtime-tool-generate.1.md
+++ b/man/oci-runtime-tool-generate.1.md
@@ -131,7 +131,7 @@ read the configuration from `config.json`.
   Sets the total memory limit (memory + swap) in bytes.
 
 **--linux-mem-swappiness**=MEMSWAPPINESS
-  Sets the swappiness of how the kernel will swap memory pages (Range from 0 to 100).
+  Sets the swappiness of how the kernel will swap memory pages.
 
 **--linux-mems**=MEMS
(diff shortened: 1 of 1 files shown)


Mashimiao:
I think we should keep it. 
Spec doesn't contain kernel limit, because in case of kernel limit changes, spec has to change. That's not spec wants.
But in fact, when run containers on Linux, we should follow it's limit. So, I think we should keep this limit for users to generate their container configuration.

man/oci-runtime-tool-generate.1.md
```diff
@@ -131,7 +131,7 @@ read the configuration from `config.json`.
   Sets the total memory limit (memory + swap) in bytes.
 
 **--linux-mem-swappiness**=MEMSWAPPINESS
-  Sets the swappiness of how the kernel will swap memory pages (Range from 0 to 100).
+  Sets the swappiness of how the kernel will swap memory pages.
```
wking: This limit is [still in the spec][1].  I'd [argued for keeping it out of the spec][2], but did not succeed.  So I think we want to close this PR and keep the parenthetical you're removing.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc6/config-linux.md#L285
[2]: https://github.com/opencontainers/runtime-spec/pull/876#discussion_r123856763

PR #431, opened 2017-07-19
Bump up runtime-spec dependency to v1.0.0

mrunalp:
@opencontainers/runtime-tools-maintainers PTAL
Signed-off-by: Mrunal Patel <mrunalp@gmail.com>

PR #432, opened 2017-07-19
vendor: update to runtime-spec@v1.0.0

cyphar:
There was a very late post-v1.0.0-rc6 hotfix that broke the build
between the two versions. It's not a semantic change, a field was moved
to a sub-struct.


wking:
Dup of #431?

cyphar:
Heh, yeah. Looks like people are racing to bump everything, I'm currently about to bump `umoci` and hit this.

PR #433, opened 2017-07-24
validate: Add non-empty checks 

q384566678:
Signed-off-by: zhouhao zhouhao@cn.fujitsu.com

### DIFF
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -102,9 +102,7 @@ func (v *Validator) CheckAll() (msgs []string) {
 	msgs = append(msgs, v.CheckMounts()...)
 	msgs = append(msgs, v.CheckProcess()...)
 	msgs = append(msgs, v.CheckHooks()...)
-	if v.spec.Linux != nil {
-		msgs = append(msgs, v.CheckLinux()...)
-	}
+	msgs = append(msgs, v.CheckLinux()...)
 
 	return
 }
@@ -210,6 +208,10 @@ func checkEventHooks(hookType string, hooks []rspec.Hook, hostSpecific bool) (ms
 // CheckProcess checks v.spec.Process
 func (v *Validator) CheckProcess() (msgs []string) {
 	logrus.Debugf("check process")
+
+	if v.spec.Process == nil {
+		return
+	}
 
 	process := v.spec.Process
 	if !filepath.IsAbs(process.Cwd) {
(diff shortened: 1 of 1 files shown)


validate/validate.go
```diff
@@ -96,11 +96,15 @@ func NewValidatorFromPath(bundlePath string, hostSpecific bool, platform string)
 // CheckAll checks all parts of runtime bundle
 func (v *Validator) CheckAll() (msgs []string) {
 	msgs = append(msgs, v.CheckPlatform()...)
-	msgs = append(msgs, v.CheckRootfsPath()...)
+	if v.spec.Root != nil {
```
Mashimiao: This is not a perfect fix, should fix as in #428

q384566678: ok. I'll delet it , and please fix in #428 .

Mashimiao:
LGTM

validate/validate.go
```diff
@@ -100,7 +100,9 @@ func (v *Validator) CheckAll() (msgs []string) {
 	msgs = append(msgs, v.CheckMandatoryFields()...)
 	msgs = append(msgs, v.CheckSemVer()...)
 	msgs = append(msgs, v.CheckMounts()...)
-	msgs = append(msgs, v.CheckProcess()...)
+	if v.spec.Process != nil {
+		msgs = append(msgs, v.CheckProcess()...)
```
wking: `Hooks` is a similar pointer, and we're currently handling that by [unconditionally adding `CheckHooks`][1] and then [guarding against the `nil` inside `CheckHooks`][2].  I don't have a strong preference about whether the nil guard is inside `CheckAll` or the more specific `Check*` method, but I think we want to stay consistent.

[1]: https://github.com/opencontainers/runtime-tools/blob/2d270b8764c02228eeb13e36f076f5ce6f2e3591/validate/validate.go#L104
[2]: https://github.com/opencontainers/runtime-tools/blob/2d270b8764c02228eeb13e36f076f5ce6f2e3591/validate/validate.go#L175

Mashimiao: There are many places are not consistent, we can fix this in a follow PR.

liangchenye: @Mashimiao I think we can do it in this PR. Since the PR does not do anything else.
@q384566678 How about it?

q384566678: ok, I'll do it.

q384566678: updated.

Mashimiao:
please squash them into one patch. there is no need to work with two patches

liangchenye:
LGTM

Mashimiao:
LGTM

PR #437, opened 2017-07-28
error: Pull the RFC 2119 error representation into its own package

wking:
As discussed [here][1].  This makes it easier for other projects (e.g. image-tools) to use the same tooling if they want.  Some components of the old `validate/error.go` were runtime-spec-specific (e.g. the reference template and `ociErrors`), so they've stayed in the validate package.

I've also expanded `NewError` to take an explicit version (as requested [here][2]).  That allows us to link to the proper spec even if we're capable of validating several spec versions (e.g. 1.0 and 1.1 configurations or runtimes).

[1]: https://github.com/opencontainers/runtime-tools/pull/354#discussion_r109216291
[2]: https://github.com/opencontainers/runtime-tools/pull/354#discussion_r108980617

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -22,7 +22,8 @@ import (
 	"github.com/urfave/cli"
 
 	"github.com/opencontainers/runtime-tools/cmd/runtimetest/mount"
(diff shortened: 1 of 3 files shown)


liangchenye:
LGTM

Mashimiao:
LGTM

PR #441, opened 2017-08-03
generate: solve conflicting options problem

Mashimiao:
In my option, there is not any other good ways to solve conflicting
options problem except adding note for warning.


q384566678:
SGTM, `process-cap-drop-all` should also be modified.

Mashimiao:
updated, thanks

q384566678:
LGTM

q384566678:
LGTM

liangchenye:
LGTM

PR #445, opened 2017-08-17
filepath: Add a stand-alone package for explicit-OS path logic

wking:
As discussed [here][0] and later, and based on the Gist I'd created for that comment.

Go's path/filepath has lots of useful path operations, but there is a compile-time decision to select only the logic that applies to your `$GOOS`.  That makes it hard to validate a Windows config from a Linux host (or vice versa) because Go's builtin tools won't tell you whether a path is absolute on the target platform (just whether the path is absolute on *your* platform).  This commit adds a new package to do the same sorts of things but with an explicit OS argument.

In some cases, there's also an explicit workding directory argument. For example, Go's [`Abs`][1] has:

> If the path is not absolute it will be joined with the current working directory to turn it into an absolute path.

but that doesn't make sense for a cross-platform `Abs` call because the real current working directory will be for the wrong platform.  Instead, cross-platform calls to `Abs` and similar should fake a working directory as if they were being called from the other platform.

The Windows implementation is not very complete, with `IsAbs` definitely missing a lot of stuff; `Abs`, `Clean`, and `IsAncestor` probably missing stuff; and a lack of Windows-path tests.  But the current tools are broken for validating Windows configs anyway, so I've left completing Windows support to future work.

Besides adding the new package, I updated the config validation to use the new package where appropriate.  For example, checks for absolute hook paths (based on [this][2]) now appropriately account for the target platform (although `Abs` has limited Windows support at the moment, as mentioned above).

There are still a number of config validation checks that use Go's stock filepath, because they're based around actual filesystem access (e.g. reading `config.json` off the disk, asserting that `root.path` exists on the disk, etc.).  Some of those will need logic to convert between path platforms (which I'm leaving to future work).  For example, if `root.path` is formed for another platform, then:

* If `root.path` is absolute (on the target platform), there's no way to check whether root.path exists inside the bundle.
* If `root.path` is relative, we should be converting it from the target platform to the host platform before joining it to our bundle path.  For example, with a Windows bundle in `/foo` on a Linux host where `root.path` is `bar\baz`, then runtime-tools should be checking for the root directory in `/foo/bar/baz`, not in `/foo/bar\baz`.  The `root.path` example is somewhat guarded by the [bundle requirement for siblinghood][3], but I'd rather enforce siblinghood independently from existence, since the spec has [separate][3] [requirements][4] for both.

[0]: https://github.com/opencontainers/runtime-tools/pull/256#issuecomment-258365559
[1]: https://golang.org/pkg/path/filepath/#Abs
[2]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config.md#L375
[3]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/bundle.md#L17-L19
[4]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config.md#L43

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,6 +53,6 @@ test: .gofmt .govet .golint .gotest
 .golint:
 	golint -set_exit_status $(PACKAGES)
 
-UTDIRS = ./validate/...
+UTDIRS = ./filepath/... ./validate/...
 .gotest:
 	go test $(UTDIRS)
--- a/filepath/abs.go
+++ b/filepath/abs.go
@@ -1,0 +1,52 @@
+package filepath
+
+import (
+	"errors"
+	"regexp"
+	"strings"
+)
+
+var windowsAbs = regexp.MustCompile(`^[a-zA-Z]:\\.*$`)
+
+// Abs is a version of path/filepath's Abs with an explicit operating
+// system and current working directory.
+func Abs(os, path, cwd string) (_ string, err error) {
+	if os == "windows" {
+		return "", errors.New("Abs() does not support windows yet")
+	}
+	if IsAbs(os, path) {
+		return Clean(os, path), nil
+	}
+	return Clean(os, Join(os, cwd, path)), nil
+}
+
+// IsAbs is a version of path/filepath's IsAbs with an explicit
+// operating system.
+func IsAbs(os, path string) bool {
+	if os == "windows" {
+		// FIXME: copy hideous logic from Go's
+		// src/path/filepath/path_windows.go into somewhere where we can
+		// put 3-clause BSD licensed code.
+		return windowsAbs.MatchString(path)
+	}
+	sep := Separator(os)
+
+	// POSIX has [1]:
+	//
+	// > If a pathname begins with two successive <slash> characters,
+	// > the first component following the leading <slash> characters
+	// > may be interpreted in an implementation-defined manner,
+	// > although more than two leading <slash> characters shall be
+	// > treated as a single <slash> character.
+	//
+	// And Boost treats // as non-absolute [2], but Linux [3,4], Python
+	// [5] and Go [6] all treat // as absolute.
+	//
+	// [1]: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap04.html#tag_04_13
+	// [2]: https://github.com/boostorg/filesystem/blob/boost-1.64.0/test/path_test.cpp#L861
+	// [3]: http://man7.org/linux/man-pages/man7/path_resolution.7.html
+	// [4]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/filesystems/path-lookup.md?h=v4.12#n41
+	// [5]: https://github.com/python/cpython/blob/v3.6.1/Lib/posixpath.py#L64-L66
+	// [6]: https://go.googlesource.com/go/+/go1.8.3/src/path/path.go#199
+	return strings.HasPrefix(path, string(sep))
+}
--- a/filepath/abs_test.go
+++ b/filepath/abs_test.go
@@ -1,0 +1,171 @@
+package filepath
+
+import (
+	"fmt"
+	"testing"
+)
+
+func TestAbs(t *testing.T) {
+	for _, test := range []struct {
+		os       string
+		path     string
+		cwd      string
+		expected string
+	}{
+		{
+			os:       "linux",
+			path:     "/",
+			cwd:      "/cwd",
+			expected: "/",
+		},
+		{
+			os:       "linux",
+			path:     "/a",
+			cwd:      "/cwd",
+			expected: "/a",
+		},
+		{
+			os:       "linux",
+			path:     "/a/",
+			cwd:      "/cwd",
+			expected: "/a",
+		},
+		{
+			os:       "linux",
+			path:     "//a",
+			cwd:      "/cwd",
+			expected: "/a",
+		},
+		{
+			os:       "linux",
+			path:     ".",
+			cwd:      "/cwd",
+			expected: "/cwd",
+		},
+		{
+			os:       "linux",
+			path:     "./c",
+			cwd:      "/a/b",
+			expected: "/a/b/c",
+		},
+		{
+			os:       "linux",
+			path:     ".//c",
+			cwd:      "/a/b",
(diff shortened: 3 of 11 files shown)


liangchenye:
LGTM 
I tested it locally, it works. But can you add `filepath` to Makefile so travis will test it automatically? @wking 

The `validation` directory is not used for unit test, we had to add directory like `filepath` to Makefile by hand.

wking:
On Thu, Aug 24, 2017 at 02:26:35PM +0000, 梁辰晔 (Liang Chenye) wrote:
> But can you add `filepath` to Makefile so travis will test it
> automatically?

Done with 8af7e86 → df8d285.

liangchenye:
LGTM 
@mrunalp @Mashimiao @q384566678

Mashimiao:
I think this cause a regression. Before this PR we can check whether windows mounts are nested, but now we can't any more.

wking:
On Fri, Aug 25, 2017 at 09:36:45AM +0000, Ma Shimiao wrote:
> I think this cause a regression. Before this PR we can check whether
> windows mounts are nested, but now we can't any more.

I'm not sure master has working Windows nested checks.  For example,
[1] is broken on Windows, where / is not the root path.  And [2] is
broken on Windows unless you're compiled for Windows, because Go's
non-Windows filepath.Clean may mangle Windows paths.  If you're
worried about the validating-Windows-configs-from-Windows use case
(but not the validating-Windows-configs-from-other-OSes use case), we
could, as a temporary measure, have my IsAncestor fall back to Go's
IsAncestor if os matches GOOS.  Would that be sufficient for now?

[1]: https://github.com/opencontainers/runtime-tools/blob/3db644c31789a6793c11a1c975312c489befab3d/validate/validate.go#L846
[2]: https://github.com/opencontainers/runtime-tools/blob/3db644c31789a6793c11a1c975312c489befab3d/validate/validate.go#L857

Mashimiao:
I don't think tha's IsAncestor's problem, it's Abs()'s problem.  I think we can check whether a windows path is absolute or not as https://github.com/opencontainers/runtime-tools/blob/master/validate/validate.go#L843

wking:
On Tue, Aug 29, 2017 at 11:30:44PM -0700, Ma Shimiao wrote:
> I don't think tha's IsAncestor's problem, it's Abs()'s problem.  I
> think we can check whether a windows path is absolute or not as
> https://github.com/opencontainers/runtime-tools/blob/master/validate/validate.go#L843

This link currently points at a rspec.UTSNamespace line, and no longer
points at anything filepath-like.  Looking through master's history, I
think you made the comment after #448 landed but before #374 landed,
which would mean you ment to reference [1]:

  func nestedValid(os, pathA, pathB string) (bool, error) {
    if pathA == pathB {
      return false, nil
    }
    …

In this PR, I'm replacing nestedValid with IsAncestor [2], and
IsAncestor isn't counting matching paths as ancestors [3].  That
matches the code you reference, where matching paths are not counted
as nested.  So I don't see an Abs issue there, or even a
platform-specific issue there.  Can you be more specific about the
issue you're thinking about?

We've also had previous discussion pointing out the ambiguity of the
spec's current “nested” condition [3,4] (although our interpretation
of that is not changing in this PR).  I've also had two off-list
threads with @RobDolinMS on this point, but haven't followed up on
those either.  @RobDolinMS, if it helps, the tails of those threads
are [5,6].

[1]: https://github.com/opencontainers/runtime-tools/blob/3db644c31789a6793c11a1c975312c489befab3d/validate/validate.go#L843
[2]: https://github.com/opencontainers/runtime-tools/pull/445/commits/df8d285d099f712d4e4fa1b1d45793ecff04db57#diff-8351286f57eb81e245c9c99c07f3b34fL427
[3]: https://github.com/opencontainers/runtime-tools/pull/445/commits/df8d285d099f712d4e4fa1b1d45793ecff04db57#diff-accfd27539832ca1c0a3950cf8403c80R12
[4]: https://github.com/opencontainers/runtime-tools/pull/256#discussion_r84830662
[5]: Message-ID: <20161026050002.GB7420@odin.tremily.us>
     Subject: Re: Helping close CLI (#513)
     Date: Tue, 25 Oct 2016 22:00:02 -0700
[6]: Message-ID: <20161026213320.GA28698@odin.tremily.us>
     Subject: Date: Wed, 26 Oct 2016 14:33:20 -0700
     Date: Wed, 26 Oct 2016 14:33:20 -0700

Mashimiao:
On 08/31/2017 05:23 AM, W. Trevor King wrote:
> points at anything filepath-like. Looking through master's history, I
> think you made the comment after #448 landed but before #374 landed,
> which would mean you ment to reference [1]:

No, it should be here: https://github.com/opencontainers/runtime-tools/blob/3db644c31789a6793c11a1c975312c489befab3d/validate/validate.go#L826

wking:
> No, it should be here...

Ah, so maybe copy part of that into `IsAbs`, and then replace our current `pathValid` calls with `IsAbs`?

Mashimiao:
On 08/31/2017 09:45 AM, W. Trevor King wrote:
> Ah, so maybe copy part of that into |IsAbs|, and then replace our current |pathValid| calls with |IsAbs|?

Yeah, currently pathValid mainly check whether the path is absolute for windows or linux.
And should add more tests of Clean() for windows path.

wking:
On Wed, Aug 30, 2017 at 07:01:22PM -0700, Ma Shimiao wrote:
> Yeah, currently pathValid mainly check whether the path is absolute
> for windows or linux.

I've dropped pathValid with df8d285 → cf64923, which also adds some
IsAbs tests (including for Windows) and adjusts the Windows abs-path
regexp based on some Windows docs.  I've also dropped the [Windows
source check][1], which I don't see support for [in the spec][2].  The
update also rebases around #451.

> And should add more tests of Clean() for windows path.

I'm happy to do that, but I don't have example Windows paths or
expected results to feed in.  Can you list the examples you'd like to
see tested and the cleaned result?

[1]: https://github.com/opencontainers/runtime-tools/pull/256/files#diff-8351286f57eb81e245c9c99c07f3b34fR413
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#mounts

Mashimiao:
LGTM

wking:
@liangchenye, still look good to you?

liangchenye:
I'll review the changes @wking .

liangchenye:
LGTM

Issue #305, opened 2017-01-12
Collect container exit codes across platforms


wking:
I think we're pretty clear that we [want][1] a [subreaper][2] to collect the container exit code when we're validating runC and other runtimes that use local Linux namespacing/cgroups to implement runtime-spec.  However, that approach will not work for runtimes that use VMs, or on Windows, or anywhere that the host cannot see the container process (opencontainers/runtime-spec#459).  Do we have a plan for validating those runtimes?  Are we going to have a validation approach that ignores the container process exit code and looks for a magic token in the container's stdout?  Are we going to adjust the in-flight command line API (opencontainers/runtime-spec#513) to allow for a long-running `create`?  Are we going to land an event operation (opencontainers/runtime-spec#508)?  With a spec 1.0 approaching, we probably want to at least have a strategy for testing these alternative runtimes, even if we don't have all the tooling implemented yet.

[1]: https://github.com/opencontainers/runtime-tools/pull/232#discussion_r82655833
[2]: https://github.com/opencontainers/runtime-tools/pull/285#issuecomment-264515242

wking:
Also tied up in “how do I collect the container exit code?” is “how do I know when the container process has exited?”.  An alternative not mentioned in my brief list above would be to poll `state` and look for a `stopped` status, and then to read out the container process exit code from a (new) state field (`exit-code`?).  That sounds like a pretty ugly user experience compared to opencontainers/runtime-spec#508, but it's probably the easiest-to-implement way to get over this particular hurdle, and I think it would be better than cutting a spec 1.0 without addressing this issue at all.

PR #447, opened 2017-08-24
add 'Runtime' struct to make test easier; add create test

liangchenye:

### DIFF
--- a/validation/validation_test.go
+++ b/validation/validation_test.go
@@ -1,7 +1,6 @@
 package validation
 
 import (
-	"fmt"
 	"io/ioutil"
 	"os"
 	"os/exec"
@@ -9,8 +8,12 @@ import (
 	"testing"
 
 	"github.com/mrunalp/fileutils"
-	"github.com/opencontainers/runtime-tools/generate"
+	rspecs "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/satori/go.uuid"
+	"github.com/stretchr/testify/assert"
+
+	rerr "github.com/opencontainers/runtime-tools/error"
+	"github.com/opencontainers/runtime-tools/generate"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


liangchenye:
Hi @Mashimiao , this is my approach to test lifecycle.
We need more test cases to cover 'create' operation. 
Mark as WIP now.

validation/validation_test.go
```diff
@@ -11,17 +11,21 @@ import (
...
 
 var (
 	runtime = "runc"
 )
 
 func init() {
-	runtime = os.Getenv("RUNTIME")
+	runtimeInEnv := os.Getenv("RUNTIME")
+	if runtimeInEnv != "" {
+		runtime = runtimeInEnv
```
wking: This looks like a non-WIP improvement to me.  Can we move it into its own PR to get it landed more quickly?

validation/validation_test.go
```diff
@@ -11,17 +11,21 @@ import (
...
+	}
 }
 
-func runtimeValidate(runtime string, g *generate.Generator) error {
+func runtimeValidate(runtime string, g *generate.Generator, args []string) error {
```
wking: This is not going to work, because when you test `start`, `state`, `delete` etc. you'll need to call `create` first (in most cases).  `runtimeValidate` contains all the bundle setup and teardown code, so you need a way to call multiple runtime commands inside the test, not just a way to change which single runtime command you call.  Pulling the bundle creation and teardown functionality out into separate functions seems like a more flexible approach, allowing callers to easily do whatever they want in between the two.

validation/validation_test.go
liangchenye: https://github.com/opencontainers/runtime-tools/pull/448 is created.

validation/validation_test.go
liangchenye: Your are right, move 'create/start/state/delete' to 'args' is not enough. 
https://github.com/opencontainers/runtime-tools/pull/391/  @Mashimiao  is the way to prepare the bundle.

liangchenye:
It is similar with @Mashimiao 's https://github.com/opencontainers/runtime-tools/pull/391.
I just add a 'Runtime' struct and add more test cases in 'create' test.

In current testing, `runc` does not follow the lifecycle spec...

liangchenye:
Updated, PTAL @wking @Mashimiao @mrunalp @q384566678

liangchenye:
change to [WIP] . the `stretchr/testify` pkg is not included.
I'll add it in https://github.com/opencontainers/runtime-tools/pull/451

liangchenye:
Updated with #451 rfc code feature.
PTAL @q384566678 @Mashimiao

Mashimiao:
LGTM

q384566678:
LGTM

validation/validation_test.go
```diff
@@ -24,81 +27,106 @@ func init() {
...
-
-	return nil
-}
-
-func getDefaultGenerator() *generate.Generator {
-	g := generate.New()
-	g.SetRootPath(".")
-	g.SetProcessArgs([]string{"/runtimetest"})
-	return &g
+	return r.Start()
```
wking: This is broken, since you've deferred a `delete` call (via `Clean`) [earlier in the function][1].  That doesn't work reliably though, because [`delete` MUST generate an error if the container is not `stopped`][2], and the fact that `start` completes doesn't mean you're `stopped` (you [could also be `running`][3]).  In order for `Clean()` to work reliably, you either need to attempt to send a `KILL` to the container process with [`kill`][4] (although the spec is not clear on what signals are supported, you need something like #321 to see that [`KILL` MUST be supported][5]) or code that waits for the container process to exit on its own (more on this [here][6], in #305, and in most of the other PRs that attempted to split `run` into `create` and`start`, e.g. [here][7] and [here][8]).

[1]: https://github.com/opencontainers/runtime-tools/pull/447/files#diff-b5e8ff579203d652582cf33ed2b39ad9R65
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/runtime.md#delete
[3]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/runtime.md#L21-L22
[4]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/runtime.md#kill
[5]: https://github.com/wking/ocitools-v2/blob/a2725b42d7b312681ed5315cf7e663a13c487445/docs/command-line-interface.md#kill
[6]: https://github.com/wking/ocitools-v2/blob/a2725b42d7b312681ed5315cf7e663a13c487445/docs/command-line-interface.md#container-process-exit
[7]: https://github.com/opencontainers/runtime-tools/pull/232#discussion_r82655833
[8]: https://github.com/opencontainers/runtime-tools/pull/285#issuecomment-264515242

PR #448, opened 2017-08-25
use default runtime to test when the runtime env is not set

liangchenye:


q384566678:
LGTM

Mashimiao:
LGTM

PR #450, opened 2017-08-29
use rfc code in bundle validate

initlove:


initlove:
This is a big change to bundle validation. It uses RFC

PR #451, opened 2017-08-29
use rfc code in bundle validate

liangchenye:


liangchenye:
This is a big change to bundle validation. It uses RFC error.
It helps to sort out "which spec item do we already follow'.

validate/error.go
```diff
@@ -13,22 +14,81 @@ const referenceTemplate = "https://github.com/opencontainers/runtime-spec/blob/v
...
+)
+
 var ociErrors = map[ErrorCode]errorTemplate{
-	DefaultFilesystems: errorTemplate{
-		Level: rfc2119.Should,
-		Reference: func(version string) (reference string, err error) {
-			return fmt.Sprintf(referenceTemplate, version, "config-linux.md#default-filesystems"), nil
-		},
-	},
+	// NonRFCError represents that an error is not a rfc2119 error
```
wking: I don't think we need to duplicate this comment here (you already have it on line 19).

validate/error.go
```diff
@@ -13,22 +14,81 @@ const referenceTemplate = "https://github.com/opencontainers/runtime-spec/blob/v
...
+	ConfigFileExistence:  errorTemplate{Level: rfc2119.Must, Reference: containerFormatRef},
+	ArtifactsInSingleDir: errorTemplate{Level: rfc2119.Must, Reference: containerFormatRef},
+
+	// Config.md
+	// Specification Version
+	SpecVersion: errorTemplate{Level: rfc2119.Must, Reference: specVersionRef},
+	// Root
+	RootOnNonHyperV: errorTemplate{Level: rfc2119.Required, Reference: rootRef},
+	RootOnHyperV:    errorTemplate{Level: rfc2119.Must, Reference: rootRef},
+	// TODO
```
wking: Probably expand this comment to:

> TODO: Add tests for these spec conditions

or some such.

validate/error.go
```diff
@@ -13,22 +14,81 @@ const referenceTemplate = "https://github.com/opencontainers/runtime-spec/blob/v
...
+	PathFormatOnWindows: errorTemplate{Level: rfc2119.Must, Reference: rootRef},
+	PathName:            errorTemplate{Level: rfc2119.Should, Reference: rootRef},
+	PathExistence:       errorTemplate{Level: rfc2119.Must, Reference: rootRef},
+	ReadonlyFilesystem:  errorTemplate{Level: rfc2119.Must, Reference: rootRef},
```
wking: I think you intend this for the spec's:

> * **`readonly`** (bool, OPTIONAL) If true then the root filesystem MUST be read-only inside the container, defaults to false.

But that is a runtime requirement which is tested by `validation` and `runtimetest`.  It doesn't belong in `validate`, which is about config/bundle validation.

validate/error.go
```diff
@@ -48,5 +108,30 @@ func NewError(code ErrorCode, msg string, version string) (err error) {
...
 		Reference: reference,
 		Err:       errors.New(msg),
+		ErrCode:   int(code),
+	}
+}
+
+// FindError finds an error from a source error (mulitple error) and
+// returns the error code if founded.
+// If the source error is nil or empty, return NonErr.
+// If the source error is not a multiple error, return NonRFCErr.
```
wking: `NonErr` → `NonError` and `NonRFCErr` → `NonRFCError`.

validate/validate.go
```diff
@@ -139,78 +148,83 @@ func (v *Validator) CheckRoot() (msgs []string) {
...
-		msgs = append(msgs, fmt.Sprintf("Cannot find the root path %q", rootfsPath))
+		errs = multierror.Append(errs,
+			NewError(PathExistence, fmt.Sprintf("Cannot find the root path %q", rootfsPath), rspec.Version))
 	} else if !fi.IsDir() {
-		msgs = append(msgs, fmt.Sprintf("The root path %q is not a directory.", rootfsPath))
+		errs = multierror.Append(errs,
+			NewError(PathExistence, fmt.Sprintf("The root path %q is not a directory.", rootfsPath), rspec.Version))
 	}
 
 	rootParent := filepath.Dir(absRootPath)
```
wking: We will need a guard for the case where `absRootPath` was not initialized because the `Abs` call failed above.  Although that issue predates this PR and can be handled in a separate PR if you like.

validate/validate.go
```diff
@@ -139,78 +148,83 @@ func (v *Validator) CheckRoot() (msgs []string) {
...
 	version := v.spec.Version
 	_, err := semver.Parse(version)
 	if err != nil {
-		msgs = append(msgs, fmt.Sprintf("%q is not valid SemVer: %s", version, err.Error()))
+		errs = multierror.Append(errs,
+			NewError(SpecVersion, fmt.Sprintf("%q is not valid SemVer: %s", version, err.Error()), rspec.Version))
 	}
 	if version != rspec.Version {
-		msgs = append(msgs, fmt.Sprintf("internal error: validate currently only handles version %s, but the supplied configuration targets %s", rspec.Version, version))
+		errs = multierror.Append(errs, fmt.Errorf("internal error: validate currently only handles version %s, but the supplied configuration targets %s", rspec.Version, version))
```
wking: This is clearly an internal error because it is a non-RFC error.  I think we should drop “internal error: ” from the value.

validate/validate.go
```diff
@@ -139,78 +148,83 @@ func (v *Validator) CheckRoot() (msgs []string) {
...
 
 	return
 }
 
-func checkEventHooks(hookType string, hooks []rspec.Hook, hostSpecific bool) (msgs []string) {
+func checkEventHooks(hookType string, hooks []rspec.Hook, hostSpecific bool) (errs error) {
 	for _, hook := range hooks {
 		if !filepath.IsAbs(hook.Path) {
-			msgs = append(msgs, fmt.Sprintf("The %s hook %v: is not absolute path", hookType, hook.Path))
+			errs = multierror.Append(errs, fmt.Errorf("The %s hook %v: is not absolute path", hookType, hook.Path))
```
wking: The MUST for this check is [here][1].  For `poststart` and `poststop` the requirement is indirect, via [this][2] and [this][3].  Perhaps we should expand `Error.Reference` to an array of strings so we can represent trails like that?  Although if we're sticking to Markdown links, [this][4] is as specific as we can get at the moment, and it covers all the lines I linked earlier.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config.md#L375
[2]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config.md#L381
[3]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config.md#L383
[4]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#posix-platform-hooks

validate/validate.go
```diff
@@ -228,17 +242,17 @@ func (v *Validator) CheckProcess() (msgs []string) {
 
 	process := v.spec.Process
 	if !filepath.IsAbs(process.Cwd) {
-		msgs = append(msgs, fmt.Sprintf("cwd %q is not an absolute path", process.Cwd))
+		errs = multierror.Append(errs, fmt.Errorf("cwd %q is not an absolute path", process.Cwd))
```
wking: The MUST for this check is [here][1].

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config.md#L155

validate/validate.go
```diff
@@ -228,17 +242,17 @@ func (v *Validator) CheckProcess() (msgs []string) {
...
 	}
 
 	for _, env := range process.Env {
 		if !envValid(env) {
-			msgs = append(msgs, fmt.Sprintf("env %q should be in the form of 'key=value'. The left hand side must consist solely of letters, digits, and underscores '_'.", env))
+			errs = multierror.Append(errs, fmt.Errorf("env %q should be in the form of 'key=value'. The left hand side must consist solely of letters, digits, and underscores '_'.", env))
```
wking: The spec punts to POSIX [here][1], and [POSIX has][2]:

> These strings have the form name=value; names shall not contain the character '='.

which I read as a MUST-level requirement.  POSIX also has:

> Environment variable names used by the utilities in the Shell and Utilities volume of POSIX.1-2008 consist solely of uppercase letters, digits, and the <underscore> ( '_' ) from the characters defined in Portable Character Set and do not begin with a digit. Other characters may be permitted by an implementation; applications shall tolerate the presence of such names.

which I read as a SHOULD-level requirement for names, because OCI runtimes are not “used by the utilities in the Shell and Utilities volume of POSIX.1-2008”.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config.md#L156
[2]: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08_01

validate/error.go
liangchenye: Yes, it is same with  `DefaultFilesystems`. I prefer to put the error definition of runtime test and bundle test together. Maybe I should also move the `validate/error.go` file to the `error` directory?

validate/validate.go
liangchenye: From the golang document, it seems only a terrible system failure will cause an Abs Error.
I think we can just `return`  when we catch an Abs error. I'll add it in the following update.

validate/validate.go
liangchenye: This PR is quite big. I think we could change other errors in the following PRs.

validate/validate.go
liangchenye: I opened an issue to track this: https://github.com/opencontainers/runtime-tools/issues/452.

liangchenye:
- move validate/error.go to error/runtime_spec.go
- move error/error.go to error/rfc2119.go
- change `msg []string` to error in bundle validation checks
- use rfc runtime spec error in CheckRoot and CheckVersion
- update unittest and add `assert` pkg which is very useful
- fix golint error and other errors reported by @wking 

The next plan:
- replace most previous errors (except inner error) to rfc runtime spec error 
- complete unittest

PTAL @wking @q384566678 @Mashimiao @mrunalp @hqhq

Mashimiao:
LGTM
I'm OK with such changes and look forward to following changes.

q384566678:
LGTM

PR #455, opened 2017-08-30
specerror: Pull runtime-spec-specific error handling into its own package

wking:
With #437, I'd created a package that was completely independent of runtime-spec.  As I pointed out in that PR, this made it possible for image-tools and other projects to reuse the generic RFC 2119 handling (which they care about) without involving the runtime-spec-specific error enumeration and such (which they don't care about).

In #451, some runtime-spec-specific functionality leaked into the error package.  I'd recommended [keeping configuration and runtime requirements separate][1], because you're unlikely to be testing both of those at once.  But @liangchenye [wanted][2] them [collected][3].  And the `NewError` and `FindError` utilities would be the same regardless of target, so that's a good argument for keeping them together.  This commit moves the runtime-spec-specific functionality into a new `specerror` package where both config and runtime validators can share it, but where it won't pollute the generic RFC 2119 `error` package.

I've also changed `NewError` to take an error argument instead of a string message, because creating an error from a string is easy (e.g. with `fmt.Errorf(…)`), and using an error allows you to preserve any additional structured information from a system error (e.g. as returned by `GetMounts`).

The second commit in this PR cleans up some of the generic RFC 2119 package's godocs.

[1]: https://github.com/opencontainers/runtime-tools/pull/451#discussion_r135844806
[2]: https://github.com/opencontainers/runtime-tools/pull/451#discussion_r135953382
[3]: https://github.com/opencontainers/runtime-tools/pull/451#issuecomment-325874457

### DIFF
--- a/validation/validation_test.go
+++ b/validation/validation_test.go
@@ -1,6 +1,7 @@
 package validation
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 	"os/exec"
@@ -12,8 +13,8 @@ import (
 	"github.com/satori/go.uuid"
 	"github.com/stretchr/testify/assert"
 
-	rerr "github.com/opencontainers/runtime-tools/error"
 	"github.com/opencontainers/runtime-tools/generate"
+	"github.com/opencontainers/runtime-tools/specerror"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


validation/validation_test.go
```diff
@@ -114,9 +114,9 @@ func TestValidateCreate(t *testing.T) {
 		errExpected bool
 		err         error
 	}{
-		{"", false, rerr.NewError(rerr.CreateWithID, "'Create' MUST generate an error if the ID is not provided", rspecs.Version)},
-		{containerID, true, rerr.NewError(rerr.CreateNewContainer, "'Create' MUST create a new container", rspecs.Version)},
-		{containerID, false, rerr.NewError(rerr.CreateWithUniqueID, "'Create' MUST generate an error if the ID provided is not unique", rspecs.Version)},
+		{"", false, specerror.NewError(specerror.CreateWithID, "'Create' MUST generate an error if the ID is not provided", rspecs.Version)},
```
Mashimiao: I'm fine with new format of NewError(), but it seems you leave behind here three place. I wonder why there is no error when compiling.

liangchenye: validation_test.go will not run in our CI.

liangchenye:
I'm ok with this change. Only the test case issue need to be fixed. @wking

validation/validation_test.go
wking: > I'm fine with new format of NewError(), but it seems you leave behind here three place.

Oops.  Thanks.  Fixed with c3a8d2c → 7e95c6f.

wking:
I fixed a `go fmt …` error with 7e95c6f → f9152f1.

Mashimiao:
LGTM

q384566678:
LGTM

PR #456, opened 2017-08-30
cmd/runtimetest/main.go: Unify validateMountsExist and validateMountsOrder

wking:
Also increase the error message detail and continue through the remaining mounts instead of breaking on the first missing/misordered mount.  Based on previous discussion [here][1] and [here][2].  With this commit, a configuration like:

```
  "mounts": [
    {
      "destination": "/tmp",
      "type": "tmpfs",
      "source": "none"
    },
    {
      "destination": "/tmp",
      "type": "tmpfs",
      "source": "none"
    },
    {
      "destination": "/dev",
      "type": "devtmpfs",
      "source": "devtmpfs"
    }
  ]
```

and mountinfo like:

```
$ grep -n '/dev \|/tmp ' /proc/self/mountinfo
2:19 17 0:6 / /dev rw,nosuid,relatime - devtmpfs devtmpfs rw,size=10240k,nr_inodes=2043951,mode=755
25:41 17 0:38 / /tmp rw,relatime - tmpfs none rw
```

will generate errors like:

```
* mounts[1] {/tmp tmpfs none []} does not exist
* mounts[2] {/dev devtmpfs devtmpfs []} is system mount 1, while mounts[0] {/tmp tmpfs none []} is system mount 24
```

Grep reports 2 and 25 because it's counting from one, and `runtimetest` reports 1 and 24 because it's counting from zero.

Before this commit, the error was just:

```
* Mounts[1] /tmp is not mounted in order
```

[1]: https://github.com/opencontainers/runtime-tools/pull/444#discussion_r134346130
[2]: https://github.com/opencontainers/runtime-tools/pull/444#issuecomment-324122061

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -576,62 +576,31 @@ func validateGIDMappings(spec *rspec.Spec) error {
 	return validateIDMappings(spec.Linux.GIDMappings, "/proc/self/gid_map", "linux.gidMappings")
 }
 
-func mountMatch(configMount rspec.Mount, sysMount rspec.Mount) error {
-	if filepath.Clean(configMount.Destination) != sysMount.Destination {
-		return fmt.Errorf("mount destination expected: %v, actual: %v", configMount.Destination, sysMount.Destination)
+func mountMatch(configMount rspec.Mount, sysMount *mount.Info) error {
+	sys := rspec.Mount{
+		Destination: sysMount.Mountpoint,
+		Type:        sysMount.Fstype,
+		Source:      sysMount.Source,
 	}
 
-	if configMount.Type != sysMount.Type {
-		return fmt.Errorf("mount %v type expected: %v, actual: %v", configMount.Destination, configMount.Type, sysMount.Type)
+	if filepath.Clean(configMount.Destination) != sys.Destination {
+		return fmt.Errorf("mount destination expected: %v, actual: %v", configMount.Destination, sys.Destination)
 	}
 
-	if filepath.Clean(configMount.Source) != sysMount.Source {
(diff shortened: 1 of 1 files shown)


wking:
@Mashimiao had [expected this unified approach to be long and complicated][1], but it ended up dropping over 20 lines even with the extra bookkeeping for `configSys` and `highestMatchedConfig` (which are used in the improved error messages).

[1]: https://github.com/opencontainers/runtime-tools/pull/444#issuecomment-324198637

Mashimiao:
The changes are good, but I don't think something like `system mount 1` really make sense.  we don't need to care about what it really mounted number, we just need to care about if it is mounted in order.
And `system mount` is not easy understand.
I think `mounts[2] {/dev devtmpfs devtmpfs []} MUST be mounted in order after mounts[0] {/tmp tmpfs none []}, but not` may be better.

wking:
On Thu, Aug 31, 2017 at 06:50:20AM +0000, Ma Shimiao wrote:
> The changes are good, but I don't think something like `system mount
> 1` really make sense.  we don't need to care about what it really
> mounted number, we just need to care about if it is mounted in
> order.  And `system mount` is not easy understand.  I think
> `mounts[2] {/dev devtmpfs devtmpfs []} MUST be mounted in order
> after mounts[0] {/tmp tmpfs none []}, but not` may be better.

I like the system mount order information (or I wouldn't have put it
in ;), but I don't feel that strongly about it.  Removed in 60ab7ad →
bcb5379, so now the errors look like:

  * mounts[1] {/tmp tmpfs none []} does not exist
  * mounts[2] {/dev devtmpfs devtmpfs []} mounted before mounts[0] {/tmp tmpfs none []}

Mashimiao:
LGTM

liangchenye:
LGTM

PR #457, opened 2017-08-31
Makefile: add gofmt opition

q384566678:


Mashimiao:
we already have .gofmt

q384566678:
> we already have .gofmt

.gofmt is just testing code for compliance.
I added this feature to directly modify the code that does not conform to the specification.

q384566678:
Sorry. I made a mistake.

PR #458, opened 2017-08-31
rebase the specerror framework ; complete rfc code of config.md

liangchenye:

### DIFF
--- a/specerror/bundle.go
+++ b/specerror/bundle.go
@@ -1,0 +1,29 @@
+package specerror
+
+import (
+	"fmt"
+
+	rfc2119 "github.com/opencontainers/runtime-tools/error"
+)
+
+// define error codes
+const (
+	// ConfigInRootBundleDir represents "This REQUIRED file MUST reside in the root of the bundle directory"
+	ConfigInRootBundleDir = "This REQUIRED file MUST reside in the root of the bundle directory."
+	// ConfigConstName represents "This REQUIRED file MUST be named `config.json`."
+	ConfigConstName = "This REQUIRED file MUST be named `config.json`."
+	// ArtifactsInSingleDir represents "When supplied, while these artifacts MUST all be present in a single directory on the local filesystem, that directory itself is not part of the bundle."
+	ArtifactsInSingleDir = "When supplied, while these artifacts MUST all be present in a single directory on the local filesystem, that directory itself is not part of the bundle."
+)
+
+var (
+	containerFormatRef = func(version string) (reference string, err error) {
+		return fmt.Sprintf(referenceTemplate, version, "bundle.md#container-format"), nil
+	}
+)
+
+func init() {
+	register(ConfigInRootBundleDir, rfc2119.Must, containerFormatRef)
+	register(ConfigConstName, rfc2119.Must, containerFormatRef)
+	register(ArtifactsInSingleDir, rfc2119.Must, containerFormatRef)
+}
--- a/specerror/config-linux.go
+++ b/specerror/config-linux.go
@@ -1,0 +1,23 @@
+package specerror
+
+import (
+	"fmt"
+
+	rfc2119 "github.com/opencontainers/runtime-tools/error"
+)
+
+// define error codes
+const (
+	// DefaultFilesystems represents "The following filesystems SHOULD be made available in each container's filesystem:"
+	DefaultFilesystems = "The following filesystems SHOULD be made available in each container's filesystem:"
+)
+
+var (
+	defaultFilesystemsRef = func(version string) (reference string, err error) {
+		return fmt.Sprintf(referenceTemplate, version, "config-linux.md#default-filesystems"), nil
+	}
+)
+
+func init() {
+	register(DefaultFilesystems, rfc2119.Should, defaultFilesystemsRef)
+}
--- a/specerror/error.go
+++ b/specerror/error.go
@@ -13,46 +13,13 @@ const referenceTemplate = "https://github.com/opencontainers/runtime-spec/blob/v
 
 // Code represents the spec violation, enumerating both
 // configuration violations and runtime violations.
-type Code int
+type Code string
 
 const (
 	// NonError represents that an input is not an error
-	NonError Code = iota
+	NonError = "the input is not an error"
 	// NonRFCError represents that an error is not a rfc2119 error
-	NonRFCError
-
-	// ConfigFileExistence represents the error code of 'config.json' existence test
-	ConfigFileExistence
-	// ArtifactsInSingleDir represents the error code of artifacts place test
-	ArtifactsInSingleDir
-
-	// SpecVersion represents the error code of specfication version test
-	SpecVersion
-
-	// RootOnNonHyperV represents the error code of root setting test on non hyper-v containers
-	RootOnNonHyperV
-	// RootOnHyperV represents the error code of root setting test on hyper-v containers
-	RootOnHyperV
-	// PathFormatOnWindows represents the error code of the path format test on Window
-	PathFormatOnWindows
-	// PathName represents the error code of the path name test
-	PathName
-	// PathExistence represents the error code of the path existence test
-	PathExistence
-	// ReadonlyFilesystem represents the error code of readonly test
-	ReadonlyFilesystem
-	// ReadonlyOnWindows represents the error code of readonly setting test on Windows
-	ReadonlyOnWindows
-
-	// DefaultFilesystems represents the error code of default filesystems test
-	DefaultFilesystems
-
-	// CreateWithID represents the error code of 'create' lifecyle test with 'id' provided
-	CreateWithID
-	// CreateWithUniqueID represents the error code of 'create' lifecyle test with unique 'id' provided
-	CreateWithUniqueID
-	// CreateNewContainer represents the error code 'create' lifecyle test that creates new container
-	CreateNewContainer
+	NonRFCError = "the error is not a rfc2119 error"
 )
 
 type errorTemplate struct {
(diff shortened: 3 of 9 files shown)


liangchenye:
I'll rebase it after https://github.com/opencontainers/runtime-tools/pull/455.

error/runtime_spec.go
```diff
@@ -41,6 +41,73 @@ const (
...
+
+	// Platform-specific configuration
+	// WindowsPlatform represents the error code of windows config test on Windows platform
+	WindowsPlatform
+
+	// POSIX-platform Hooks
+	// PosixHookPathAbs represents the error code of absolute path test of posix hook path
+	PosixHookPathAbs
+	// PosixHookTimeout represents the error code of hook timeout test
+	PosixHookTimeout
```
wking: I'd recommend more specific names for future-proofing.  For example, `PosixHookTimeoutPositive` for “[If set, `timeout` MUST be greater than zero][1]”.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#platform-specific-configuration

error/runtime_spec.go
```diff
@@ -30,7 +30,7 @@ const (
 	RootOnNonHyperV
 	// RootOnHyperV represents the error code of root setting test on hyper-v containers
 	RootOnHyperV
-	// PathFormatOnWindows represents the error code of the path format test on Window
+	// PathFormatOnWindows represents the error code of the path format test on Windows
 	PathFormatOnWindows
 	// PathName represents the error code of the path name test
 	PathName
```
wking: This could be more specific, e.g. `PosixRootPathConvention` or `PosixRootPathRootfs`, because it's a POSIX-only SHOULD about `root.path` suggesting “[the conventional `rootfs`][1]”.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#root

error/runtime_spec.go
```diff
@@ -41,6 +41,73 @@ const (
 	// ReadonlyOnWindows represents the error code of readonly setting test on Windows
 	ReadonlyOnWindows
```
wking: This could be a more specific `WindowsRootReadonlyFalse`, and the earlier `ReadonlyFilesystem` could be `RootReadonlyImplemented` or some such, to distinguish from other `root.readonly` issues like using a non-boolean value.

error/runtime_spec.go
```diff
@@ -41,6 +41,73 @@ const (
...
 
+	// MountInOrder represents the error code of order test of mount list
+	MountInOrder
+	// MountDestinationAbs represents the error code of absolute path test of mount destination
+	MountDestinationAbs
+	// MountNotNestedOnWindows represents the error code of nested mount point on Windows
+	MountNotNestedOnWindows
```
wking: I recommend consistent platform namespacing.  You have an `OnWindows` suffix here, but have other entries with a platform prefix (e.g. `WindowsPlatform`, `LinuxOomScoreAdj`, `PosixHookPathAbs`).  I recommend a template like:

    {platform, if platform-specific}{JSON path in camelCase}{something about what's being tested}

which would make this one `WindowsMountsDestinationNotNested`.

error/runtime_spec.go
```diff
@@ -41,6 +41,73 @@ const (
...
+
+	// PosixProcess
+	// RlimitType represents the error code of rlimit type test
+	RlimitType
+	// RlimitSoftCur represents the error code of rlimit soft test
+	RlimitSoftCur
+	// RlimitHardMax represents the error code of rlimit hard test
+	RlimitHardMax
+	// RlimitDuplicated represents the error code of rlimit duplicate type test
+	RlimitDuplicated
```
wking: I don't see an entry in your rlimits codes for “[For each entry in `rlimits`, a `getrlimit(3)` on `type` MUST succeed][1]”.  That would be something like `PosixProcessRlimitsTypeGet`.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#posix-process

error/runtime_spec.go
```diff
@@ -41,6 +41,73 @@ const (
...
+	// PosixHookPathAbs represents the error code of absolute path test of posix hook path
+	PosixHookPathAbs
+	// PosixHookTimeout represents the error code of hook timeout test
+	PosixHookTimeout
+	// PosixHookCalledInOrder represents the error code of order test of called hooks
+	PosixHookCalledInOrder
+	// PosixHookStateOverStdin represents the error code of `state` information test if it was passed to hook process
+	PosixHookStateOverStdin
+	// PosixHookPrestartOrder represents the error code of order test of pre-start hook
+	PosixHookPrestartOrder
```
wking: Is this a subcase of `PosixHookCalledInOrder`?  Or is this intended for “[The pre-start hooks MUST be called after the start operation is called but before the user-specified program command is executed][1]”?  I don't see a need for the former, and think a better name for the latter would be `PosixHooksPrestartTiming` or something so we can distinguish between the order of one hook relative to another and the order of the hooks as a whole relative to the surrounding lifecycle.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#prestart

error/runtime_spec.go
```diff
@@ -41,6 +41,73 @@ const (
...
+	// PosixHookPoststartOrder represents the error code of order test of post-start hook
+	PosixHookPoststartOrder
+	// PosixHookPoststopOrder represents the error code of order test of post-stop hook
+	PosixHookPoststopOrder
+
+	// Annotations
+	// AnnotationKeyValue represents the error code of annotations key-value map test
+	AnnotationKeyValue
```
wking: There are [several RFC 2119 conditions for annotations][1], but you only have one code here.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#annotations

error/runtime_spec.go
```diff
@@ -41,6 +41,73 @@ const (
...
+
+	// Extensibility
+	// IgnoreUnknownProperties represents the error code of unknown properties test
+	IgnoreUnknownProperties
```
wking: There are [two MUSTs for this in the spec][1], but I've filed opencontainers/runtime-spec#916 to collapse them to one.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config.md#L457-L458

liangchenye:
@wking I'll use the inline language. It is always difficult for me to find suitable English word to describe a spec error.
Ideally, if we have more specific https://github.com/opencontainers/runtime-spec/issues/634, we can generate the spec error code from spec files automatically.

error/runtime_spec.go
liangchenye: I saw they were quoted, should we remove the this quote from the spec?
```
Keys MUST be strings.
Keys MUST NOT be an empty string.
Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myKey`.
Keys using the `org.opencontainers` namespace are reserved and MUST NOT be used by subsequent specifications.
Implementations that are reading/processing this configuration file MUST NOT generate an error if they encounter an unknown annotation key.

Values MUST be strings.
Values MAY be an empty string.
```

wking: > I saw they were quoted…

I see some backticks in the spec you reference, but no quotes.  And backticks are just markup, they don't affect the number of RFC 2119 conditions contained in the text.

wking:
On Fri, Sep 01, 2017 at 03:07:19AM +0000, 梁辰晔 (Liang Chenye) wrote:
> Ideally, if we have more specific
> https://github.com/opencontainers/runtime-spec/issues/634, we can
> generate the spec error code from spec files automatically.

Yeah, that would be good.  But I'd much rather have gotten them via
opencontainers/runtime-spec#628 and am not interested in maintaining
per-sentence link anchors by hand (the approach taken with
opencontainers/runtime-spec#634).  If someone else wants to step up
and add those to the spec, then great, but I'll be very surprised if
it happens in the next three months ;).

liangchenye:
I wrote a simple tool today to help make life easier:
https://github.com/liangchenye/markdown-to-json
So far, the output seems good.
https://github.com/liangchenye/markdown-to-json/blob/master/testdata/config.output
Just for fun!

liangchenye:
We have several `RootPath` so we need to change them by hand.
But we can map the error code to the string automatically, (`The value SHOULD be the conventional...`), so it would be easy to sync the spec errors with the runtime-spec.

```
   RootPath
	// RootPath represents "The value SHOULD be the conventional `rootfs`."
        RootPath
	// RootPath represents "A directory MUST exist at the path declared by the field."
```

liangchenye:
updated base on https://github.com/liangchenye/markdown-to-json/blob/master/testdata/config.output
The current naming rules are:
-  Full Property Name + Condition + Expected
   Take `RootPathOnPosixConvention` for example, 
          `RootPath` is the full property, 'OnPosix' is the condition, 'Convention' is the expected result.
- Make it shorter
   Proc -- Process for example.
- Polish the name   
    Now following most of @wking 's suggestion, like 'Timing'...

The `ref` and the `comment` are generated automatically by tool.

Mashimiao:
awesome!

liangchenye:
The https://github.com/liangchenye/markdown-to-json tool is updated, I try to make as useful as i18n tool (`make update-po`).

specerror/error.go
```diff
@@ -55,56 +122,140 @@ const (
...
-	ReadonlyOnWindows:   {Level: rfc2119.Must, Reference: rootRef},
+	RootOnWindowsRequired:      {Level: rfc2119.Required, Reference: rootRef},
+	RootOnHyperVNotSet:         {Level: rfc2119.Must, Reference: rootRef},
+	RootOnNonHyperVRequired:    {Level: rfc2119.Required, Reference: rootRef},
+	RootPathOnWindowsGUID:      {Level: rfc2119.Must, Reference: rootRef},
+	RootPathOnPosixConvention:  {Level: rfc2119.Should, Reference: rootRef},
+	RootPathExist:              {Level: rfc2119.Must, Reference: rootRef},
+	RootReadonlyImplement:      {Level: rfc2119.Must, Reference: rootRef},
+	RootReadonlyOnWindowsFalse: {Level: rfc2119.Must, Reference: rootRef},
+	// Mounts
```
wking: The constants now have namespaced names (e.g. `Mounts*`), so there's no need for comments like `// Mounts`, your earlier `// Root`, etc.  There may still be a use for the file-level comments (`// Bundle.md` and `// Config.md`) with the current names, but in order to avoid collisions it may be better to include that level in the constant names too (`ConfigFileExistence` → `BundleConfigFileExistence`, `RootOnWindowsRequired` → `ConfigRootOnWindowsRequired`, etc.).

wking:
Based on #465, this should go on the [v0.3.0 milestone][1].

Mashimiao:
yes, marked

liangchenye:
I got near 200 RFC errors and found it is difficult to maintain all of them in one file.
So I'll add different go files to match 'config.md', 'config-linux.md' and the etc...

liangchenye:
- rewrite the spec error framework
  the error.go just provide the basic template and common function.
  provide a 'register' function so each .go (generate from .md) could register
  its RFC codes.
- complete the config.go
   this is originally get from markdown-to-json tool
- add small part rfc codes of bundle.md/runtime.md

PTAL @wking @Mashimiao @q384566678

specerror/error.go
```diff
@@ -79,52 +46,14 @@ type LevelErrors struct {
...
-		return fmt.Sprintf(referenceTemplate, version, "config.md#root"), nil
-	}
-	defaultFSRef = func(version string) (reference string, err error) {
-		return fmt.Sprintf(referenceTemplate, version, "config-linux.md#default-filesystems"), nil
-	}
-	runtimeCreateRef = func(version string) (reference string, err error) {
-		return fmt.Sprintf(referenceTemplate, version, "runtime.md#create"), nil
+var ociErrors = map[Code]errorTemplate{}
+
+func registOCIError(code Code, level rfc2119.Level, ref func(versiong string) (string, error)) {
```
wking: This package is just about runtime-spec requirement, so this can be `register`.  No need to have `OCI` or `error` in the name, because they're covered by the package path.

specerror/error.go
```diff
@@ -79,52 +46,14 @@ type LevelErrors struct {
...
+	if _, ok := ociErrors[code]; ok {
+		panic(fmt.Sprintf("should not regist a same code twice: %s", code))
```
wking: Maybe: `regist a` -> `register the`.

specerror/bundle.go
```diff
@@ -0,0 +1,31 @@
+// Package specerror implements runtime-spec-specific tooling for
+// tracking RFC 2119 violations.
```
wking: To stay DRY, I think you only want this comment in `error.go`.

specerror/config-linux.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"fmt"
+
+	rfc2119 "github.com/opencontainers/runtime-tools/error"
+)
+
+// define error codes
+const (
+	// DefaultFilesystems represents "The following filesystems SHOULD be made available in each container's filesystem:"
+	DefaultFilesystems = "The following filesystems SHOULD be made available in each container's filesystem:"
```
wking: Is this really our only `config-linux.md` requirement?  Either way, I'd rather get this PR landed, because the change feels to big to wait until it's complete and perfect.

liangchenye: no，these items were already recorded in runtime tools implementation. I add them first. The whole rfc errors will come later.

specerror/error.go
liangchenye: I‘ll change it tomorrow morning.

liangchenye:
updated, rename registOCIError to `register`.

q384566678:
LGTM

Mashimiao:
LGTM

PR #459, opened 2017-09-01
validate: Increase the valid value of devices.type

q384566678:
According to the [spec.](https://github.com/opencontainers/runtime-spec/blame/master/config-linux.md#L225)

### DIFF
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -705,7 +705,7 @@ func (v *Validator) CheckLinuxResources() (errs error) {
 	}
 	for index := 0; index < len(r.Devices); index++ {
 		switch r.Devices[index].Type {
-		case "a", "b", "c":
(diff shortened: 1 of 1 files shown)


Mashimiao:
type validation is just value check, I prefer to put it into schema validation. We will remove it after #197  landed.

Mashimiao:
And `""` is not a valid value for devices.type. From the spec, unset value means it's nil not `""`

wking:
We can drop the whole `r.Devices[index].Type` check if/when opencontainers/runtime-spec#690 lands,after which 1.0.x JSON Schema checks will handle that for us.

Mashimiao:
As #197  landed, let's close this PR. We will remove all duplicate validation items with schema in the future.

PR #460, opened 2017-09-01
validate: modify the condition of the deviceValid

q384566678:
Accord to the [mknod(1)](http://man7.org/linux/man-pages/man1/mknod.1.html). We can get the following information:

> Both MAJOR and MINOR must be specified when TYPE is b, c, or u, and they must be omitted when TYPE is p.


Mashimiao:
LGTM

q384566678:
@hqhq @liangchenye PTAL

liangchenye:
LGTM

PR #462, opened 2017-09-05
cmd: modify the err judgment

q384566678:
Modify the following error situation:

```
✗ oci-runtime-tool validate 
0 errors occurred:


```


Mashimiao:
This is really not a good fix...., you'd better fix take advantage of multierror to fix this in CheckAll()

q384566678:
updated, PTAL.

Mashimiao:
LGTM

mrunalp:
LGTM

PR #469, opened 2017-09-09
cap: drop existent item before validating

Mashimiao:
fix pointed out issue at [here](https://github.com/opencontainers/runtime-tools/pull/466#issuecomment-327980693)

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -1050,9 +1050,6 @@ func (g *Generator) AddProcessCapability(c string) error {
 // DropProcessCapability drops a process capability from g.spec.Process.Capabilities.
 func (g *Generator) DropProcessCapability(c string) error {
 	cp := strings.ToUpper(c)
-	if err := validate.CapValid(cp, false); err != nil {
-		return err
-	}
 
 	g.initSpecProcessCapabilities()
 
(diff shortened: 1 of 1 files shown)


runcom:
LGTM

q384566678:
LGTM

mrunalp:
LGTM

generate/generate.go
```diff
@@ -1050,9 +1050,6 @@ func (g *Generator) AddProcessCapability(c string) error {
 // DropProcessCapability drops a process capability from g.spec.Process.Capabilities.
 func (g *Generator) DropProcessCapability(c string) error {
 	cp := strings.ToUpper(c)
-	if err := validate.CapValid(cp, false); err != nil {
-		return err
-	}
 
 	g.initSpecProcessCapabilities()
```
wking: Unrelated in this PR, and I don't care enough about generation to work up a fix myself, but if anyone else is interested, I think calling `initSpecProcessCapabilities` in `DropProcessCapability` is strange.  It would be more intuitive to me if the function just returned `nil` without changing the config if any of the earlier pointers were `nil`.  As it stands:

```
$ ./oci-runtime-tool generate --template <(echo '{}') --process-cap-drop CAP_SYS_ADMIN
{
        "ociVersion": "1.0.0",
        "process": {
                "user": {
                        "uid": 0,
                        "gid": 0
                },
                "args": null,
                "cwd": "/",
                "capabilities": {}
        },
        "root": {
                "path": "rootfs"
        }
}
```

which is creating a useless empty object for `process.capabilities`.

PR #472, opened 2017-09-14
validate: add root.path validation when platform is windows

q384566678:
According to the [spec.](https://github.com/opencontainers/runtime-spec/blame/master/config.md#L37)

>On Windows, path MUST be a volume GUID path.

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -1089,11 +1089,7 @@ func (g *Generator) DropProcessCapability(c string) error {
 		}
 	}
 
-	if err := validate.CapValid(cp, false); err != nil {
-		return err
-	}
-
-	return nil
+	return validate.CapValid(cp, false)
 }
 
 func mapStrToNamespace(ns string, path string) (rspec.LinuxNamespace, error) {
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -178,6 +178,23 @@ func (v *Validator) CheckRoot() (errs error) {
 		return
 	}
 
+	if v.platform == "windows" {
+		matched, err := regexp.MatchString(`\\\\[?]\\Volume[{][a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}[}]\\`, v.spec.Root.Path)
+		if err != nil {
+			errs = multierror.Append(errs, err)
+		} else if !matched {
+			errs = multierror.Append(errs,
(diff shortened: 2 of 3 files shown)


validate/validate.go
```diff
@@ -219,6 +219,11 @@ func (v *Validator) CheckRoot() (errs error) {
 	}
 
 	if v.platform == "windows" {
+		if !strings.HasPrefix(v.spec.Root.Path, `\\?\`) {
```
wking: We can be stricter than this.  The spec links [here][1], whivh would be a regex like `[\][\][?][\]Volume[
{][a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}[}][\]?"`.  The Windows docs say "GUID", I think they mean "[UUID][2]", based on their `\\?\Volume{26a21bda-a627-11d7-9931-806e6f6e6963}\` example.

Also, unit tests :).

[1]: https://msdn.microsoft.com/en-us/library/windows/desktop/aa365248(v=vs.85).aspx
[2]: https://tools.ietf.org/html/rfc4122#section-3

q384566678: updated, PTAL.

validate/validate_test.go
```diff
@@ -114,6 +114,7 @@ func TestCheckRoot(t *testing.T) {
 	}{
 		{rspec.Spec{Windows: &rspec.Windows{HyperV: &rspec.WindowsHyperV{}}, Root: &rspec.Root{}}, "windows", specerror.RootOnHyperV},
 		{rspec.Spec{Windows: &rspec.Windows{HyperV: &rspec.WindowsHyperV{}}, Root: nil}, "windows", specerror.NonError},
+		{rspec.Spec{Windows: &rspec.Windows{}, Root: &rspec.Root{Path: filepath.Join(tmpBundle, "rootfs")}}, "windows", specerror.PathFormatOnWindows},
```
wking: Can we also check that [this][1] passes validation?  The regex is large, and a positive match would help convince me that it wasn't broken.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#example-windows

q384566678: updated.

generate/generate.go
```diff
@@ -1089,11 +1089,7 @@ func (g *Generator) DropProcessCapability(c string) error {
 		}
 	}
 
-	if err := validate.CapValid(cp, false); err != nil {
-		return err
-	}
-
-	return nil
+	return validate.CapValid(cp, false)
```
wking: I like this change, but it seems orthogonal.  Can 1a9532e get its own PR?

Mashimiao:
LGTM

q384566678:
ping @mrunalp @liangchenye

liangchenye:
LGTM

PR #474, opened 2017-09-14
validate/validate_test: Handle JSON Schema test not raising an error

wking:
Without this, tests that expect an error tried to cast the `nil` to a multierror (which failed) and then tried to format a failure message with `errs.Error()` (which panicked).


Mashimiao:
LGTM

q384566678:
LGTM

PR #475, opened 2017-09-14
validate/validate_test: Better error messages for unexpected JSON Schema errors

wking:
Instead of raising messages like:

```
  Error:  Not equal:
          expected: <nil>(<nil>)
          actual:
            *multierror.Error(*multierror.Error{Errors:[]error{(*errors.errorString)(0xc42037adc0)},
            ErrorFormat:(multierror.ErrorFormatFunc)(nil)})
```

raise messages like:

```
  * linux.rootfsPropagation: linux.rootfsPropagation must be one of
      the following: "private", "shared", "slave", "unbindable"
```

### DIFF
--- a/validate/validate_test.go
+++ b/validate/validate_test.go
@@ -72,8 +72,10 @@ func TestJSONSchema(t *testing.T) {
 			v := &Validator{spec: tt.config}
 			errs := v.CheckJSONSchema()
 			if tt.error == "" {
-				assert.Equal(t, nil, errs)
-				return
+				if errs == nil {
+					return
+				}
+				t.Fatalf("expected no error, but got: %s", errs.Error())
 			}
 			merr, ok := errs.(*multierror.Error)
 			if !ok {


validate/validate_test.go
```diff
@@ -72,8 +72,10 @@ func TestJSONSchema(t *testing.T) {
 			v := &Validator{spec: tt.config}
 			errs := v.CheckJSONSchema()
 			if tt.error == "" {
-				assert.Equal(t, nil, errs)
-				return
+				if errs == nil {
+					return
+				}
+				t.Fatalf("expected no error, but got: %s", errs.Error())
```
Mashimiao: I don't understand the code change. I can't see any relations between the change with your comment.

wking: > I don't understand the code change.

If `tt.error` is an empty string, and `CheckJSONSchema` returns `nil`, then great, the test passes.  This worked fine before, and still works fine after this commit.

If `CheckJSONSchema` raises an error, the old `assert.Equal` was rendering ghe unhelpful `actual: *multierror...`.  With this commit, we get the useful `* `linux.rootfsPropagation...` (or whatever the returned error happens to be.

> I can't see any relations between the change with your comment.

The examples in the commit message / PR comment are from working up #476.  There's not a great example to use for test-failure messages, because all of the tests in master pass ;).

Mashimiao:
LGTM

q384566678:
LGTM

PR #473, opened 2017-09-14
validate: Delete the extra validation

q384566678:
Since #197 merged, it has verified.

Signed-off-by: zhouhao <zhouhao@cn.fujitsu.com>

PR #476, opened 2017-09-14
validate/validate_test: Add linux.rootfsPropagation checks

wking:
To demonstrate that this is covered by the 1.0.0 JSON Schema checks, and show that the omitempty protects us from errors for empty-string values despite there being no empty-string entry in [the JSON Schema][1].

These tests are in support for #473.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/schema/defs-linux.json#L4-L11

### DIFF
--- a/validate/validate_test.go
+++ b/validate/validate_test.go
@@ -63,6 +63,40 @@ func TestJSONSchema(t *testing.T) {
 		},
 		{
 			config: &rspec.Spec{
+				Version: "1.0.0",
+				Linux:   &rspec.Linux{},
+			},
+			error: "",
+		},
+		{
+			config: &rspec.Spec{
+				Version: "1.0.0",
+				Linux: &rspec.Linux{
+					RootfsPropagation: "",
+				},
+			},
+			error: "",
+		},
+		{
+			config: &rspec.Spec{
+				Version: "1.0.0",
+				Linux: &rspec.Linux{
+					RootfsPropagation: "shared",
+				},
+			},
+			error: "",
+		},
+		{
+			config: &rspec.Spec{
+				Version: "1.0.0",
+				Linux: &rspec.Linux{
+					RootfsPropagation: "rshared",
+				},
+			},
+			error: "linux.rootfsPropagation: linux.rootfsPropagation must be one of the following: \"private\", \"shared\", \"slave\", \"unbindable\"",
+		},
+		{
+			config: &rspec.Spec{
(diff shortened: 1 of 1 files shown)


validate/validate_test.go
```diff
@@ -63,6 +63,40 @@ func TestJSONSchema(t *testing.T) {
...
+				Version: "1.0.0",
+				Linux:   &rspec.Linux{},
+			},
+			error: "",
+		},
+		{
+			config: &rspec.Spec{
+				Version: "1.0.0",
+				Linux: &rspec.Linux{
+					RootfsPropagation: "",
```
Mashimiao: "" is not valid value for RootfsPropagation why error is ""?

wking: > "" is not valid value for RootfsPropagation why error is ""?

I exoect because it's `omitempty`, although I haven't tracked that down to specific gojsonschema code.

Mashimiao: I think omitempty  means it can be unset but not ""

wking: > I think omitempty means it can be unset but not ""

For a non-pointer `omitempty` like [`RootfsPropagation`][1], an empty string is as close as you can get to unset.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/specs-go/config.go#L154

q384566678: I think the `RootfsPropagation: "" `test should be deleted.

wking: > I think the `RootfsPropagation: ""` test should be deleted.

Why?  With the current non-pointer property, that's how you ex[licitly unset the property.  I think we need to be sure clearing `RootfsPropagation` is legal, because as this thread demonstrates, it's not immediately obvious.

q384566678: If it is not set, its value is also `""`, which is repeated with the previous test (without setting the value), and I think the test of this value is meaningless.

wking: > If it is not set, its value is also "", which is repeated with the previous test (without setting the value)...

Yeah, but it doesn't hurt to double check, and the explicit-empty-string case surprised @Mashimiao [at first][1].  If test rephrasings help expose surprises, O=I think we want them.

[1]: https://github.com/opencontainers/runtime-tools/pull/476#discussion_r139081493

Mashimiao:
LGTM

liangchenye:
LGTM

PR #478, opened 2017-09-20
rootfs.tar.gz: Bump to BusyBox 1.25.1

wking:
Update from the old 1.24.1 with a new extract from `stage3-amd64-20170907.tar.bz2`.

Generated with:

```
$ cd contrib/rootfs-builder
$ touch get-stage3.sh
$ make rootfs.tar.gz
$ cp -f rootfs.tar.gz ../..
```

This change adds the following commands to the rootfs:

* `blkdiscard`
* `nsenter`
* `ubirename`
* `unshare`

of which `nsenter` and `unshare` are interesting for runtime-tools (although having them *inside* the container probably doesn't matter much).

This bump is mostly because an update every two years is a good idea ;).


wking:
Travis failure is addressed by #472's 1a9532e.

liangchenye:
LGTM
(#472 is already merged, don't why the CI is not rebuilt.)

Mashimiao:
LGTM

PR #479, opened 2017-09-21
rootfs-*.tar.gz: Add per-arch tarballs (currently for amd64 and 386)

wking:
This PR also includes a number of other rootfs-builder improvements; details in the commit messages.

We can build tarballs for other platforms if we get access to a box that can run the target `busybox` binary, because part of the extraction process is asking the extracted `busybox` what commands it supports.

Builds on #478; review that first.


wking:
Rebased onto master with cec5f49 → afcac88 now that #472 and #478 have landed.

Mashimiao:
LGTM

mrunalp:
LGTM

PR #461, opened 2017-09-04
validate: add the validation of rlimit.type when platform is solaris

q384566678:
This is specific [spec](https://github.com/opencontainers/runtime-spec/blame/master/config.md#L167-L173).
Should increase the validation of `rlimit.type` on `solaris` system.

Signed-off-by: zhouhao <zhouhao@cn.fujitsu.com>

PR #480, opened 2017-09-21
validate/validate: Linux rlimits extend the POSIX rlimits

wking:
So don't duplicate those values.  Also drop `RLIMIT_LOCKS`, which [the man page documents as an obsolete-by-Linux-2.6 value][1].  A quick way to get a set of strings from the man page is:

    $ curl -s 'https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/plain/man2/getrlimit.2?h=man-pages-4.13' |
    >   sed -n 's/^\.BR\? \(RLIMIT_[^ ]*\)/\1/p' | sort | uniq

Also rename the old `default*Rlimits`, because these are valid-value arrays, not defaults.

Following up on the just-merged #461.

[1]: https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/man2/getrlimit.2?h=man-pages-4.13#n209

### DIFF
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -30,7 +30,8 @@ import (
 const specConfig = "config.json"
 
 var (
-	defaultRlimits = []string{
+	// http://pubs.opengroup.org/onlinepubs/9699919799/functions/getrlimit.html
+	posixRlimits = []string{
 		"RLIMIT_AS",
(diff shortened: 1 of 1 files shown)


Mashimiao:
LGTM

q384566678:
LGTM

PR #481, opened 2017-09-21
validate: Soften unrecognized rlimit types to SHOULD violations

wking:
This PR builds on #480; review that first.

The spec isn't particuarly clear on this, [saying][1]:

> * Linux: valid values are defined in the `getrlimit(2)` man page, such as `RLIMIT_MSGQUEUE`.
> * Solaris: valid values are defined in the `getrlimit(3)` man page, such as `RLIMIT_CORE`.

[and][2]:

> For each entry in `rlimits`, a `getrlimit(3)` on `type` MUST succeed.

It doesn't say:

> Linux: The value MUST be listed in the `getrlimit(2)` man page…

and it [doesn't require the runtime to support the values listed in the man page][4] (opencontainers/runtime-spec#813).  So there are three sets:

* Values listed in the man page
* Values supported by the host kernel
* Values supported by the runtime

And as the spec stands, these sets are only weakly coupled, and any of them could be a sub- or superset of any other.  In practice, I expect the sets to all coincide, with the kernel occasionally adding or removing values, and the man page and runtimes trailing along behind.

To address that, this commit weakens the previous hard error to a SHOULD-level error.  The `PosixProcRlimitsValueError` constant is new to this commit, because the spec contains neither a MUST nor a SHOULD for this condition, although I expect a SHOULD-level suggestion was implied by [this][1].  The `posixProcRef` constant is cherry-picked from #458.

Also make `CheckRlimits` a no-op on Windows, because [the spec does not define `process.rlimits` for that OS][5].

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config.md#L168-L169
[2]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config.md#L168-L169
[4]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config.md#L463
[5]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#posix-process

### DIFF
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -424,6 +424,10 @@ func (v *Validator) CheckCapabilities() (errs error) {
 
 // CheckRlimits checks v.spec.Process.Rlimits
 func (v *Validator) CheckRlimits() (errs error) {
+	if v.platform == "windows" {
+		return
+	}
+
 	process := v.spec.Process
 	for index, rlimit := range process.Rlimits {
 		for i := index + 1; i < len(process.Rlimits); i++ {
@@ -870,14 +874,14 @@ func (v *Validator) rlimitValid(rlimit rspec.POSIXRlimit) (errs error) {
 				return
 			}
 		}
-		errs = multierror.Append(errs, fmt.Errorf("rlimit type %q is invalid", rlimit.Type))
+		errs = multierror.Append(errs, specerror.NewError(specerror.PosixProcRlimitsTypeValueError, fmt.Errorf("rlimit type %q may not be valid", rlimit.Type), v.spec.Version))
 	} else if v.platform == "solaris" {
 		for _, val := range posixRlimits {
 			if val == rlimit.Type {
 				return
 			}
 		}
-		errs = multierror.Append(errs, fmt.Errorf("rlimit type %q is invalid", rlimit.Type))
+		errs = multierror.Append(errs, specerror.NewError(specerror.PosixProcRlimitsTypeValueError, fmt.Errorf("rlimit type %q may not be valid", rlimit.Type), v.spec.Version))
 	} else {
 		logrus.Warnf("process.rlimits validation not yet implemented for platform %q", v.platform)
 	}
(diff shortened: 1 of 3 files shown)


Mashimiao:
duplicate commit 894cae7b69398d2cc10d63df306b03c775575418 with #480 ?

wking:
> duplicate commit 894cae7 with #480 ?

Yup, which is why I opened with:

> This PR builds on #480; review that first.

If/when #480 settles, I'll rebse this one.

wking:
On Fri, Sep 22, 2017 at 08:19:28AM +0000, W. Trevor King wrote:
> If/when #480 settles, I'll rebase this one.

Done with 0696cb8 → d395004.

validate/validate.go
```diff
@@ -423,6 +423,10 @@ func (v *Validator) CheckCapabilities() (errs error) {
 
 // CheckRlimits checks v.spec.Process.Rlimits
 func (v *Validator) CheckRlimits() (errs error) {
+	if v.platform == "windows" {
```
q384566678: Can you add a warning output on this side? such as:
>logrus.Warnf("process.rlimits validation not yet implemented for platform %q", v.platform)

Also, does this have some overlap with the validation of the platform in `rlimitValid`?

wking: > Can you add a warning output on this side?

I don't think we want that.  “Not yet implemented” warnings are for cases where runtime-tools has not yet caught up with the spec's requirements.  But [Windows does not, and likely never will support `rlimits`][1] and [the spec does not define `rlimits` for Windows][2].

> Also, does this have some overlap with the validation of the platform in `rlimitValid`?

I'm touching `rlimitValid` in this PR.  Do you think we need `rlimitValid` changes beyond those?

[1]: https://github.com/opencontainers/runtime-spec/pull/835#issuecomment-303455386
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#posix-process

liangchenye:
rebase required @wking , in the current specerror/config.go, the releated code is PosixProcRlimitsTypeGeneError.

wking:
On Mon, Oct 23, 2017 at 07:55:53AM +0000, 梁辰晔 (Liang Chenye) wrote:
> … in the current specerror/config.go, the releated code is
> PosixProcRlimitsTypeGeneError.

Rebased with d395004 → 7553161.  I kept a specific-to-this-situation
error (now PosixProcRlimitsTypeValueError) because this check is based
on the man-page reference.  The PosixProcRlimitsTypeGeneError, on the
other hand, is [1]:

* a MUST, and
* a runtime requirement, not a config requirement.

[1]: https://github.com/opencontainers/runtime-tools/blob/6554addc535ff61f02ce943e1a048148944be497/specerror/config.go#L45-L46

liangchenye:
LGTM

liangchenye:
and add an track https://github.com/liangchenye/markdown-to-json/issues/1

Mashimiao:
LGTM

PR #484, opened 2017-09-25
fix rootfs-propagation

q384566678:

### DIFF
--- a/completions/bash/oci-runtime-tool
+++ b/completions/bash/oci-runtime-tool
@@ -130,11 +130,9 @@ __oci-runtime-tool_complete_compliance_level() {
 __oci-runtime-tool_complete_propagations() {
 	COMPREPLY=( $( compgen -W "
 		private
-		rprivate
-		rshared
-		rslave
 		shared
 		slave
+		unbindable
 	" -- "$cur" ) )
 }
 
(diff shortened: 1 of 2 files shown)


completions/bash/oci-runtime-tool
```diff
@@ -130,11 +130,9 @@ __oci-runtime-tool_complete_compliance_level() {
 __oci-runtime-tool_complete_propagations() {
 	COMPREPLY=( $( compgen -W "
 		private
-		rprivate
-		rshared
-		rslave
```
wking: This catches up with the 1.0 spec, but we may want to see how opencontainers/runtime-spec#923 plays out before removing them.  Or we can remove them now and restore them later if the the `r*` forms get added to the spec.

q384566678: I think we can remove them now, and make changes when spec changes.

Mashimiao:
LGTM

q384566678:
I have modified a similar situation in `generate.go`. PTAL

q384566678:
ping @Mashimiao @liangchenye @hqhq @mrunalp

Mashimiao:
LGTM

q384566678:
@liangchenye @hqhq PTAL

completions/bash/oci-runtime-tool
liangchenye: +1, we can remove them now.

liangchenye:
LGTM

PR #486, opened 2017-09-27
validate: Code optimization

q384566678:
Make the error message more specific.

### DIFF
--- a/specerror/config-linux.go
+++ b/specerror/config-linux.go
@@ -10,6 +10,9 @@ import (
 const (
 	// DefaultFilesystems represents "The following filesystems SHOULD be made available in each container's filesystem:"
 	DefaultFilesystems Code = 0xc001 + iota
+
+	// NSTypeValueError represents "The following namespace types are supported:"
+	NSTypeValueError
 	// NSPathAbs represents "This value MUST be an absolute path in the runtime mount namespace."
 	NSPathAbs
 	// NSProcInPath represents "The runtime MUST place the container process in the namespace associated with that `path`."
@@ -105,6 +108,7 @@ var (
 
 func init() {
 	register(DefaultFilesystems, rfc2119.Should, defaultFilesystemsRef)
+	register(NSTypeValueError, rfc2119.Should, namespacesRef)
 	register(NSPathAbs, rfc2119.Must, namespacesRef)
 	register(NSProcInPath, rfc2119.Must, namespacesRef)
 	register(NSPathMatchTypeError, rfc2119.Must, namespacesRef)
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -597,8 +597,8 @@ func (v *Validator) CheckLinux() (errs error) {
 
 	for index := 0; index < len(v.spec.Linux.Namespaces); index++ {
 		ns := v.spec.Linux.Namespaces[index]
-		if !v.namespaceValid(ns) {
-			errs = multierror.Append(errs, fmt.Errorf("namespace %v is invalid", ns))
+		if err := v.namespaceValid(ns); err != nil {
+			errs = multierror.Append(errs, err)
 		}
 
 		tmpItem := nsTypeList[ns.Type]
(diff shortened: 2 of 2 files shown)


validate/validate.go
```diff
@@ -903,14 +903,14 @@ func namespaceValid(ns rspec.LinuxNamespace) bool {
 	case rspec.UserNamespace:
 	case rspec.CgroupNamespace:
 	default:
-		return false
+		return fmt.Errorf("invalid namespace type %s", ns.Type)
 	}
 
 	if ns.Path != "" && !filepath.IsAbs(ns.Path) {
-		return false
+		return fmt.Errorf("%v is not an absolute path", ns.Path)
```
Mashimiao: maybe fmt.Errorf(path %v of namespace %v is not absolute, ns.Type, ns.Path) better?

q384566678: Fine.

validate/validate.go
```diff
@@ -903,14 +903,14 @@ func namespaceValid(ns rspec.LinuxNamespace) bool {
 	case rspec.UserNamespace:
 	case rspec.CgroupNamespace:
 	default:
-		return false
+		return fmt.Errorf("invalid namespace type %s", ns.Type)
```
wking: The spec actually has no RFC2119 wording for this check, saying only “[The following namespace types are supported][1]”.  It doesn't say that runtimes MUST support them, or that they MUST NOT support additional values, or even that configurations SHOULD use values from that list.  I'd guess the spec-maintainer intention was closer to:

> Configurations SHOULD use, and runtimes SHOULD support, the following values:

Compliance testing would be easier with “runtimes MUST support”, but see opencontainers/runtime-spec#813.

Anyhow, from the config-validation side, I'd be happier seeing this supported by an effective-SHOULD, similar to what I've proposed in #481.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config-linux.md#L29-L36

wking:
This PR belongs in [v0.3.0][1].  It's not using `specerror.NewError` yet, but returning an `error` from `namespaceValid` is necessary groundwork for future `NewError` use.

q384566678:
updated, @Mashimiao @liangchenye @wking PTAL

specerror/config-linux.go
```diff
@@ -10,14 +10,21 @@ import (
...
 		return fmt.Sprintf(referenceTemplate, version, "config-linux.md#default-filesystems"), nil
 	}
+	namespacesRef = func(version string) (reference string, err error) {
+		return fmt.Sprintf(referenceTemplate, version, "config-linux.md#namespaces"), nil
+	}
 )
 
 func init() {
 	register(DefaultFilesystems, rfc2119.Should, defaultFilesystemsRef)
+	register(NamespacesType, rfc2119.Should, namespacesRef)
```
liangchenye: The spec says: "The following namespace types are supported", We don't have key words like 'SHOULD'. Maybe we can add a PR to spec project?

specerror/config-linux.go
```diff
@@ -10,14 +10,21 @@ import (
 const (
 	// DefaultFilesystems represents "The following filesystems SHOULD be made available in each container's filesystem:"
 	DefaultFilesystems = "The following filesystems SHOULD be made available in each container's filesystem:"
+	NamespacesType     = "The following namespace types are supported:"
+	NamespacesPath     = "This value MUST be an absolute path."
```
liangchenye: I prefer to extend it to "This value MUST be an absolute path in the runtime mount namespace. "
There are several 'an absolute path' RFCs,  https://github.com/opencontainers/runtime-tools/blob/master/specerror/config.go#L32,  https://github.com/opencontainers/runtime-tools/blob/master/specerror/config.go#L42.
It will be easier to differentiate them.

liangchenye: rename to NamespacePathAbs?

specerror/config-linux.go
q384566678: >Maybe we can add a PR to spec project?

Approve.

q384566678: I open https://github.com/opencontainers/runtime-spec/pull/931.

liangchenye:
rebase required

q384566678:
rebased.

q384566678:
Closed because the code has been deleted.

PR #488, opened 2017-09-28
validate: fix cap validation

q384566678:

### DIFF
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -410,7 +410,7 @@ func (v *Validator) CheckCapabilities() (errs error) {
 			if effective && !permitted {
 				errs = multierror.Append(errs, fmt.Errorf("effective capability %q is not allowed, as it's not permitted", capability))
 			}
-			if ambient && !(effective && inheritable) {
+			if ambient && !(permitted && inheritable) {
 				errs = multierror.Append(errs, fmt.Errorf("ambient capability %q is not allowed, as it's not permitted and inheribate", capability))
 			}
 		}


Mashimiao:
This is not a correct fix. The manpage says ` The ambient capability set obeys the invariant that  no  capability can ever be ambient if it is not both permitted and inheritable. So just keep `&&`
`

q384566678:
My understanding is that both the `permitted` and the `inheritable` are valid for `flase`, but in the current case, when one of them is `false`, it is executed, which is not in accordance with the rules.
Also we need to change `effective` to `permitted`.

wking:
On Thu, Sep 28, 2017 at 08:24:38AM +0000, Zhou Hao wrote:
> Also we need to change `effective` to `permitted`.

I agree with this, but as @Mashimiao points out [1], we want to keep
&&.  Demonstrating in Python, but the boolean logic is true regardless
of language:

  for permitted in (True, False):
      for inheritable in (True, False):
          ambient_legal = (permitted and inheritable)
          print('permitted: {}, inheritable: {}, ambient legal: {}'.format(permitted, inheritable, ambient_legal))
  permitted: True, inheritable: True, ambient legal: True
  permitted: True, inheritable: False, ambient legal: False
  permitted: False, inheritable: True, ambient legal: False
  permitted: False, inheritable: False, ambient legal: False

However, the code you have in 1333713 produces:

  for permitted in (True, False):
      for inheritable in (True, False):
          ambient_legal = (permitted or inheritable)
          print('permitted: {}, inheritable: {}, ambient legal: {}'.format(permitted, inheritable, ambient_legal))
  # permitted: True, inheritable: True, ambient legal: True
  # permitted: True, inheritable: False, ambient legal: True
  # permitted: False, inheritable: True, ambient legal: True
  # permitted: False, inheritable: False, ambient legal: False

If the currrent !(A && B) is confusing, you can unpack with DeMorgan's
law [2]:

  ambient && (!permitted || !inheritable)

And a great way to make sure we're doing this right is to have unit
tests excercising CheckCapabilities in the different cases :).

[1]: https://github.com/opencontainers/runtime-tools/pull/488#issuecomment-332757122
[2]: https://en.wikipedia.org/wiki/De_Morgan's_laws

wking:
This PR should be part of [v0.3.0][1].  As a config-validation bugfix, it really belonged in [v0.2.0][2], but we've cut v0.2.0, and v0.3.0 is the next-most-urgent milestone ;).

[2]: https://github.com/opencontainers/runtime-tools/milestone/4

wking:
> permitted: True, inheritable: False, ambient legal: True

This is not [both permitted and inheritable][2].

[2]: https://github.com/opencontainers/runtime-tools/pull/488#issuecomment-332757122

q384566678:
Sorry, I misunderstood.

Mashimiao:
LGTM

q384566678:
ping @liangchenye @hqhq @mrunalp

q384566678:
@liangchenye @hqhq PTAL

liangchenye:
LGTM

PR #490, opened 2017-09-30
fix compile issue

liangchenye:
Signed-off-by: Liang Chenye <liangchenye@huawei.com>

PR #491, opened 2017-09-30
validate: allow unset "type" fields in resource devices whitelist

cyphar:
According to the spec, an unset .Type field in the
.Linux.Resources.Devices list is permitted (and it maps to "all").
This was broken recently, making our own default profile (as well as
umoci's and runc's) not pass validation anymore.

### DIFF
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -749,7 +749,7 @@ func (v *Validator) CheckLinuxResources() (errs error) {
 	}
 	for index := 0; index < len(r.Devices); index++ {
 		switch r.Devices[index].Type {
-		case "a", "b", "c":
+		case "a", "b", "c", "":
 		default:
 			errs = multierror.Append(errs, fmt.Errorf("type of devices %s is invalid", r.Devices[index].Type))
 		}
--- a/validation/generate_test.go
+++ b/validation/generate_test.go
@@ -1,0 +1,56 @@
+package validation
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"runtime"
+	"testing"
+
+	rfc2119 "github.com/opencontainers/runtime-tools/error"
+	"github.com/opencontainers/runtime-tools/generate"
+	"github.com/opencontainers/runtime-tools/specerror"
+	"github.com/opencontainers/runtime-tools/validate"
+)
+
+// Smoke test to ensure that _at the very least_ our default configuration
+// passes the validation tests. If this test fails, something is _very_ wrong
+// and needs to be fixed immediately (as it will break downstreams that depend
+// on us for a "sane default" and do compliance testing -- such as umoci).
+func TestGenerateValid(t *testing.T) {
+	bundle, err := ioutil.TempDir("", "TestGenerateValid_bundle")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(bundle)
+
+	// Create our toy bundle.
+	rootfsPath := filepath.Join(bundle, "rootfs")
+	if err := os.Mkdir(rootfsPath, 0755); err != nil {
+		t.Fatal(err)
+	}
+	configPath := filepath.Join(bundle, "config.json")
+	g := generate.New()
+	if err := (&g).SaveToFile(configPath, generate.ExportOptions{Seccomp: false}); err != nil {
+		t.Fatal(err)
+	}
+
+	// Validate the bundle.
+	v, err := validate.NewValidatorFromPath(bundle, true, runtime.GOOS)
+	if err != nil {
+		t.Errorf("unexpected NewValidatorFromPath error: %+v", err)
+	}
+	if err := v.CheckAll(); err != nil {
+		levelErrors, err := specerror.SplitLevel(err, rfc2119.Must)
+		if err != nil {
+			t.Errorf("unexpected non-multierror: %+v", err)
+			return
+		}
+		for _, e := range levelErrors.Warnings {
+			t.Logf("unexpected warning: %v", e)
+		}
+		if err := levelErrors.Error; err != nil {
+			t.Errorf("unexpected MUST error(s): %+v", err)
+		}
+	}
+}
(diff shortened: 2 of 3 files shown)


cyphar:
CI failure is caused by a build failure I cannot reproduce on my local machine.

wking:
Build failures were fixed by #490.

The check you're patching is from #374, and I think we can drop at least the `Type` check in favor of JSON Schema (with opencontainers/runtime-spec#690 hopefully landing soon ;).  But if we want localcode until that lands, this change looks fine to me.

cyphar:
I've commented on opencontainers/runtime-spec#690, which currently has the same bug as the one I'm fixing here. But since it might take a while to land in runtime-spec, we should fix this bug as currently our validation isn't even consistent with our own defaults (which _really_ shouldn't have happened in the first place -- don't we test that our own default configuration is compliant?).

wking:
> ... don't we test that our own default configuration is compliant?

+1 to unit tests for this code, runtime-tools is currently sparsely-tested.

I'm fine with either this PR or dropping the local `Type` check with the expectionthat opencontainers/runtime-spec#690 will land soon.  I don't think anyone is in favor of keeping the overly-strict local test we have in master now.

cyphar:
Rebased, added the smoke test, and corrected another bug in our default configuration.

cyphar:
Rebased, added the smoke test, and corrected another bug in our default configuration.

validate/validate.go
```diff
@@ -206,11 +206,6 @@ func (v *Validator) CheckRoot() (errs error) {
 		return
 	}
 
-	if filepath.Base(v.spec.Root.Path) != "rootfs" {
-		errs = multierror.Append(errs,
-			specerror.NewError(specerror.PathName, fmt.Errorf("path name should be the conventional 'rootfs'"), rspec.Version))
```
wking: This is a [spec SHOULD][1], and removing it is why [your Travis tests fail][2].  I think we need to keep this check, and users that don't care about SHOULD-level violations can ignore those errors [like this][3].

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config.md#L41
[2]: https://travis-ci.org/opencontainers/runtime-tools/jobs/282233019#L502
[3]: https://github.com/opencontainers/runtime-tools/blob/v0.2.0/cmd/runtimetest/main.go#L767

cyphar: I know it's a SHOULD. But `CheckAll` doesn't appear to understand that SHOULD doesn't mean MUST. The following error happens:

```
=== RUN   TestGenerateValid
--- FAIL: TestGenerateValid (0.57s)
        generate_test.go:42: unexpected error(s) validating default: 3 errors occurred:

                * path name should be the conventional 'rootfs'
                Refer to: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#root
                * root.path is "", but it MUST be a child of "/tmp/TestGenerateValid_bundle231889172"
                Refer to: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/bundle.md#container-format
                * 'Root.Path' should not be empty
```

And this means that `oci-runtime-tool validate` returns a non-zero error code.

wking: > But `CheckAll` doesn't appear to understand that SHOULD doesn't mean MUST.

`CheckAll` shouldn't care; it's up to the caller to decide what is fatal or not.  Using `rootfs` in the default config should be enough for your `TestGenerateValid`.  `cmd/oci-runtime-tool/validate.go` needs a `--compliance-level` like `runtimetest` (which may be in flight somewhere, I don't have my dev box handy to grep open PRs).

wking: > `cmd/oci-runtime-tool/validate.go` needs a `--compliance-level` like `runtimetest` (which may be in flight somewhere, I don't have my dev box handy to grep open PRs).

I couldn't find this in flight, so I filed it as #492.

cyphar:
Rebased off #492.

validation/generate_test.go
```diff
@@ -0,0 +1,57 @@
...
+	// Validate the bundle.
+	v, err := validate.NewValidatorFromPath(bundle, true, runtime.GOOS)
+	if err != nil {
+		t.Errorf("unexpected NewValidatorFromPath error: %+v", err)
+	}
+	if err := v.CheckAll(); err != nil {
+		levelErrors, err2 := specerror.SplitLevel(err, rfc2119.Must)
+		if err2 != nil {
+			t.Errorf("unexpected error(s) filtering errors default: %+v", err2)
+			t.Errorf("unexpected error(s): %+v", err)
```
wking: What you call `err2` will only ever be non-`nil` if [it exactly matches the input `error`][1].  So I think you want to drop the previous line and keep only this one to avoid redundancy.  Perhaps the `SplitLevel` comment can also be improved to make this more clear?

[1]: https://github.com/opencontainers/runtime-tools/commit/929bd5e5ec3823e41c8c8cd362d30ea9645d63e1#diff-87c0b1eb29ae71890bc6a05f207759edR188

validation/generate_test.go
```diff
@@ -0,0 +1,56 @@
...
+		t.Fatal(err)
+	}
+	configPath := filepath.Join(bundle, "config.json")
+	g := generate.New()
+	if err := (&g).SaveToFile(configPath, generate.ExportOptions{Seccomp: false}); err != nil {
+		t.Fatal(err)
+	}
+
+	// Validate the bundle.
+	v, err := validate.NewValidatorFromPath(bundle, true, runtime.GOOS)
```
wking: I'm not excited about the asymmetry between unparmeterized generation and host-specific validation.  This will cerainly die if `runtime.GOOS` is not Linux.  It may die even on Linux if the host system is sufficiently ancient or the stock config becomes sufficiently demanding.  But I'm ok crossing those bridges later.

cyphar: Do you want me to set the OS explicitly?

wking: > Do you want me to set the OS explicitly?

I'm fine leaving it as it stands, as a reminder that `New` will need similar options when we get our first non-Linux user and/or stabilize the API.

Mashimiao:
LGTM

hqhq:
LGTM

PR #439, opened 2017-07-28
cmd/runtimetest/main: Use TAP diagnostics for errors

wking:
This lets us print SHOULD violations even if they're non-fatal.  It also gets us back to consumable TAP output.  I added a `/foo` entry to `defaultFS` for testing, and before this commit:

    $ ./runtimetest
    TAP version 13
    ok 1 - root filesystem
    ok 2 - hostname
    ok 3 - mounts
    not ok 4 - capabilities
    ok 5 - default symlinks
    ok 6 - default devices
    ok 7 - linux devices
    not ok 8 - linux process
    ok 9 - masked paths
    ok 10 - oom score adj
    ok 11 - read only paths
    not ok 12 - rlimits
    ok 13 - sysctls
    ok 14 - uid mappings
    ok 15 - gid mappings
    1..15
    3 errors occurred:

    * Unexpected bounding capability chown set for process
    * UID expected: 1, actual: 1000
    * Wrong rlimit value:
    $ echo $?
    1

The TAP 13 spec [doesn't cover script exit-code handling][1], but that exit code confuses prove unless you set `--ignore-exit`:

    $ prove ./runtimetest
    ./runtimetest .. 1/? 3 errors occurred:

    * Unexpected bounding capability chown set for process
    * UID expected: 1, actual: 1000
    * Wrong rlimit value:
    ./runtimetest .. Dubious, test returned 1 (wstat 256, 0x100)
    Failed 4/16 subtests

    Test Summary Report
    -------------------
    ./runtimetest (Wstat: 256 Tests: 16 Failed: 4)
      Failed tests:  4, 6, 9, 13
      Non-zero exit status: 1
    Files=1, Tests=16,  0 wallclock secs ( 0.02 usr +  0.00 sys =  0.02 CPU)
    Result: FAIL

Note the “Dubious, test returned 1”.  And with the diagnostics written to stderr, there's no way to have prove catch that.

With this commit, we're back to TAP-compatible output and exit codes, using [TAP's diagnostics][2] to share the failure details.

    $ ./runtimetest
    TAP version 13
    ok 1 - root filesystem
    ok 2 - hostname
    ok 3 - mounts
    not ok 4 - capabilities
    # Unexpected bounding capability chown set for process
    ok 5 - default symlinks
    ok 6 - default file system
    # /foo SHOULD exist and expected type is tmpfs
    # Refer to: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config-linux.md#default-filesystems
    ok 7 - default devices
[...]

PR #492, opened 2017-10-02
cmd/oci-runtime-tool: Implement --compliance-level

wking:
The man-page entry and Bash completions for this option were added in 4029999d (#354), but the option was left unimplemented.  The implementation in this commit is very similar to the existing runtimetest implementation from 4029999d, except we warn instead of silently skipping non-fatal spec violations (more on that [here][1]).

The warning (vs. error) on invalid level arguments follows the runtimetest implementation from 6316a4e0 (#354).  I'd personally prefer hard errors on invalid level arguments, but didn't want to break runtime-tools consistency over that.

[1]: https://github.com/opencontainers/runtime-tools/pull/354#discussion_r110955465

### DIFF
--- a/cmd/oci-runtime-tool/validate.go
+++ b/cmd/oci-runtime-tool/validate.go
@@ -3,7 +3,11 @@ package main
 import (
 	"fmt"
 
+	"github.com/hashicorp/go-multierror"
+	rfc2119 "github.com/opencontainers/runtime-tools/error"
+	"github.com/opencontainers/runtime-tools/specerror"
 	"github.com/opencontainers/runtime-tools/validate"
+	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 )
 
@@ -19,6 +23,12 @@ var bundleValidateCommand = cli.Command{
 	Before: before,
 	Action: func(context *cli.Context) error {
 		hostSpecific := context.GlobalBool("host-specific")
+		complianceLevelString := context.GlobalString("compliance-level")
+		complianceLevel, err := rfc2119.ParseLevel(complianceLevelString)
+		if err != nil {
+			complianceLevel = rfc2119.Must
+			logrus.Warningf("%s, using 'MUST' by default.", err.Error())
+		}
 		inputPath := context.String("path")
 		platform := context.String("platform")
 		v, err := validate.NewValidatorFromPath(inputPath, hostSpecific, platform)
@@ -27,8 +37,20 @@ var bundleValidateCommand = cli.Command{
 		}
 
 		if err := v.CheckAll(); err != nil {
-			return err
-
+			merr, ok := err.(*multierror.Error)
+			if !ok {
+				return err
+			}
+			var validationErrors error
+			for _, err = range merr.Errors {
+				e, ok := err.(*specerror.Error)
+				if ok && e.Err.Level < complianceLevel {
+					logrus.Warn(e)
+					continue
+				}
+				validationErrors = multierror.Append(validationErrors, err)
+			}
+			return validationErrors
 		}
 		fmt.Println("Bundle validation succeeded.")
 		return nil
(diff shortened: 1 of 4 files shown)


cmd/oci-runtime-tool/validate.go
```diff
@@ -27,8 +37,20 @@ var bundleValidateCommand = cli.Command{
 		}
 
 		if err := v.CheckAll(); err != nil {
-			return err
-
+			merr, ok := err.(*multierror.Error)
```
cyphar: It'd be nice if this was a helper function we could reuse in tests.

wking: > It'd be nice if this was a helper function we could reuse in tests.

We already have [`FindError`][1], which you can use [like this][2].

I've just pushed 91aa0e9 onto this series adding a `FilterError` helper, in case that's what you're looking for.

[1]: https://github.com/opencontainers/runtime-tools/blob/v0.2.0/specerror/error.go#L148-L152
[2]: https://github.com/opencontainers/runtime-tools/blob/2b007bc3e824086155eeb3cd0f32e522d56369a8/validate/validate_test.go#L170

wking: > I've just pushed 91aa0e9…

I've pushed 91aa0e9 → be12cce which flips the order of the `FilterError` response variables to keep `golint` happy (avoiding “error should be the last type when returning multiple items”).

wking:
be12cce has an API that passes `golint`.

cmd/oci-runtime-tool/validate.go
```diff
@@ -27,8 +36,13 @@ var bundleValidateCommand = cli.Command{
 		}
 
 		if err := v.CheckAll(); err != nil {
+			removed, err := specerror.FilterError(err, complianceLevel)
```
Mashimiao: I think `removed` is not so easy to understand, how about `filtered`?

wking: > I think `removed` is not so easy to understand, how about `filtered`?

I'm open to rewording, but `FilterError` is [currently][1] using `removed` for “these were removed from the input error due to their low level” and `filtered` for “the high-level and non-`specerror` stuff you still need to worry about”.  If we decide to use `filtered` for the low-level stuff, what should we use for the high-level and non-`specerror` stuff?

[1]: https://github.com/opencontainers/runtime-tools/pull/492/files#diff-87c0b1eb29ae71890bc6a05f207759edR175

Mashimiao: How about just printing `removed errors` in FilterError, FilterError just returns filtered errors?

wking: > How about just printing `removed errors` in FilterError, FilterError just returns filtered errors?

I'd rather not make UI decisions in a library function.  Maybe the caller wants to silently ignore low-level errors.  Or maybe they want to present them as TAP diagnostics (#439).  Returning the low-level error slice leaves that sort of thing up to the caller.

Mashimiao: but `removed, err := specerror.FilterError(err, complianceLevel)` looks very strange. That looks like `removed` is the expected  result, err represents error message.

wking: > but `removed, err := specerror.FilterError(err, complianceLevel)` looks very strange. That looks like removed is the expected result, err represents error message.

So maybe `low, high, alwaysNil := specerror.FilterError(...)`?

Mashimiao: I think `alwaysNil` is also not so good. How about just returning a struct includes `filtered` and `low` errors?

wking: > I think `alwaysNil` is also not so good. How about just returning a struct includes `filtered` and `low` errors?

Works for me; I'll reroll in a few hours.

wking: I've rerolled to return a helper struct in 1cbca20 → 6089f63.

Mashimiao:
LGTM

q384566678:
LGTM

PR #494, opened 2017-10-11
fix nil deference

q384566678:

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -105,6 +105,10 @@ func validateGeneralProcess(spec *rspec.Spec) error {
 }
 
 func validateLinuxProcess(spec *rspec.Spec) error {
+	if spec.Process == nil {
+		return nil
+	}
+
 	validateGeneralProcess(spec)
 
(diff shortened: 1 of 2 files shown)


cmd/runtimetest/main.go
```diff
@@ -178,7 +182,7 @@ func validateCapabilities(spec *rspec.Spec) error {
 	expectedCaps3 := make(map[string]bool)
 	expectedCaps4 := make(map[string]bool)
 	expectedCaps5 := make(map[string]bool)
-	if spec.Process.Capabilities != nil {
+	if spec.Process != nil && spec.Process.Capabilities != nil {
```
wking: I'd rather move this guard to the top of this function, and have:

```go
if spec.Process == nil || spec.Process.Capabilities == nil {
    return nil
}
```

I don't see the point to initializing `processCaps`, etc. when we aren't going to use them.

q384566678: updated, thanks.

Mashimiao:
LGTM

liangchenye:
LGTM

PR #495, opened 2017-10-19
 Add validation when host-specific is set 

### LINKED ISSUE #414: host-specific and platform overlapped in validation
While reviewing https://github.com/opencontainers/runtime-tools/issues/187,
I checked both generate and validate logic.

In validate.go, we have both 'host-specific' and 'platform',
I think we should overwrite the platform when we have 'host-specific' option.

q384566678:
Fixes #414 

I'm wondering if it's necessary to change the default value of the `--platform` to the host platform.

### DIFF
--- a/cmd/oci-runtime-tool/validate.go
+++ b/cmd/oci-runtime-tool/validate.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"runtime"
 
 	rfc2119 "github.com/opencontainers/runtime-tools/error"
 	"github.com/opencontainers/runtime-tools/specerror"
@@ -12,7 +13,7 @@ import (
 
 var bundleValidateFlags = []cli.Flag{
 	cli.StringFlag{Name: "path", Value: ".", Usage: "path to a bundle"},
-	cli.StringFlag{Name: "platform", Value: "linux", Usage: "platform of the target bundle (linux, windows, solaris)"},
+	cli.StringFlag{Name: "platform", Value: runtime.GOOS, Usage: "platform of the target bundle (linux, windows, solaris)"},
 }
 
 var bundleValidateCommand = cli.Command{
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -66,23 +66,20 @@ type Validator struct {
 }
 
 // NewValidator creates a Validator
-func NewValidator(spec *rspec.Spec, bundlePath string, hostSpecific bool, platform string) Validator {
+func NewValidator(spec *rspec.Spec, bundlePath string, hostSpecific bool, platform string) (Validator, error) {
 	if hostSpecific && platform != runtime.GOOS {
-		platform = runtime.GOOS
+		return Validator{}, fmt.Errorf("When hostSpecific is set, platform must be same as the host platform")
 	}
 	return Validator{
 		spec:         spec,
 		bundlePath:   bundlePath,
 		HostSpecific: hostSpecific,
 		platform:     platform,
-	}
+	}, nil
 }
 
 // NewValidatorFromPath creates a Validator with specified bundle path
 func NewValidatorFromPath(bundlePath string, hostSpecific bool, platform string) (Validator, error) {
-	if hostSpecific && platform != runtime.GOOS {
-		platform = runtime.GOOS
-	}
 	if bundlePath == "" {
 		return Validator{}, fmt.Errorf("bundle path shouldn't be empty")
 	}
@@ -104,7 +101,7 @@ func NewValidatorFromPath(bundlePath string, hostSpecific bool, platform string)
 		return Validator{}, err
 	}
 
-	return NewValidator(&spec, bundlePath, hostSpecific, platform), nil
+	return NewValidator(&spec, bundlePath, hostSpecific, platform)
 }
 
 // CheckAll checks all parts of runtime bundle
--- a/validate/validate_test.go
+++ b/validate/validate_test.go
@@ -23,13 +23,16 @@ func TestNewValidator(t *testing.T) {
 		val      Validator
 		expected Validator
 	}{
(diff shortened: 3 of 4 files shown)


validate/validate.go
```diff
@@ -68,6 +68,7 @@ type Validator struct {
 // NewValidator creates a Validator
 func NewValidator(spec *rspec.Spec, bundlePath string, hostSpecific bool, platform string) Validator {
 	if hostSpecific && platform != runtime.GOOS {
+		logrus.Warn("platform is overwriteen by the host platform")
```
wking: typo: “overwriteen” → “overwritten”, and again below.

In #414 I was calling for this situation to be a *fatal* “we can't do that on this host” error, not the “we can't do what you asked for so we're going to do something else instead” warning you have here ;).

wking:
> I'm wondering if it's necessary to change the default value of the `--platform` to the host platform.

I think this would certainly help.  An alternative would be to not give `--platform` a default, and to switch on `context.IsSet("platform") in `cmd/oci-runtime-tool/validate.go`.  I don't see much difference between those two approaches, so whichever makes the most sense to you.

validate/validate.go
q384566678: The role of `host-specific` is to execute host-specific validations, so I think I should change the value of the `platform`  and output the warning,  then continue to execute, instead of reporting an error and terminating the execution.

wking: > ...so I think I should change the value of the platform and output the warning, then continue to execute...

But once we fix `--platform`, any mismatch will be because the caller specifically requested host-specific validation and specifically requested the wrong platform.  I think that's a fatally-wrong request.

q384566678: >any mismatch will be because the caller specifically requested host-specific validation and specifically requested the wrong platform.

In my opinion, warning is enough. @liangchenye @Mashimiao What do you think?

Mashimiao: I'm OK to change default value of --platform.
About value of platform doesn't equal to runtime.GOOS, I think the simplest way is to return an error and give user the reason.

q384566678: Fine, updated.

validate/validate.go
```diff
@@ -66,22 +66,22 @@ type Validator struct {
...
 		platform:     platform,
-	}
+	}, nil
 }
 
 // NewValidatorFromPath creates a Validator with specified bundle path
 func NewValidatorFromPath(bundlePath string, hostSpecific bool, platform string) (Validator, error) {
 	if hostSpecific && platform != runtime.GOOS {
-		platform = runtime.GOOS
+		return Validator{}, fmt.Errorf("When hostSpecific is set, platform must be same as the host platform")
```
wking: I'm fine with this early check, as you have it.  But I'd also be fine dropping this block, and letting `NewValidator` handle the check for us.  Leaving it to `NewValidator would be slightly more DRY (good), but a slightly later check (maybe bad if you validate lots of bundles, or on wimpy hardware?).

q384566678: I think it's better to do it on this side, otherwise the other checks in `NewValidatorFromPath` would be redundant if it went wrong.

wking: > … otherwise the other checks in `NewValidatorFromPath` would be redundant if it went wrong…

I'm not clear on this point.  `NewValidator` has exactly the same check as you have here (you are touching both checks in this PR), and `NewValidatorFromPath` calls `NewValidator` internally (you're touching that line below in this PR).  So the choice is either “check for this explicitly in both `NewValidator` and `NewValidatorFromPath`” (which is what you're doing in 65846d77) or “check for this explicitly in `NewValidator`, and have `NewValidatorFromPath` check indirectly via `NewValidator`” (which is what we'd have if we dropped this block from `NewValidatorFromPath`).  As I said, I'm fine with either of those, but checking only in `NewValidator` (and dropping this block here) would be slightly more DRY.

cmd/oci-runtime-tool/validate.go
```diff
@@ -12,7 +13,7 @@ import (
 
 var bundleValidateFlags = []cli.Flag{
 	cli.StringFlag{Name: "path", Value: ".", Usage: "path to a bundle"},
-	cli.StringFlag{Name: "platform", Value: "linux", Usage: "platform of the target bundle (linux, windows, solaris)"},
+	cli.StringFlag{Name: "platform", Value: runtime.GOOS, Usage: "platform of the target bundle (linux, windows, solaris)"},
```
liangchenye: +1 for this change.

q384566678:
@Mashimiao @liangchenye  PTAL

validate/validate.go
Mashimiao: Please remove this block

q384566678: updated.

Mashimiao:
LGTM

validate/validate_test.go
```diff
@@ -23,13 +23,16 @@ func TestNewValidator(t *testing.T) {
...
 	}{
-		{Validator{testSpec, testBundle, true, testPlatform}, Validator{testSpec, testBundle, true, runtime.GOOS}},
 		{Validator{testSpec, testBundle, true, runtime.GOOS}, Validator{testSpec, testBundle, true, runtime.GOOS}},
 		{Validator{testSpec, testBundle, false, testPlatform}, Validator{testSpec, testBundle, false, testPlatform}},
 	}
 
 	for _, c := range cases {
-		assert.Equal(t, c.expected, NewValidator(c.val.spec, c.val.bundlePath, c.val.HostSpecific, c.val.platform))
+		v, err := NewValidator(c.val.spec, c.val.bundlePath, c.val.HostSpecific, c.val.platform)
+		if err != nil {
```
liangchenye: we can use this instead.
```
asset.Nil(t, err, "unexpected NewValidator error")
```

Anyway it looks good to me.

liangchenye:
LGTM

PR #497, opened 2017-10-24
complete specerror of config-linux.md config-windows.md

liangchenye:


liangchenye:
Too many RFC codes to translate so `runtime.md` is not finish yet.
It will come in the next PR.

Mashimiao:
LGTM   If there are not any vital errors, I think we can let this in. If we found any error when taking advantage of them, we can fix other PRs.

q384566678:
LGTM

specerror/config-linux.go
```diff
@@ -10,14 +10,125 @@ import (
...
+	// NSPathMatchTypeError represents "The runtime MUST generate an error if `path` is not associated with a namespace of type `type`."
+	NSPathMatchTypeError = "The runtime MUST generate an error if `path` is not associated with a namespace of type `type`."
+	// NSNewNSWithoutPath represents "If `path` is not specified, the runtime MUST create a new container namespace of type `type`."
+	NSNewNSWithoutPath = "If `path` is not specified, the runtime MUST create a new container namespace of type `type`."
+	// NSInheritWithoutType represents "If a namespace type is not specified in the `namespaces` array, the container MUST inherit the runtime namespace of that type."
+	NSInheritWithoutType = "If a namespace type is not specified in the `namespaces` array, the container MUST inherit the runtime namespace of that type."
+	// NSErrorOnDup represents "If a `namespaces` field contains duplicated namespaces with same `type`, the runtime MUST generate an error."
+	NSErrorOnDup = "If a `namespaces` field contains duplicated namespaces with same `type`, the runtime MUST generate an error."
```
wking: This is an odd one, because we [allow runtimes to skip validation][1], but I guess not here.  And we have similar wording [for rlimits][2].  But neither of those are worded quite right for a non-runtime validator like our `validate`, and it feels odd to force runtimes to error on these but allow them to handle all sorts of other broken configurations.  Maybe not worth holding up this PR, but I'll file a spec PR trying to clean this up.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/runtime.md#L108
[2]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config.md#L181

PR #507, opened 2017-11-03
runtimetest: add host platform validation

q384566678:

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -706,13 +706,16 @@ func run(context *cli.Context) error {
 	}
 	logrus.SetLevel(logLevel)
 
+	platform := runtime.GOOS
+	if platform != "linux" && platform != "solaris" && platform != "windows" {
+		return fmt.Errorf("runtime-tools has not implemented testing for your platform %q, because the spec has nothing to say about it", platform)
+	}
+
 	inputPath := context.String("path")
 	spec, err := loadSpecConfig(inputPath)
 	if err != nil {
 		return err
 	}
-
-	platform := runtime.GOOS
 
 	defaultValidations := []validation{
 		{


cmd/runtimetest/main.go
```diff
@@ -320,6 +320,14 @@ func testWriteAccess(path string) error {
 	return nil
 }
 
+func validatePlatform(spec *rspec.Spec) error {
+	if runtime.GOOS != "linux" && runtime.GOOS != "solaris" && runtime.GOOS != "windows" {
+		return fmt.Errorf("host platform %q is not supported", runtime.GOOS)
```
wking: This is presumably based on [this spec section][1], but there's no RFC 2119 language there, and the platform target is closer to selecting the relevant spec than it is to being a requirement of the spec.  I'm fine with having something like this being a fatal error (i.e. “runtime-tools has not implemented testing for your platform [because the spec has nothing to say about it][1]”).  But I don't see a point to making it soft error after which we continue on to run our other tests.  If we're comfortable running the other tests, we should just run them, and let each test take care of skipping all the things that are specific to one of the above platforms (which is most things).

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/spec.md#platforms

q384566678: updated, PTAL.

cmd/runtimetest/main.go
```diff
@@ -713,6 +721,9 @@ func run(context *cli.Context) error {
 	}
 
 	platform := runtime.GOOS
+	if err := validatePlatform(platform); err != nil {
```
wking: This can be moved before `loadSpecConfig`.  Non need to hit the disk at all for unsupported platforms.

q384566678: updated. PTAL.

cmd/runtimetest/main.go
```diff
@@ -698,6 +698,14 @@ func validateMounts(spec *rspec.Spec) error {
 	return mountErrs
 }
 
+func validatePlatform(platform string) error {
+	if platform != "linux" && platform != "solaris" && platform != "windows" {
+		return fmt.Errorf("runtime-tools has not implemented testing for your platform %q, because the spec has nothing to say about it", platform)
```
Mashimiao: This is hard to say whether right or not.
Though spec don't have specific configs for those platform, But spec has common configs for all platform.
And spec seems never say it just supports those three platforms.

liangchenye:
I think this judgement is useful, testing on an unknown platform can not prove anything, either right or wrong. We can just abort the testing.

I think we should not name it 'validatePlatform', it is misleading.  It sounds like there is a RFC requirement in the spec and we are validating it. But in fact, we are just making sure `runtime-tool` works as expected.

I suggest we can just do it in the `run` function.
```
func run(context *cli.Context) error {
        ....

	platform := runtime.GOOS	
        if platform != "linux" && platform != "solaris" && platform != "windows" {
		return fmt.Errorf("runtime-tools has not implemented testing for your platform %q, because the spec has nothing to say about it", platform)
	}
}
```

q384566678:
updated, thanks.

liangchenye:
LGTM

Mashimiao:
LGTM

PR #508, opened 2017-11-03
runtimetest: Raise ConfigInRootBundleDir for missing config.json

wking:
This is a bit tricky, because before we load the config, we don't know the target spec version.  If a future spec version relaxes the current requirement, we'll have adjust runtimetest to tell it what version it's testing (where I've currently hard-coded 1.0.0).

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -68,7 +68,7 @@ func loadSpecConfig(path string) (spec *rspec.Spec, err error) {
 	cf, err := os.Open(configPath)
 	if err != nil {
 		if os.IsNotExist(err) {
-			return nil, fmt.Errorf("%s not found", specConfig)
+			return nil, specerror.NewError(specerror.ConfigInRootBundleDir, err, rspec.Version)
 		}
 
 		return nil, err


cmd/runtimetest/main.go
```diff
@@ -68,7 +68,7 @@ func loadSpecConfig(path string) (spec *rspec.Spec, err error) {
 	cf, err := os.Open(configPath)
 	if err != nil {
 		if os.IsNotExist(err) {
-			return nil, fmt.Errorf("%s not found", specConfig)
+			return nil, specerror.NewError(specerror.ConfigInRootBundleDir, err, "1.0.0")
```
q384566678: I think we still need to use `rspec.Version`. If the specification changes and becomes feasible, we can delete this error. Validation is performed according to the latest specification.

wking: > Validation is performed according to the latest specification.

I don't think we want to be that quick to drop old versions.  Once the spec cuts 1.1.0, we still want to be able to certify new runtimes as 1.0.0 compliant, even if they aren't yet 1.1.0 compliant.  And we want to continue to ensure that 1.1.0 runtimes appropriately handle 1.0.0 configs.

However, if/when the spec drops the `config.json` requirement, I'm fine dropping this block altogether, and having this error fall through as a non-spec “you called us wrong” error.  So updated to use `rspec.Version` in 7d2742b → 6c98b46.

q384566678:
LGTM

Mashimiao:
LGTM

PR #510, opened 2017-11-03
 runtimetest: add posixValidations

q384566678:

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -81,11 +81,7 @@ func loadSpecConfig(path string) (spec *rspec.Spec, err error) {
 	return spec, nil
 }
 
-func validateUser(spec *rspec.Spec) error {
-	if runtime.GOOS == "windows" {
-		return nil
-	}
-
+func validatePosixUser(spec *rspec.Spec) error {
 	if spec.Process == nil {
 		return nil
 	}
@@ -118,7 +114,6 @@ func validateUser(spec *rspec.Spec) error {
 	return nil
 }
 
-// should be included by other platform specified process validation
 func validateProcess(spec *rspec.Spec) error {
 	if spec.Process.Cwd != "" {
 		cwd, err := os.Getwd()
@@ -277,10 +272,6 @@ func validateHostname(spec *rspec.Spec) error {
 }
 
 func validateRlimits(spec *rspec.Spec) error {
-	if runtime.GOOS == "windows" {
-		return nil
-	}
-
 	if spec.Process == nil {
 		return nil
 	}
@@ -642,12 +633,7 @@ func mountMatch(configMount rspec.Mount, sysMount *mount.Info) error {
 	return nil
 }
 
-func validateMounts(spec *rspec.Spec) error {
-	if runtime.GOOS == "windows" {
-		logrus.Warnf("mounts validation not yet implemented for OS %q", runtime.GOOS)
-		return nil
-	}
-
+func validatePosixMounts(spec *rspec.Spec) error {
 	mountInfos, err := mount.GetMounts()
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


cmd/runtimetest/main.go
```diff
@@ -82,7 +82,7 @@ func loadSpecConfig(path string) (spec *rspec.Spec, err error) {
 }
 
 // should be included by other platform specified process validation
```
wking: The “fix process validation” commit should remove this comment too.

cmd/runtimetest/main.go
```diff
@@ -81,32 +81,11 @@ func loadSpecConfig(path string) (spec *rspec.Spec, err error) {
...
-		parts := strings.Split(env, "=")
-		key := parts[0]
-		expectedValue := parts[1]
-		actualValue := os.Getenv(key)
-		if actualValue != expectedValue {
-			return fmt.Errorf("Env %v expected: %v, actual: %v", key, expectedValue, actualValue)
-		}
+func validateUser(spec *rspec.Spec) error {
+	if runtime.GOOS == "windows" {
+		return nil
```
wking: Perhaps this function should be `validatePOSIXUser` to [match the spec][1] and then used in both `linuxValidatations` and a new `solarisValidations`?  Or a new `posixValidations` which is included for both `linux` and `solaris` platforms?  The spec *has* [Windows-specific user stuff][2], but all of its RFC 2119 language is currently config-oriented and covered by the JSON Schema validation.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#posix-platform-user
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#windows-user

q384566678: >Or a new posixValidations which is included for both linux and solaris platforms?

I think it would be better. I'm going to do that.

q384566678: updated, PTAL.

cmd/runtimetest/main.go
```diff
@@ -82,10 +82,6 @@ func loadSpecConfig(path string) (spec *rspec.Spec, err error) {
 }
 
 func validateUser(spec *rspec.Spec) error {
```
wking: This should be `validatePosixUser` to make it extra-clear that it's not covering [Windows users][1].

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#windows-user

cmd/runtimetest/main.go
```diff
@@ -643,11 +634,6 @@ func mountMatch(configMount rspec.Mount, sysMount *mount.Info) error {
 }
 
 func validateMounts(spec *rspec.Spec) error {
-	if runtime.GOOS == "windows" {
-		logrus.Warnf("mounts validation not yet implemented for OS %q", runtime.GOOS)
```
wking: We want to keep this.  [`mounts` is specified for all platforms][1], and we just aren't testing the Windows-specific parts of that yet.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#mounts

q384566678: I think this can  change to `validatePosixMounts`. Because this is only validated against [POSIX-platform Mounts](https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#posix-platform-mounts).

wking: > I think this can change to `validatePosixMounts`. Because this is only validated against [POSIX-platform Mounts][1].

It's also [checking the mount order][2], which [is a cross-platform requirement][3].  Windows checks for mount order could involve extending `cmd/runtimetest/mount` to support Windows, in which case we'd want to keep this valdation generic, with per-platform guards for the [POSIX-specific `type` checks][1], etc.  Or we can duplicate the order checks in a separate `validateWindowsMounts`, in which case calling all of the current logic `validatePosixMounts` is fine.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#posix-platform-mounts
[2]: https://github.com/opencontainers/runtime-tools/blob/eead1367811940eef85d4d953e5435abd2bccdd4/cmd/runtimetest/main.go#L685
[3]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#mounts

liangchenye:
LGTM

Mashimiao:
LGTM

PR #511, opened 2017-11-03
runtimetest: add rootfs propagation test

Mashimiao:

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -24,6 +24,8 @@ import (
 	"github.com/opencontainers/runtime-tools/cmd/runtimetest/mount"
 	rfc2119 "github.com/opencontainers/runtime-tools/error"
 	"github.com/opencontainers/runtime-tools/specerror"
+
+	"golang.org/x/sys/unix"
 )
 
 // PrGetNoNewPrivs isn't exposed in Golang so we define it ourselves copying the value from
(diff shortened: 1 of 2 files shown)


cmd/runtimetest/main.go
```diff
@@ -778,6 +843,14 @@ func run(context *cli.Context) error {
 			description: "read only paths",
 		},
 		{
+			test:        validateRootfsPropagation,
+			description: "rootfs propagation",
+		},
+		{
+			test:        validateRlimits,
```
wking: Duplicate entry?

```
$ git grep -n validateRlimits origin/pr/511
origin/pr/511:cmd/runtimetest/main.go:271:func validateRlimits(spec *rspec.Spec) error {
origin/pr/511:cmd/runtimetest/main.go:803:                      test:        validateRlimits,
origin/pr/511:cmd/runtimetest/main.go:850:                      test:        validateRlimits,
```

cmd/runtimetest/main.go
```diff
@@ -342,6 +344,69 @@ func validateRootFS(spec *rspec.Spec) error {
...
+		return nil
+	}
+
+	targetDir, err := ioutil.TempDir("/", "target")
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(targetDir)
+
+	switch spec.Linux.RootfsPropagation {
```
wking: I think we probably want a `default` case that logs an “unrecognized linux.rootfsPropagation: …” warning.

cmd/runtimetest/main.go
Mashimiao: Oh, this is mistake, updated. Thanks.

cmd/runtimetest/main.go
```diff
@@ -342,6 +344,71 @@ func validateRootFS(spec *rspec.Spec) error {
 	return nil
 }
 
+func validateRootfsPropagation(spec *rspec.Spec) error {
+	if spec.Linux == nil {
+		return nil
+	}
+
+	targetDir, err := ioutil.TempDir("/", "target")
```
q384566678: Can you change the create path? Perhaps `ioutil.TempDir("", "target")`? If you do not change, `runtimetest` must use `sudo`.

Mashimiao:
On 11/04/2017 12:36 PM, W. Trevor King wrote:
> These tests don't seem to distinguish between |slave| and |private|.

Yes, I have to admit that we can't in this PR.
Because we can't get the mount source of rootfs.
So, we can just judge propagation by binding rootfs.
Unfortunately, we get the same behaviour with slave and private rootfs after binding them.

Maybe we can find a way and fix this in the future.

liangchenye:
According to https://www.kernel.org/doc/Documentation/filesystems/sharedsubtree.txt,
This PR does tests 'share and unbind' mode. 
I think we can remove `TestValidateRootfsPropagationPrivate` and `TestValidateRootfsPropagationSlave`.

liangchenye:
The 'share' and 'unbind' looks good to me, just need to rebase.

liangchenye:
LGTM

q384566678:
LGTM

PR #512, opened 2017-11-06
add more test cases for default runtime validation

Mashimiao:

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -336,6 +336,11 @@ func validateRootFS(spec *rspec.Spec) error {
 		if err == nil {
 			return specerror.NewError(specerror.RootReadonlyImplement, fmt.Errorf("rootfs must be readonly"), rspec.Version)
 		}
+	} else {
+		err := testWriteAccess("/")
+		if err != nil {
+			return specerror.NewError(specerror.RootReadonlyImplement, fmt.Errorf("rootfs must not be readonly"), rspec.Version)
+		}
 	}
 
 	return nil
--- a/validation/validation_test.go
+++ b/validation/validation_test.go
@@ -88,6 +88,49 @@ func TestValidateBasic(t *testing.T) {
 	assert.Nil(t, runtimeInsideValidate(g))
 }
 
+// Test whether rootfs Readonly can be applied as false
+func TestValidateRootFSReadWrite(t *testing.T) {
+	g := getDefaultGenerator()
+	g.SetRootReadonly(false)
+
+	assert.Nil(t, runtimeInsideValidate(g))
+}
+
+// Test whether rootfs Readonly can be applied as true
+func TestValidateRootFSReadonly(t *testing.T) {
+	if "windows" == runtime.GOOS {
+		t.Skip("skip this test on windows platform")
+	}
+
+	g := getDefaultGenerator()
+	g.SetRootReadonly(true)
+
+	assert.Nil(t, runtimeInsideValidate(g))
+}
+
+// Test whether hostname can be applied or not
+func TestValidateHostname(t *testing.T) {
+	g := getDefaultGenerator()
+	g.SetHostname("hostname-specific")
+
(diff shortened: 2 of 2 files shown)


validation/validation_test.go
```diff
@@ -88,13 +88,44 @@ func TestValidateBasic(t *testing.T) {
...
+func TestValidateHostname(t *testing.T) {
+	g := getDefaultGenerator()
+	g.SetHostname("hostname-specific")
+
+	assert.Nil(t, runtimeInsideValidate(g))
+}
+
+// Test whether mounts are correctlt mounted
+func TestValidateMounts(t *testing.T) {
+	// TODO mounts generation options have not been implemented
```
liangchenye: we can add this test when 'mounts generate' will be done.

validation/validation_test.go
```diff
@@ -88,13 +88,44 @@ func TestValidateBasic(t *testing.T) {
 	assert.Nil(t, runtimeInsideValidate(g))
 }
 
+// Test whether rootfs Readonly can be applied or not
+func TestValidateRootFS(t *testing.T) {
+	g := getDefaultGenerator()
+	g.SetRootReadonly(true)
```
liangchenye: Maybe we can add another test like:
```
g.SetRootReadonly(false)
```

Mashimiao: Added, PTAL

validation/validation_test.go
Mashimiao: Added such TODO comment

validation/validation_test.go
```diff
@@ -88,13 +88,53 @@ func TestValidateBasic(t *testing.T) {
...
+	g := getDefaultGenerator()
+	g.SetRootReadonly(false)
+
+	assert.Nil(t, runtimeInsideValidate(g))
+}
+
+// Test whether rootfs Readonly can be applied as true
+func TestValidateRootFS(t *testing.T) {
+	g := getDefaultGenerator()
+	g.SetRootReadonly(true)
```
wking: I think we need to skip this on Windows, where [the spec forbids setting `readonly` to `true`][1].  Also, you probably need to use more-specific function names:

```
$ git grep '^func TestValidateRootFS' origin/pr/512
origin/pr/512:validation/validation_test.go:func TestValidateRootFS(t *testing.T) {
origin/pr/512:validation/validation_test.go:func TestValidateRootFS(t *testing.T) {
```

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config.md#L46

validation/validation_test.go
```diff
@@ -88,13 +88,53 @@ func TestValidateBasic(t *testing.T) {
...
+
+// Test whether hostname can be applied or not
+func TestValidateHostname(t *testing.T) {
+	g := getDefaultGenerator()
+	g.SetHostname("hostname-specific")
+
+	assert.Nil(t, runtimeInsideValidate(g))
+}
+
+// Test whether mounts are correctlt mounted
```
wking: “correctlt” → “correctly”.  Or just use `// Validate mounts`, or leave the comments off?  `TestValidateMounts` sounds pretty clear to me even without a comment.

Mashimiao:
updated. @wking  @liangchenye  @q384566678  PTAL

liangchenye:
LGTM

cmd/runtimetest/main.go
```diff
@@ -336,6 +336,11 @@ func validateRootFS(spec *rspec.Spec) error {
 		if err == nil {
 			return specerror.NewError(specerror.RootReadonlyImplement, fmt.Errorf("rootfs must be readonly"), rspec.Version)
 		}
+	} else {
+		err := testWriteAccess("/")
+		if err == nil {
```
q384566678: Should use `err != nil`.

Mashimiao: good catch, updated

q384566678:
LGTM

liangchenye:
LGTM

PR #308, opened 2017-01-17
runtimetest: Make TAP output more granular

wking:
Granular TAP output makes it easier to see what's being tested (and what's going wrong).  You could feed this into any TAP harness you like, but I've chosen prove (it seems popular).  You need a TAP harness looking for test failures, because runtimetest now returns zero when it successfully runs the tests (regardless of whether the tests all pass).  To avoid confusing the TAP consumer, all of the non-TAP logging goes to stderr (and not stdout).

With this change, the verbose output looks like:

    # prove -v ./test_runtime.sh :: -r ~/bin/runc
    ./test_runtime.sh ..
    TAP version 13
    ok 1 # SKIP root.readonly falsy
    ok 2 - hostname matches expected value
    ok 3 - mount "/proc" found
    ok 4 - mount "/dev" found
    ok 5 - mount "/dev/pts" found
    ok 6 - mount "/dev/shm" found
    ok 7 - mount "/dev/mqueue" found
    ok 8 - mount "/sys" found
    ok 9 - expected capability CAP_CHOWN set
    ok 10 - expected capability CAP_DAC_OVERRIDE set
    ok 11 - unexpected capability CAP_DAC_READ_SEARCH not set
    ok 12 - expected capability CAP_FOWNER set
    ok 13 - expected capability CAP_FSETID set
    ok 14 - expected capability CAP_KILL set
    ok 15 - expected capability CAP_SETGID set
    ok 16 - expected capability CAP_SETUID set
    ok 17 - expected capability CAP_SETPCAP set
    ok 18 - unexpected capability CAP_LINUX_IMMUTABLE not set
    ok 19 - expected capability CAP_NET_BIND_SERVICE set
    ok 20 - unexpected capability CAP_NET_BROADCAST not set
    ok 21 - unexpected capability CAP_NET_ADMIN not set
    ok 22 - expected capability CAP_NET_RAW set
    ok 23 - unexpected capability CAP_IPC_LOCK not set
    ok 24 - unexpected capability CAP_IPC_OWNER not set
    ok 25 - unexpected capability CAP_SYS_MODULE not set
    ok 26 - unexpected capability CAP_SYS_RAWIO not set
    ok 27 - expected capability CAP_SYS_CHROOT set
    ok 28 - unexpected capability CAP_SYS_PTRACE not set
    ok 29 - unexpected capability CAP_SYS_PACCT not set
    ok 30 - unexpected capability CAP_SYS_ADMIN not set
[...]

PR #514, opened 2017-11-09
runtimetest: add validateSeccomp

q384566678:

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -576,6 +576,30 @@ func validateMaskedPaths(spec *rspec.Spec) error {
 	return nil
 }
 
+func validateSeccomp(spec *rspec.Spec) error {
+	if spec.Linux == nil || spec.Linux.Seccomp == nil {
+		return nil
+	}
+	t := tap.New()
+	for _, sys := range spec.Linux.Seccomp.Syscalls {
+		if sys.Action == "SCMP_ACT_ERRNO" {
+			for _, name := range sys.Names {
+				if name == "getcwd" {
+					_, err := os.Getwd()
+					if err == nil {
+						t.Diagnostic("getcwd did not return an error")
+					}
+				} else {
+					t.Skip(1, fmt.Sprintf("%s syscall returns errno", name))
+				}
+			}
+		} else {
+			t.Skip(1, fmt.Sprintf("syscall action %s", sys.Action))
+		}
+	}
+	return nil
+}
+
 func validateROPaths(spec *rspec.Spec) error {
 	if spec.Linux == nil {
 		return nil
@@ -863,6 +887,10 @@ func run(context *cli.Context) error {
 		{
 			test:        validateOOMScoreAdj,
 			description: "oom score adj",
+		},
+		{
+			test:        validateSeccomp,
+			description: "seccomp",
 		},
 		{
 			test:        validateROPaths,
--- a/validation/linux_seccomp.go
+++ b/validation/linux_seccomp.go
@@ -1,0 +1,20 @@
+package main
+
+import (
+	"github.com/opencontainers/runtime-tools/generate/seccomp"
+	"github.com/opencontainers/runtime-tools/validation/util"
+)
+
+func main() {
+	g := util.GetDefaultGenerator()
+	syscallArgs := seccomp.SyscallOpts{
+		Action:  "errno",
+		Syscall: "getcwd",
+	}
+	g.SetDefaultSeccompAction("allow")
+	g.SetSyscallAction(syscallArgs)
+	err := util.RuntimeInsideValidate(g, nil)
+	if err != nil {
+		util.Fatal(err)
+	}
+}


cmd/runtimetest/main.go
```diff
@@ -493,6 +493,25 @@ func validateMaskedPaths(spec *rspec.Spec) error {
...
+	if spec.Linux == nil || spec.Linux.Seccomp == nil {
+		return nil
+	}
+	for _, sys := range spec.Linux.Seccomp.Syscalls {
+		if sys.Action == "SCMP_ACT_ERRON" {
+			for _, name := range sys.Names {
+				if name == "getcwd" {
+					_, err := os.Getwd()
+					if err == nil {
+						return fmt.Errorf("when action is %v, return value should being passed to user space as the errno value without executing the system call", sys.Action)
```
wking: I don't see [any runtime-oriented RFC 2119 requirements for `linux.seccomp`][1], which means [we only have the config-oriented `SeccSyscallsNamesRequired` in `specerror`][2], which means we have no grounds for making this a fatal error in compliance testing.  I think the best we can do (at least for 1.0.x configs) is warn the caller that the runtime is skipping some not-strictly-required-but-probably-expected functionality.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/config-linux.md#seccomp
[2]: https://github.com/opencontainers/runtime-tools/blob/v0.3.0/specerror/config-linux.go#L131

validation/validation_test.go
```diff
@@ -88,6 +89,21 @@ func TestValidateBasic(t *testing.T) {
 	assert.Nil(t, runtimeInsideValidate(g))
 }
 
+func TestValidateSeccomp(t *testing.T) {
+	g := getDefaultGenerator()
+	syscallArgs := seccomp.SyscallOpts{
+		Action:   "errno",
+		Syscall:  "getcwd",
+		Index:    "0",
```
Mashimiao: what are `Index`, `Value` used for in this case?

q384566678: According to the code in [`parse_action.go`](https://github.com/opencontainers/runtime-tools/blob/master/generate/seccomp/parse_action.go#L32-L36):

```
	action, _ := parseAction(arguments[0])
	if action == config.DefaultAction && args.argsAreEmpty() {
		// default already set, no need to make changes
		return nil
	}
```

If args is empty, you can't add syscall.

Mashimiao: If you really want to test the Action, you should change the defautAction not equal to the specified action.
Then, empty args will not be a problem.

cmd/runtimetest/main.go
Mashimiao: Before spec adds RFC requirements, we'd better just give a warn instead of error

validation/validation_test.go
q384566678: updated.

cmd/runtimetest/main.go
```diff
@@ -493,6 +493,25 @@ func validateMaskedPaths(spec *rspec.Spec) error {
...
+	if spec.Linux == nil || spec.Linux.Seccomp == nil {
+		return nil
+	}
+	for _, sys := range spec.Linux.Seccomp.Syscalls {
+		if sys.Action == "SCMP_ACT_ERRON" {
+			for _, name := range sys.Names {
+				if name == "getcwd" {
+					_, err := os.Getwd()
+					if err == nil {
+						logrus.Warnf("when action is %v, return value should being passed to user space as the errno value without executing the system call", sys.Action)
```
Mashimiao: I think the warning is not suitable, that's not what we really want to tell users.
We want to tell them even seccomp  is configured but this runtime can not apply it or similar warning

Mashimiao:
LGTM

cmd/runtimetest/main.go
```diff
@@ -576,6 +576,25 @@ func validateMaskedPaths(spec *rspec.Spec) error {
...
 }
 
+func validateSeccomp(spec *rspec.Spec) error {
+	if spec.Linux == nil || spec.Linux.Seccomp == nil {
+		return nil
+	}
+	for _, sys := range spec.Linux.Seccomp.Syscalls {
+		if sys.Action == "SCMP_ACT_ERRON" {
+			for _, name := range sys.Names {
+				if name == "getcwd" {
```
wking: Can we write `t.Skip`s for other names and actions?  I don't want to give users the false impression that we're testing them by silently ignoring other values.

cmd/runtimetest/main.go
```diff
@@ -576,6 +576,25 @@ func validateMaskedPaths(spec *rspec.Spec) error {
...
+					_, err := os.Getwd()
+					if err == nil {
+						logrus.Warnf("Syscall action %v can not be properly applied in the runtime", sys.Action)
```
wking: This is a non-TAP warning that will be [gobbled by `RuntimeInsideValidate`][1].  I think you want something like:

```go
t.Skip(1, fmt.Sprintf("%s syscall returns errno", sys.Name))
if err == nil {
  t.Diagnostic("did not return an error")
} else {
  t.Diagnosticf("returned %d (%s)", err.WhateverOtherErrno(), err.Error())
}
```

[1]: https://github.com/opencontainers/runtime-tools/blob/c3755c1175a1baad7230dfb03751e3a4b7182f32/validation/util/test.go#L131-L142

cmd/runtimetest/main.go
```diff
@@ -576,6 +576,25 @@ func validateMaskedPaths(spec *rspec.Spec) error {
 	return nil
 }
 
+func validateSeccomp(spec *rspec.Spec) error {
+	if spec.Linux == nil || spec.Linux.Seccomp == nil {
+		return nil
+	}
+	for _, sys := range spec.Linux.Seccomp.Syscalls {
+		if sys.Action == "SCMP_ACT_ERRON" {
```
wking: `SCMP_ACT_ERRON` → [`SCMP_ACT_ERRNO`][1].

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/config-linux.md#L549

cmd/runtimetest/main.go
wking: > Can we write `t.Skip`s…

This is probably blocking on passing the [TAP instance][1] down into the validators, which I do in #308.  I don't see a clean way to handle this case before #308 lands; if maintainers want a stub in the meantime that's fine, and I don't really care if/how this gets handled in the stub.

[1]: https://github.com/opencontainers/runtime-tools/blob/c3755c1175a1baad7230dfb03751e3a4b7182f32/cmd/runtimetest/main.go#L889

q384566678: Does that mean that all the action and name combinations will be exported as skipped information? Maybe I misunderstood.

wking: > Does that mean that all the action and name combinations will be exported as skipped information?

Yes.  That would make it easy for folks writing tests in `validation` to see what `runtimetest` was ignoring.

cmd/runtimetest/main.go
q384566678: I tried to modify it, what do you think?

cmd/runtimetest/main.go
```diff
@@ -576,6 +576,28 @@ func validateMaskedPaths(spec *rspec.Spec) error {
 	return nil
 }
 
+func validateSeccomp(spec *rspec.Spec) error {
+	if spec.Linux == nil || spec.Linux.Seccomp == nil {
+		return nil
+	}
+	t := tap.New()
```
wking: This TAP-within-a-TAP approach is a bit strange, but I think it might work…  We can drop this once #308 lands and the `t` instance is getting passed around.

cmd/runtimetest/main.go
```diff
@@ -576,6 +576,28 @@ func validateMaskedPaths(spec *rspec.Spec) error {
...
+	for _, sys := range spec.Linux.Seccomp.Syscalls {
+		if sys.Action == "SCMP_ACT_ERRNO" {
+			for i, name := range sys.Names {
+				if name == "getcwd" {
+					_, err := os.Getwd()
+					if err == nil {
+						t.Diagnostic("Syscall action ERRNO can not be properly applied in the runtime")
```
wking: Maybe the diagnostic could be:

> getcwd did not return an error

Once #308 lands with a local test, that test will be the `%s syscall returns errno` you have in the skip below, so the total output will be:

```
not ok … - getcwd syscall returns errno
# getcwd did not return an error
```

cmd/runtimetest/main.go
```diff
@@ -576,6 +576,28 @@ func validateMaskedPaths(spec *rspec.Spec) error {
...
+					}
+				} else {
+					t.Skip(i, fmt.Sprintf("%s syscall returns errno", name))
```
wking: This should be `t.Skip(1, …)`, since we're only skipping one test.

cmd/runtimetest/main.go
```diff
@@ -576,6 +576,28 @@ func validateMaskedPaths(spec *rspec.Spec) error {
...
+				}
+			}
+		}
```
wking: And here we want an `else` case saying that we're skipping the action:

```
} else {
    t.Skip(1, fmt.Sprintf("syscall action %s", sys.Action))
}
```

or similar.

cmd/runtimetest/main.go
q384566678: Some other modifications can wait until #308 land. Thank you for your advice.

q384566678:
@liangchenye @Mashimiao PTAL

Mashimiao:
LGTM

q384566678:
ping @liangchenye

liangchenye:
LGTM

PR #516, opened 2017-11-14
fix ParseLevel

q384566678:

### DIFF
--- a/error/error.go
+++ b/error/error.go
@@ -67,15 +67,15 @@ func ParseLevel(level string) (Level, error) {
 	case "SHOULDNOT":
 		fallthrough
 	case "RECOMMENDED":
-		fallthrough
-	case "NOTRECOMMENDED":
 		return Should, nil
+	case "NOTRECOMMENDED":
+		fallthrough
 	case "MUST":
 		fallthrough
 	case "MUSTNOT":
 		fallthrough
 	case "SHALL":
-		fallthrough
+		return Must, nil
 	case "SHALLNOT":
 		fallthrough
 	case "REQUIRED":


error/error.go
```diff
@@ -67,15 +67,15 @@ func ParseLevel(level string) (Level, error) {
...
 		return Should, nil
+	case "NOTRECOMMENDED":
+		fallthrough
 	case "MUST":
 		fallthrough
 	case "MUSTNOT":
 		fallthrough
 	case "SHALL":
-		fallthrough
+		return Must, nil
```
wking: I don't think we need to make this change.  As it stands in master, `SHALL` falls through to `SHALLNOT`, which falls through to `REQUIRED`, which returns `Must`.  That's [the appropriate level for `SHALL`][1].

Hard-coding `Must` here is fine, but if we want to go with that approach I'd recommend dropping `fallthrough` from this whole `switch` for consistency.

[1]: https://tools.ietf.org/html/rfc2119#section-1

error/error.go
```diff
@@ -67,15 +67,15 @@ func ParseLevel(level string) (Level, error) {
 	case "SHOULDNOT":
 		fallthrough
 	case "RECOMMENDED":
-		fallthrough
-	case "NOTRECOMMENDED":
 		return Should, nil
+	case "NOTRECOMMENDED":
+		fallthrough
```
wking: I don't think we want to make this change. As it stands in master, `NOTRECOMMENDED` returns `Should`.  That's [the appropriate level for `NOTRECOMMENDED`][1].  With your change, `NOTRECOMMENDED` would return `Must`, and that's too strong.

[1]: https://tools.ietf.org/html/rfc2119#section-4

error/error.go
q384566678: Sorry, I made a mistake.

PR #518, opened 2017-11-15
runtimetest: remove debug info 

q384566678:

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -342,8 +342,6 @@ func validateRootFS(spec *rspec.Spec) error {
 }
 
 func validateDefaultFS(spec *rspec.Spec) error {
-	logrus.Debugf("validating linux default filesystem")
-
 	mountInfos, err := mount.GetMounts()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/runtimetest/main.go
```diff
@@ -107,6 +107,8 @@ func validateGeneralProcess(spec *rspec.Spec) error {
 }
 
 func validateLinuxProcess(spec *rspec.Spec) error {
+	logrus.Debugf("validating linux process")
```
liangchenye: I think we can just use `logrus.Debug` .

Mashimiao:
We don't need this. 
Before we had logrus like this, but removed. Because it is duplicate with tap's output

wking:
On Wed, Nov 15, 2017 at 10:02:43AM +0000, Ma Shimiao wrote:
> Because it is duplicate with tap's output

More on this front in #308 and #439, both of which are currently
milestoned for v0.4.0.

q384566678:
Finally, I think it's better to delete debug information at present.

liangchenye:
LGTM

Mashimiao:
LGTM

PR #523, opened 2017-11-23
runtimetest: fix nil dereference

q384566678:

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -125,6 +125,10 @@ func validatePosixUser(spec *rspec.Spec) error {
 }
 
 func validateProcess(spec *rspec.Spec) error {
+	if spec.Process == nil {
+		return nil
+	}
+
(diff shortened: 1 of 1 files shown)


Mashimiao:
In logic, this is a correct fix. But in fact, this is impossible, because if process is nil, the container can not be run.

wking:
On Thu, Nov 23, 2017 at 07:07:31AM +0000, Ma Shimiao wrote:
> In logic, this is a correct fix. But in fact, this is impossible,
> because if process is nil, the container can not be run.

I sometimes call runtimetest in my host environment (i.e. without a
wrapping runtime invocation).  This guard will help in that case,
although I'm also fine with panics there, since that's not how
runtimetest is intended to be used.  So I have no opinion on this PR,
but maintainers might want to consider that use case when deciding on
it.

Mashimiao:
LGTM

liangchenye:
LGTM

Issue #470, opened 2017-09-10
Unable to escape colons in option lists


gfiedler:
I want to create a hook like this

    "hooks": {
      "prestart": [ {
        "path": "cmd",
        "args": [ "cmd", "foo:bar" ]
     } ] }

but I'm unable to do this, because colons have a special semantics and there is no posibility to escape colons in option lists:

    oci-runtime-tool generate --hooks-prestart=cmd:foo:bar

results in

    "args": [ "cmd", "foo", "bar" ]

The problem is in `generate.go` in `func parseHook`.

But maybe options with path arguments like `--mount-bind` are related to this effect too?

wking:
+1 to extending the option syntax to allow for escaping.  And for things the runtime-tools doesn't (yet) support, [there's always `jq`][1] ;).

[1]: https://github.com/opencontainers/runtime-tools/pull/176#issuecomment-237326149

Mashimiao:
I‘m focusing on this problem recently.
Do you think comma works for you, as --hooks-prestart=cmd,foo:bar? @gfiedler  @wking 
It's really not easy to set separator, almost all characters can be used in path

wking:
`jq` expects you to provide the value as JSON, which covers the general case.  You could also cover the general case with a shell lexer or a custom escape syntax.  But I think trading colons for commas just punts until we have someone who needs a literal comma.

gfiedler:
I think we should leave the `--hooks-prestart` as it is and introduce a new option `--hooks-prestart-args` which works just like `--args`.

wking:
I'm fine with `--hooks-*-args` too.

Mashimiao:
Well, --hooks-*-args may can works for your case. But what if the name of hooks path contains colon?
I hope we can work out a general way to solve the problem.

wking:
On Thu, Oct 19, 2017 at 04:58:59AM +0000, Ma Shimiao wrote:
> Well, --hooks-*-args may can works for your case. But what if the
> name of hooks path contains colon?

Ah, right.  We need a way to add more than one hook entry.  And even
though path/args is confusing enough, folks may also want to set path,
args, *and* timeout (etc.) for the same hook, which is currently not
supported.  And even shell lexing wouldn't cover timeout, etc.  So I'm
back to preferring jq-esqe:

  $ oci-runtime-tool generate --hooks-prestart '{"path": "cmd", "args": ["cmd", "foo:bar"], "timeout": 5}'

[...]

PR #524, opened 2017-11-23
replace colon by comma as separator

Mashimiao:

This is a temporary try.  try to solve separator problem in options. As discussed in #470


q384566678:
In general I agree this PR.Because the `args` option is a special case. I think it's possible to only change the colon in the `parseHook`, or to change all the colons in the input to commas. Consistency is easy to use.

Mashimiao:
That's not a perfect way to solve the problem, closing it.

PR #525, opened 2017-11-28
support json value for hooks

Mashimiao:
As discussed in #470, there is not suitable separators for hooks.
Because of path may contains them.
So, decide to support json value for hooks.

### DIFF
--- a/completions/bash/oci-runtime-tool
+++ b/completions/bash/oci-runtime-tool
@@ -305,15 +305,9 @@ _oci-runtime-tool_generate() {
 		--args
 		--env
 		--env-file
-		--hooks-poststart
-		--hooks-poststart-env
-		--hooks-poststart-timeout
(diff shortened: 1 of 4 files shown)


completions/bash/oci-runtime-tool
```diff
@@ -387,6 +381,9 @@ _oci-runtime-tool_generate() {
 
 	local boolean_options="
 		--help -h
+		--hooks-poststart-remove-all
+		--hooks-poststop-remove-all
+		--hooks-prestart-remove-add
```
q384566678: `add` -> `all`

Mashimiao: updated

q384566678:
LGTM

liangchenye:
LGTM  
I think it is the best choice so far.

PR #531, opened 2017-12-02
Fix id mappings

Mashimiao:
1. the limit depends on different kernel version and not definitely required by spec

### DIFF
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -615,10 +615,6 @@ func (v *Validator) CheckLinux() (errs error) {
 
 	if (len(v.spec.Linux.UIDMappings) > 0 || len(v.spec.Linux.GIDMappings) > 0) && !nsTypeList[rspec.UserNamespace].newExist {
 		errs = multierror.Append(errs, errors.New("the UID/GID mappings requires a new User namespace to be specified as well"))
-	} else if len(v.spec.Linux.UIDMappings) > 5 {
-		errs = multierror.Append(errs, errors.New("only 5 UID mappings are allowed (linux kernel restriction)"))
-	} else if len(v.spec.Linux.GIDMappings) > 5 {
-		errs = multierror.Append(errs, errors.New("only 5 GID mappings are allowed (linux kernel restriction)"))
 	}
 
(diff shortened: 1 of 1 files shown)


q384566678:
LGTM

wking:
On Sat, Dec 02, 2017 at 05:48:16AM +0000, Ma Shimiao wrote:
> 1. the limit depends on different kernel version and not definitely
>    required by spec

+1 to this, so I'm in favor of your validate.go change.

> 2. uid and gid mappings must be set together

You're at least allowed to write uid_map without writing gid_map.  As
an unprivileged user running ccon [1]:

  $ cat config.json 
  {
    "version": "0.4.0",
    "namespaces": {
      "user": {
        "uidMappings": [
          {
            "containerID": 0,
            "hostID": 1000,
            "size": 1
          }
        ]
      }
    },
    "process": {
      "args": [
        "busybox",
        "id"
      ]
    }
  }
  $ ccon
  uid=0(root) gid=65534(nobody) groups=65534(nobody),65534(nobody),65534(nobody),65534(nobody),65534(nobody),65534(nobody),65534(nobody),65534(nobody),65534(nobody),65534(nobody),65534(nobody),65534(nobody),65534(nobody)

The above config maps my unprivileged user 1000 to 0 but left the
groups unmapped.  And ‘busybox id’ can't handle it, but you can also
map the group while leaving the user unmapped:

  $ cat config.json 
  {
    "version": "0.4.0",
    "namespaces": {
      "user": {
        "setgroups": false,
        "gidMappings": [
          {
            "containerID": 0,
            "hostID": 1000,
            "size": 1
          }
        ]
      }
    },
    "process": {
      "args": [
        "/bin/echo",
        "hello"
      ]
    }
  }
  $ ccon
  hello

so I'm not in favor of your validation changes.

This may also be related to your opencontainers/runc#1665 PR.

[1]: https://github.com/wking/ccon

Mashimiao:
I think unmapped users or groups may cause operation or security problem. But that needs to talk in runtime-spec.  Just remove mapping limit this time.

q384566678:
LGTM

liangchenye:
LGTM

PR #536, opened 2017-12-06
filepath: fix the Abs error under Windows

q384566678:
Fix the following error when `os` is `Windows`, can‘t judge `mount.Destination`.

```
./oci-runtime-tool validate --platform windows
2 errors occurred:

* Abs() does not support windows yet
* Abs() does not support windows yet
```

### DIFF
--- a/filepath/clean.go
+++ b/filepath/clean.go
@@ -46,7 +46,7 @@ func Clean(os, path string) string {
 	}
 
 	cleaned := strings.Join(elements, string(sep))
-	if abs {
+	if abs && os == "linux" {
 		cleaned = fmt.Sprintf("%c%s", sep, cleaned)
 	}
 	if cleaned == path {
--- a/filepath/abs.go
+++ b/filepath/abs.go
@@ -1,7 +1,6 @@
 package filepath
 
 import (
-	"errors"
 	"regexp"
 	"strings"
 )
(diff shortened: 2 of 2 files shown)


Mashimiao:
LGTM

filepath/clean.go
```diff
@@ -46,7 +46,7 @@ func Clean(os, path string) string {
 	}
 
 	cleaned := strings.Join(elements, string(sep))
-	if abs {
+	if abs && os == "linux" {
```
wking: Is this really all we need to do to support Windows?

Mashimiao:
At least, after modified, we can validate config file for windows platform on Linux.

wking:
On Thu, Dec 07, 2017 at 05:16:25PM -0800, Ma Shimiao wrote:
> At least, after modified, we can validate config file for windows
> platform on Linux.

Validating on Windows with “our rough guess at IsAncestor reports
false for the paths you gave; we don't test any of this” isn't very
satisfying ;).  I've just filed #539 which removes the guard (like
this PR), but also adds tests for Windows (not in this PR).

And while I'm just guessing at the expected Windows results, I've
added new code to compare `IsAbs` and `Clean` with their stdlib
analogs.  So anyone with Windows access can confirm the expected
results just by running:

  $ go test github.com/opencontainers/runtime-tools/filepath

on Windows.

liangchenye:
I installed a golang sdk on Windows and test @wking 's #539. It works good.

q384566678:
Closed because #539  has been merged, the problem has been solved.

PR #537, opened 2017-12-08
validate_test: add test func

q384566678:
There are some waiting improvements for the validation of the `mounts`, and the test function can be added later.

### DIFF
--- a/validate/validate_test.go
+++ b/validate/validate_test.go
@@ -414,3 +414,41 @@ func TestCheckLinux(t *testing.T) {
 		assert.Equal(t, c.expected, specerror.FindError(err, c.expected), fmt.Sprintf("failed CheckLinux: %v %d", err, c.expected))
 	}
 }
+
+func TestCheckPlatform(t *testing.T) {
+	cases := []struct {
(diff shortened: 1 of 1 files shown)


validate/validate_test.go
```diff
@@ -414,3 +414,80 @@ func TestCheckLinux(t *testing.T) {
...
+							Path: "usr",
+						},
+					},
+				},
+			},
+			expected: specerror.PosixHooksPathAbs,
+		},
+	}
+	for _, c := range cases {
+		v := NewValidator(&c.val, ".", false, "linux")
```
wking: Do we want a test to demonstrate that `Hooks` is ignored on Windows?

q384566678: We can add it later because the corresponding verification code is not perfect yet.

wking: > We can add it later because the corresponding verification code is not perfect yet.

Ah right, it is blocked on #538.

liangchenye:
LGTM

Mashimiao:
LGTM

PR #538, opened 2017-12-08
validate: modify the non-conforming validation

q384566678:
1.Capabilities only support linux-based systems, so verification functions should not be called on other systems.

2.Hooks only support  POSIX platforms.

3.add a double guards to the call of the verification function
* Internally, raise an error if you call a check function for a platform that doesn't define the checked property 
* Externally, only call check functions on the appropriate platforms.

### DIFF
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -264,6 +264,10 @@ func (v *Validator) CheckSemVer() (errs error) {
 func (v *Validator) CheckHooks() (errs error) {
 	logrus.Debugf("check hooks")
 
+	if v.platform == "windows" {
+		return
+	}
+
 	if v.spec.Hooks != nil {
 		errs = multierror.Append(errs, v.checkEventHooks("prestart", v.spec.Hooks.Prestart, v.HostSpecific))
 		errs = multierror.Append(errs, v.checkEventHooks("poststart", v.spec.Hooks.Poststart, v.HostSpecific))
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -356,12 +356,13 @@ func (v *Validator) CheckProcess() (errs error) {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


Mashimiao:
LGTM

validate/validate.go
```diff
@@ -377,59 +378,55 @@ func (v *Validator) CheckProcess() (errs error) {
 // CheckCapabilities checks v.spec.Process.Capabilities
 func (v *Validator) CheckCapabilities() (errs error) {
 	process := v.spec.Process
-	if v.platform == "linux" {
```
wking: I think we need to decide whether these platform guards belong inside the check function, outside it, or both.  For example, we currently [call `CheckRlimits` regardless of platform][1] and have [a guard inside it to no-op on platforms where it isn't defined][2].  The `if` check you're removing here is basically the same sort of no-op internal guard.  And you're *adding* a similar internal no-op guard to `CheckHooks` in this PR with 4b20ce20f.

The most robust approach would be to guard on both sides:

* Internally, raise an error if you call a check function for a platform that doesn't define the checked property (e.g. calling `CheckCapabilities` for Windows or Solaris).
* Externally, only call check functions on the appropriate platforms.

Thoughts?

[1]: https://github.com/opencontainers/runtime-tools/blob/acbe085c54698d0ccb98892efbc233f61d2faed5/validate/validate.go#L362
[2]: https://github.com/opencontainers/runtime-tools/blob/acbe085c54698d0ccb98892efbc233f61d2faed5/validate/validate.go#L440-L441

q384566678: Agree. I will update.

q384566678: updated, PTAL

Mashimiao:
LGTM

liangchenye:
LGTM

PR #539, opened 2017-12-08
filepath/clean: Add Windows support

wking:
Also fix the Linux `Clean` handling for `a/..` and `/..` and add logic to compare the `GOOS` `Clean` and `IsAbs` results with the stdlib.

This is an alternative to #536.


liangchenye:
LGTM

Mashimiao:
LGTM

PR #540, opened 2017-12-11
generate: Use non-null validation instead of initialization

q384566678:
Using non-null validation makes the code more efficient.

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -534,7 +534,10 @@ func (g *Generator) AddLinuxResourcesBlockIOLeafWeightDevice(major int64, minor 
 
 // DropLinuxResourcesBlockIOLeafWeightDevice drops a item form g.spec.Linux.Resources.BlockIO.WeightDevice.LeafWeight
 func (g *Generator) DropLinuxResourcesBlockIOLeafWeightDevice(major int64, minor int64) {
-	g.initSpecLinuxResourcesBlockIO()
+	if g.spec == nil || g.spec.Linux == nil || g.spec.Linux.Resources == nil || g.spec.Linux.Resources.BlockIO == nil {
+		return
+	}
+
 	for i, weightDevice := range g.spec.Linux.Resources.BlockIO.WeightDevice {
 		if weightDevice.Major == major && weightDevice.Minor == minor {
 			if weightDevice.Weight != nil {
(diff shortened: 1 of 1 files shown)


generate/generate.go
```diff
@@ -709,7 +727,10 @@ func (g *Generator) AddLinuxResourcesHugepageLimit(pageSize string, limit uint64
 
 // DropLinuxResourcesHugepageLimit drops a hugepage limit from g.spec.Linux.Resources.HugepageLimits.
 func (g *Generator) DropLinuxResourcesHugepageLimit(pageSize string) error {
-	g.initSpecLinuxResources()
+	if g.spec == nil || g.spec.Linux == nil || g.spec.Linux.Resources == nil {
```
Mashimiao: Almost every drop functions have this, I think you can add functions like `initSpecLinuxResources()` to form them.

wking: > I think you can add functions like `initSpecLinuxResources()` to form them.

You don't need to create ancestor properties just to unset some descendant property.  If any ancestor is unset, you know the descendant is already unset.  For example, currently:

```console
$ oci-runtime-tool generate --temlate <(echo '{}') | jq .linux
null
$ oci-runtime-tool generate --template <(echo '{}') --linux-hugepage-limits-drop 2MB | jq .linux
{
  "resources": {}
}
```

With this commit we'd no longer create the useless `linux.resources` entry.

However, even without this commit, we are currently not clearing useless entries after we empty an array.  Calling `g.initSpecLinuxResources()` and seeing that the entry in question is already empty, and then removing the the useless entries is a bit of busy work, but wouldn't be the end of the world.  And if you decide to not bother removing the useless entries, that wouldn't be the end of the world either.

Mashimiao: I think you misunderstood me. I don't mean to keep `initSpecLinuxResources()`, I agree to remove it in Drop*() functions.
I mean to add a new functions to judge nil as init*() functions.
As we have almost same `if g.spec == nil || g.spec.* == nil` sentence in many place, I think we'd better add new functions to form them. Looks more elegant.

wking: > I mean to add a new functions to judge nil...

Ah.  That would work, although I don't have a problem with the inline `||` chain either.

q384566678: >I mean to add a new functions to judge nil as init*() functions.

Fine, I'll update.

q384566678: updated, PTAL

Mashimiao:
LGTM

liangchenye:
LGTM

PR #543, opened 2017-12-15
Recursive propagation flags should be legal to use

rhatdan:
Current root propagation code does not allow rslave, rshared, runbindable and rprivate.
These are all legal proagations and should be allowed.

Also added the "bad" propagation flag to the error to make diagnosis easier.

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -891,11 +891,15 @@ func (g *Generator) SetLinuxRootPropagation(rp string) error {
 	switch rp {
 	case "":
 	case "private":
+	case "rprivate":
 	case "slave":
+	case "rslave":
 	case "shared":
+	case "rshared":
 	case "unbindable":
+	case "runbindable":
 	default:
-		return fmt.Errorf("rootfs-propagation must be empty or one of private|slave|shared|unbindable")
+		return fmt.Errorf("rootfs-propagation %q must be empty or one of (r)private|(r)slave|(r)shared|(r)unbindable", rp)
 	}
 	g.initSpecLinux()
 	g.spec.Linux.RootfsPropagation = rp


wking:
On Fri, Dec 15, 2017 at 12:11:27PM +0000, Daniel J Walsh wrote:
> Current root propagation code does not allow rslave, rshared, runbindable and rprivate.
> These are all legal proagations and should be allowed.

They are not listed in the normative runtime-spec Markdown [1], and
are forbidden by the informative runtime-spec JSON Schema [2].
However, without RFC 2119 wording in [1] I would consider them
technically allowed.  I have opencontainers/runtime-spec#923 open
about clarifying the spec.

I'm agnostic for this PR.  On the one hand, being able to write these
not-RFC-2119-forbidden values is nice (although with that approach, we
should let SetLinuxRootPropagation write *any* string value).  These
configs are going to fail verification, where we rely on the spec's
JSON Schema since #197, but maybe folks won't run our verification on
them ;).

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/config-linux.md#rootfs-mount-propagation
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/schema/defs-linux.json#L3-L11

rhatdan:
@mrunalp @crosbymichael PTAL

rhatdan:
@rhvgoyal PTAL

mrunalp:
LGTM

liangchenye:
LGTM

We also need another PR to add this change to:
https://github.com/opencontainers/runtime-tools/blob/master/completions/bash/oci-runtime-tool#L130
https://github.com/opencontainers/runtime-tools/blob/master/man/oci-runtime-tool-generate.1.md

PR #279, opened 2016-11-24
generate:  support adding additional mounts for container

Mashimiao:
Add support for adding additional mounts.
Users can add custom additional mounts, like adding tmpfs without special options or with custom options.

PR #544, opened 2017-12-15
Add interface to remove mounts.

rhatdan:
We want to add a mount point to /dev/shm using the default spec,
but we don't want to mount /dev/shm via a tmpfs and then add our mountpoint
over it.

RemoveMount will remove a mount point based on the Destination, if it exists.
Does not return any error if the mount point does not exist.  Also added
Mounts() interface so that I could populate a test for new feature. Returns
the list of mounts.

### DIFF
--- a/cmd/oci-runtime-tool/generate.go
+++ b/cmd/oci-runtime-tool/generate.go
@@ -88,6 +88,7 @@ var generateFlags = []cli.Flag{
 	cli.StringSliceFlag{Name: "linux-sysctl", Usage: "add sysctl settings e.g net.ipv4.forward=1"},
 	cli.StringSliceFlag{Name: "linux-uidmappings", Usage: "add UIDMappings e.g HostID:ContainerID:Size"},
 	cli.StringSliceFlag{Name: "mounts-add", Usage: "configures additional mounts inside container"},
+	cli.StringSliceFlag{Name: "mounts-remove", Usage: "remove destination mountpoints from inside container"},
 	cli.BoolFlag{Name: "mounts-remove-all", Usage: "remove all mounts inside container"},
 	cli.StringFlag{Name: "output", Usage: "output file (defaults to stdout)"},
 	cli.BoolFlag{Name: "privileged", Usage: "enable privileged container settings"},
@@ -430,6 +431,13 @@ func setupSpec(g *generate.Generator, context *cli.Context) error {
 
 	if context.IsSet("mounts-remove-all") {
 		g.ClearMounts()
+	}
+
+	if context.IsSet("mounts-remove") {
+		mounts := context.StringSlice("mounts-remove")
+		for _, mount := range mounts {
+			g.RemoveMount(mount)
+		}
 	}
 
 	if context.IsSet("mounts-add") {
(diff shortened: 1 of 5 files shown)


rhatdan:
@mrunalp @philips @crosbymichael PTAL

validation/mounts.go
```diff
@@ -1,9 +1,16 @@
 package main
 
 import (
+	"fmt"
+
 	"github.com/opencontainers/runtime-tools/validation/util"
 )
 
 func main() {
-	util.Skip("TODO: mounts generation options have not been implemented", "")
```
wking: This test is for runtime validation (and the TODO is stale now that #279 has landed, but that's another matter).

Tests for *generation* functionality (which is what you're adding in this PR) should go in [`generate/generate_test.go`][1].

[1]: https://github.com/opencontainers/runtime-tools/blob/c3755c1175a1baad7230dfb03751e3a4b7182f32/generate/generate_test.go

rhatdan: I am not sure why you would not want to have this in the validation code?  This seems like the correct place unless you are dropping the validation code also?

wking: > I am not sure why you would not want to have this in the validation code?

There are several types of validation in this repo:

1. Config validation in `validate`, with unit tests in `validate/validate_test.go`.
2. Runtime validation (which calls the runtime executable, e.g. [here][1]) in `validation`.  There are no unit tests for this (yet?).
3. Config generation argument checks in `generate`, with unit tests (of which there are currently few) in `generate/generate_test.go`.

The tests you're currently adding here (a type-2 location) belong to type-3 (generation unit tests).

[1]: https://github.com/opencontainers/runtime-tools/blob/c3755c1175a1baad7230dfb03751e3a4b7182f32/validation/create.go

rhatdan: Ok copied it to the generate_test.go.  But not sure if I should remove it totally from Validate.  Currently tests in both places.

wking: > Ok copied it to the generate_test.go.

That looks good to me.

> But not sure if I should remove it totally from Validate.

I think you should, because the test has nothing to do with runtime compliance (which is what `validation/` is for).  It's just testing the generate implementation, so the `generate_test.go` test you've added is all you need.

cmd/oci-runtime-tool/generate.go
```diff
@@ -87,6 +87,7 @@ var generateFlags = []cli.Flag{
 	cli.StringSliceFlag{Name: "linux-sysctl", Usage: "add sysctl settings e.g net.ipv4.forward=1"},
 	cli.StringSliceFlag{Name: "linux-uidmappings", Usage: "add UIDMappings e.g HostID:ContainerID:Size"},
 	cli.StringSliceFlag{Name: "mounts-add", Usage: "configures additional mounts inside container"},
+	cli.StringSliceFlag{Name: "mounts-remove", Usage: "remove destionation mountpoints from inside container"},
```
mrunalp: nit: destionation typo

q384566678:
The appropriate content should be added to `completions/bash`.

rhatdan:
@q384566678 Added to completions/bash

cmd/oci-runtime-tool/generate.go
rhatdan: Fixed

rhatdan:
@mrunalp PTAL

mrunalp:
One of the commits isn't signed. Otherwise looks good.

rhatdan:
@mrunalp Fixed PR.

mrunalp:
LGTM

q384566678:
LGTM

PR #545, opened 2017-12-16
AddMounts should be AddMount you are only adding a single Mount

rhatdan:
AddMount should also take a rspec.Mount object rather then a json string.
We are using generate as a library not using the command line tool.
The tool should translate the json into a rspec.Mount.

### DIFF
--- a/cmd/oci-runtime-tool/generate.go
+++ b/cmd/oci-runtime-tool/generate.go
@@ -3,6 +3,7 @@ package main
 import (
 	"bufio"
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"os"
 	"strconv"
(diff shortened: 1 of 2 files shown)


wking:
> We are using generate as a library not using the command line tool.

I think this use-case would be more convenient with #266, although you can get similar access with the current [`Spec()`][1].  An API oriented around direct config access would shift the JSON umarshall under `cmd/` (like you do here) and drop `AddMounts` entirely (because it's easy to call append yourself).

[1]: https://github.com/opencontainers/runtime-tools/blob/v0.3.0/generate/generate.go#L258

mrunalp:
This is an improvement to the current API. LGTM.

mrunalp:
LGTM

q384566678:
LGTM

PR #553, opened 2018-01-12
cmd/runtimetest/main: Run validateDefaultDevices even with process unset

wking:
6d2dbbc9 (#494) put the nil-`Process` guard in, but it was a bit too coarse.  We can still run these tests with a nil `Process`, because all we need `Process` for is deciding whether or not to include `/dev/console`.


q384566678:
LGTM

wking:
Closing and reopening to kick Travis.

PR #554, opened 2018-01-18
validate_test: add TestCheckMandatoryFields

q384566678:
Because there are so many cases, I only chose part of the test.


q384566678:
@liangchenye PTAL

q384566678:
@liangchenye PTAL

liangchenye:
Looking into this.
I'm checking if the unit test cover all cases of this function.

liangchenye:
It looks good. I just found the original function does not handle one case -- the 'v.spec == nil'.
Can you fix it and add the related unit test?

q384566678:
updated, thanks. PTAL.

liangchenye:
LGTM

PR #551, opened 2018-01-11
validation/util/container: Use --bundle (and stop requiring BundleDir)

wking:
Address a previous TODO.  And now that we are using `--bundle`, we no longer need to set `cmd.Dir`.  The TODO mentions a lack of runc support, but runc supports `--bundle` since opencontainers/runc@3fe7d7f3 (opencontainers/runc#827).

PR #557, opened 2018-01-22
Fix #556 

### LINKED ISSUE #556: Complete the CLI args
Now we have [CLI spec](https://github.com/opencontainers/runtime-tools/blob/master/docs/command-line-interface.md#commands) defined, we need to complete the 'args' of 'create/kill' in 'validation/util/container' .
- [ ] add lifecycle testing function
   working on #557
- [ ] create with 'pid-file'
    working on #557
- [ ] create with 'socket'
    I don't have an idea to implement it now.
- [ ] kill with 'signal'
    'runc' does not follow this yet or we can implement it by writing a process which track the kill signal

liangchenye:
1. should rebase after #551  -- done
2. should add more test case to cover 'create'  
3. should add test case to cover 'kill' -- will do that in other PR

### DIFF
--- a/validation/pidfile.go
+++ b/validation/pidfile.go
@@ -1,0 +1,63 @@
+package main
+
+import (
+	"errors"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strconv"
+	"time"
+
+	tap "github.com/mndrix/tap-go"
+	"github.com/opencontainers/runtime-tools/validation/util"
+)
+
+func main() {
+	t := tap.New()
+	t.Header(0)
+
+	tempDir, err := ioutil.TempDir("", "oci-pid")
+	if err != nil {
+		util.Fatal(err)
+	}
+	defer os.RemoveAll(tempDir)
+	tempPidFile := filepath.Join(tempDir, "pidfile")
+
+	config := util.LifecycleConfig{
+		Actions: util.LifecycleActionCreate | util.LifecycleActionDelete,
+		PreCreate: func(r *util.Runtime) error {
+			r.PidFile = tempPidFile
+			return nil
+		},
+		PostCreate: func(r *util.Runtime) error {
+			pidData, err := ioutil.ReadFile(tempPidFile)
+			if err != nil {
+				return err
+			}
+			pid, err := strconv.Atoi(string(pidData))
+			if err != nil {
+				return err
+			}
+			state, err := r.State()
+			if err != nil {
+				return err
+			}
+			if state.Pid != pid {
+				return errors.New("Wrong pid in the pidfile")
+			}
+			return nil
+		},
+		PreDelete: func(r *util.Runtime) error {
+			return util.WaitingForStatus(*r, util.LifecycleStatusCreated, time.Second*10, time.Second*1)
+		},
+	}
+
+	g := util.GetDefaultGenerator()
+	g.SetProcessArgs([]string{"true"})
+	err = util.RuntimeLifecycleValidate(g, config)
+	if err != nil {
+		util.Fatal(err)
+	}
+
+	t.AutoPlan()
+}
--- a/validation/util/container.go
+++ b/validation/util/container.go
@@ -18,6 +18,7 @@ import (
 type Runtime struct {
 	RuntimeCommand string
 	BundleDir      string
+	PidFile        string
 	ID             string
 	stdout         *os.File
 	stderr         *os.File
@@ -63,7 +64,9 @@ func (r *Runtime) Create() (stderr []byte, err error) {
 	if r.ID != "" {
 		args = append(args, r.ID)
 	}
-
+	if r.PidFile != "" {
+		args = append(args, "--pid-file", r.PidFile)
+	}
 	if r.BundleDir != "" {
 		args = append(args, "--bundle", r.BundleDir)
 	}
@@ -146,15 +149,24 @@ func (r *Runtime) State() (rspecs.State, error) {
 }
 
 // Delete a container
-func (r *Runtime) Delete() error {
+func (r *Runtime) Delete() ([]byte, error) {
 	var args []string
+	var stderr []byte
 	args = append(args, "delete")
 	if r.ID != "" {
 		args = append(args, r.ID)
 	}
 
 	cmd := exec.Command(r.RuntimeCommand, args...)
-	return cmd.Run()
+	stdout, err := cmd.Output()
+	if e, ok := err.(*exec.ExitError); ok {
+		stderr = e.Stderr
+	}
+	if err != nil && len(stderr) == 0 {
+		stderr = stdout
+	}
+
+	return stderr, err
 }
 
 // Clean deletes the container.  If removeBundle is set, the bundle
@@ -162,7 +174,7 @@ func (r *Runtime) Delete() error {
 // forceRemoveBundle is true, after the deletion attempt regardless of
 // whether it was successful or not.
 func (r *Runtime) Clean(removeBundle bool, forceRemoveBundle bool) error {
-	err := r.Delete()
+	_, err := r.Delete()
 
 	if removeBundle && (err == nil || forceRemoveBundle) {
 		err2 := os.RemoveAll(r.bundleDir())
--- a/validation/util/test.go
+++ b/validation/util/test.go
@@ -1,6 +1,7 @@
 package util
 
 import (
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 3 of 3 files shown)


liangchenye:
This PR does two things:
1. add LifecycleFuncs 
```
type LifecycleFuncs struct {
 	Setup     func(runtime *Runtime) error
 	PreStart  func(runtime *Runtime) error
 	PostStart func(runtime *Runtime) error
 	PostStop  func(runtime *Runtime) error
 }
 ```
It is similar to ```RuntimeOutsideValidate(g *generate.Generator, f AfterFunc) ```. 
But I think we don't need to define PreFunc or AfterFunc, we can use ```func(runtime *Runtime)``` and get any information from the Runtime structure.

So if this approach is useful, we can merge 'RuntimeInsideValidate'， ‘RuntimeOutsideValidate' and 'RuntimeLifecycleValidate' into one.

2. add PIDFile test
I take this as an example of using 'RuntimeLifecycleValidate'.

File: validation/pidfile.go [hunk 1]
```diff
@@ -0,0 +1,66 @@
...
+	tap "github.com/mndrix/tap-go"
+	"github.com/opencontainers/runtime-tools/validation/util"
+)
+
+func main() {
+	t := tap.New()
+	t.Header(0)
+
+	var lifecycle util.LifecycleFuncs
+	tempFile, err := ioutil.TempFile("", "oci-pidfile")
```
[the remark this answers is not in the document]
liangchenye: It is not an existing PID file, it is also a new temp file, similar to using TempDir + 'pid'.

validation/pidfile.go
```diff
@@ -0,0 +1,66 @@
...
+			return err
+		}
+		if state.Pid != pid {
+			return errors.New("Wrong pid in the pidfile")
+		}
+		return nil
+	}
+
+	lifecycle.PostStart = func(r *util.Runtime) error {
+		time.Sleep(1 * time.Second)
```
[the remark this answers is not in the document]
liangchenye: Waiting for the container to stop, or we cannot delete it.
There should be a better function to waiting for the status change.

validation/pidfile.go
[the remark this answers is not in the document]
wking: > It is not an existing PID file, it is also a new temp file, similar to using TempDir + 'pid'.

[The `TempFile` docs][1] say it creates the file.  `TempDir` `+ "pid"` would not.

[1]: https://golang.org/pkg/io/ioutil/#TempFile

validation/pidfile.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	tempPidFile := tempFile.Name()
+	lifecycle.Setup = func(r *util.Runtime) error {
+		r.SetPidFile(tempPidFile)
+		return nil
+	}
+	defer os.Remove(tempPidFile)
+
+	lifecycle.PreStart = func(r *util.Runtime) error {
+		pidData, err := ioutil.ReadFile(tempPidFile)
```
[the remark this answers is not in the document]
liangchenye: +1, 'PostCreat'e is better, 'Prestart' might also be confusing,

validation/util/container.go
```diff
@@ -146,23 +154,32 @@ func (r *Runtime) State() (rspecs.State, error) {
 }
 
 // Delete a container
-func (r *Runtime) Delete() error {
+func (r *Runtime) Delete() ([]byte, error) {
```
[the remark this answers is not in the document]
liangchenye: I use this to debug why 'Delete' fails.
The 'time.Sleep' line is coming from this.

validation/util/test.go
```diff
@@ -184,3 +192,67 @@ func RuntimeOutsideValidate(g *generate.Generator, f AfterFunc) error {
...
+		return err
+	}
+
+	if lifecycle != nil && lifecycle.PreStart != nil {
+		if err := lifecycle.PreStart(&r); err != nil {
+			return err
+		}
+	}
+
+	stderr, err = r.Start()
```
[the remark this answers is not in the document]
liangchenye: So something like this
```
func RuntimeLifecycleValidate(g *generate.Generator, lifecycle *LifecycleFuncs,  actions LifecycleAction) {
...
if actions & LifeCycleCreate {
     stderr, err := r.Create()
}
...
if actions & LifeCycleStart {
     stderr, err = r.Start()
}
...
}
```

validation/pidfile.go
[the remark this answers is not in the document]
liangchenye: Got it.

validation/pidfile.go
```diff
@@ -0,0 +1,66 @@
...
+
+	lifecycle.PostStart = func(r *util.Runtime) error {
+		time.Sleep(1 * time.Second)
+		return nil
+	}
+
+	lifecycle.PostStop = nil
+
+	g := util.GetDefaultGenerator()
+	g.SetProcessArgs([]string{"ls"})
```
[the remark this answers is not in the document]
liangchenye: 'runc' has a bug: if the 'Process' is nil, runc crashes.

Besides this, 'runc' does not follow the CLI spec, it applies the 'process args' when create a container.
I add a test to change the 'process args' in PostCreate, in theory, runc should execute the new args, but it does not.

liangchenye:
1. add LifecycleFuncs 'Setup/PostCreate/PreStop/PostStop'
    'PostCreate' and 'PreStop' are slight different with 'PreStart' and 'PostStart', so these new names are used. 
2. add WaitingForStatus
    It is flexible to add this to the testing code.
3. add pidfile.go
    It is used to test if '--pid-file' works as expected
    I added a `WaitingForStatus(Created)` function in 'PreStop', just in case the 'creating' takes a long time.

Two issues were found, but I think they could be solved in other PRs.
1. ProcessArgs issue of 'runc'
    'runc' validates and applies ProcessArgs in the 'create' phase which is not the spec expected.
    Now I use 'true' command in pidfile test.
2. Delete a container with the 'Created' status
    I think we can delete a container with 'created' status, we should update the [spec](https://github.com/opencontainers/runtime-spec/blob/master/runtime.md#delete).   

PTAL @wking @q384566678 @Mashimiao

validation/pidfile.go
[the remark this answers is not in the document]
wking: > 'runc' has a bug: if the 'Process' is nil, runc crashes

`create` crashes?  That would be a bug.

> I add a test to change the 'process args' in PostCreate, in theory, runc should execute the new args, but it does not.

No, the runtime [must ignore any post-create config changes][1], so `runc` is fine on that point.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/runtime.md#L57

q384566678:
LGTM

validation/pidfile.go
```diff
@@ -0,0 +1,62 @@
...
+func main() {
+	t := tap.New()
+	t.Header(0)
+
+	tempDir, err := ioutil.TempDir("", "oci-pid")
+	if err != nil {
+		util.Fatal(err)
+	}
+	tempPidFile := filepath.Join(tempDir, "pidfile")
+	defer os.RemoveAll(tempDir)
```
wking: nit: I'd rather swap the order of these two lines to get the cleanup `defer` right after the `TempDir` error check.  `filepath.Join` cannot error, but having the `defer` set as early as possible would give me warm fuzzy feelings ;).

validation/util/test.go
```diff
@@ -21,6 +22,47 @@ var (
...
+	// LifecycleStatusCreated "created"
+	LifecycleStatusCreated
+	// LifecycleStatusRunning "running"
+	LifecycleStatusRunning
+	// LifecycleStatusStopped "stopped"
+	LifecycleStatusStopped
+)
+
+var lifecycleStatusMap = map[string]LifecycleStatus{
+	"creating": LifecycleStatusCreating,
```
wking: Or these could just be strings:

```go
type LivecycleStatus string

const (
  LifecycleStatusCreating = "creating"
  …
)
```

and then you cast with `LifecycleStatus(state.Status)`.  I don't see a need for indirection through an integer.

validation/util/test.go
```diff
@@ -184,3 +248,74 @@ func RuntimeOutsideValidate(g *generate.Generator, f AfterFunc) error {
 	}
 	return nil
 }
+
+// RuntimeLifecycleValidate validates runtime lifecycle.
+func RuntimeLifecycleValidate(g *generate.Generator, lifecycle *LifecycleFuncs, actions LifecycleAction) error {
```
wking: I don't see a need to split funcs from actions.  Can we rename `LifecycleFuncs` to `LifecycleConfig` (or just `Lifecycle`?) and add an `Actions` property?

validation/util/test.go
```diff
@@ -184,3 +248,74 @@ func RuntimeOutsideValidate(g *generate.Generator, f AfterFunc) error {
...
+	if actions&LifecycleActionDelete != 0 {
+		stderr, err := r.Delete()
+		if err != nil {
+			os.Stderr.WriteString("failed to delete the container\n")
+			os.Stderr.Write(stderr)
+			return err
+		}
+	}
+
+	if lifecycle != nil && lifecycle.PostStop != nil {
```
wking: I think anything running after `Delete` should be `PostDelete`.  This PR currently seems to handle `stop` and `delete` in one step (or maybe it is never intended to call `stop` and instead waits for the container to stop on its own?), but calling this setting `PostDelete` makes sense regardless of an explicit stop or not.

validation/pidfile.go
liangchenye: OK, I'll update it.

validation/util/test.go
liangchenye: An 'integer' is useful when we expect to have multiple statuses. For example, if we want to delete a container, we expected to have  'created' or 'stopped' status. So we can 'WaitingForStatus(Created|Stopped)'.

validation/util/test.go
liangchenye: +1 for this

validation/util/test.go
liangchenye: So all the embedded functions should have different new, I'll update it too.

I prefer calling the 'stop' in the related test cases, for example, if we want to test 'kill'.
We can add test codes in the 'PreDelete'.   (PreDelete is better than PreStop)

liangchenye:
Updated by using lifecycleConfig (merge funcs and actions), now the funcs are symmetrical: PreCreate/PostCreate, PreDelete/PostDelete

Mashimiao:
LGTM

q384566678:
LGTM

validation/pidfile.go
```diff
@@ -0,0 +1,63 @@
...
+			if err != nil {
+				return err
+			}
+			if state.Pid != pid {
+				return errors.New("Wrong pid in the pidfile")
+			}
+			return nil
+		},
+		PreDelete: func(r *util.Runtime) error {
+			return util.WaitingForStatus(*r, util.LifecycleStatusCreated, time.Second*10, time.Second*1)
```
wking: We shouldn't need this, because a container [must be created before `create` returns][1].  Maybe a runc bug?

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/runtime.md#L20

liangchenye: `runc` does not implement ['creating' lifecycle ](https://github.com/opencontainers/runc/blob/master/libcontainer/container.go#L17), 
it is always 'created' or 'fails'.
From the spec, it should be like this:
1. user run `runX create x` 
2. user run `runX status x`
    it should be 'creating' (assuming the creating takes a long time).

wking: > `runc` does not implement 'creating' lifecycle...

That's definitely a bug.  And there's also some [CLI spec wording][1] about `create`'s exit timing.  I think we can drop this, and I'll file a runc patch.

[1]: https://github.com/opencontainers/runtime-tools/blame/7ac117af23fb7465142f16ac0d22277c9c82fefa/docs/command-line-interface.md#L79

PR #558, opened 2018-01-22
add lifecycle validation

q384566678:
I am going to add a validation of the lifecycle, the specific norms are as follows:

>The pre-start hooks MUST be called after the start operation is called but before the user-specified program command is executed.

But I can not confirm the specific call state, so for the moment I'm just verifying that the relevant command has not been called.

This is just a preliminary idea. Lifecycle validation also need more work, any comments welcome. Or you can write a new patch.

### DIFF
--- a/validation/poststart.go
+++ b/validation/poststart.go
@@ -1,0 +1,92 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"time"
+
+	tap "github.com/mndrix/tap-go"
+	rspec "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/opencontainers/runtime-tools/specerror"
+	"github.com/opencontainers/runtime-tools/validation/util"
+	uuid "github.com/satori/go.uuid"
+)
+
+func main() {
+	t := tap.New()
+	t.Header(0)
+
+	g := util.GetDefaultGenerator()
+
+	var output string
+	config := util.LifecycleConfig{
+		Actions: util.LifecycleActionCreate | util.LifecycleActionStart | util.LifecycleActionDelete,
+		PreCreate: func(r *util.Runtime) error {
+			r.SetID(uuid.NewV4().String())
+
+			output = filepath.Join(r.BundleDir, g.Spec().Root.Path, "output")
+			poststart := rspec.Hook{
+				Path: fmt.Sprintf("%s/%s/bin/sh", r.BundleDir, g.Spec().Root.Path),
+				Args: []string{
+					"sh", "-c", fmt.Sprintf("echo 'post-start called' >> %s", output),
+				},
+			}
+
+			g.AddPostStartHook(poststart)
+			g.SetProcessArgs([]string{"sh", "-c", fmt.Sprintf("echo 'process called' >> %s", "/output")})
+			r.SetConfig(g)
+			return nil
+		},
+		PostCreate: func(r *util.Runtime) error {
+			outputData, err := ioutil.ReadFile(output)
+			if err == nil {
+				if strings.Contains(string(outputData), "post-start called") {
+					return specerror.NewError(specerror.PoststartTiming, fmt.Errorf("The post-start hooks MUST be called before the `start` operation returns"), rspec.Version)
+				} else if strings.Contains(string(outputData), "process called") {
+					return specerror.NewError(specerror.ProcNotRunAtResRequest, fmt.Errorf("The user-specified program (from process) MUST NOT be run at this time"), rspec.Version)
+				}
+				return fmt.Errorf("File %v should not exist", output)
+			}
+			return nil
+		},
+		PreDelete: func(r *util.Runtime) error {
+			util.WaitingForStatus(*r, util.LifecycleStatusStopped, time.Second*10, time.Second)
+			outputData, err := ioutil.ReadFile(output)
+			if err != nil {
+				return fmt.Errorf("%v\n%v", specerror.NewError(specerror.PoststartHooksInvoke, fmt.Errorf("The poststart hooks MUST be invoked by the runtime"), rspec.Version), specerror.NewError(specerror.ProcImplement, fmt.Errorf("The runtime MUST run the user-specified program, as specified by `process`"), rspec.Version))
+			}
+			switch string(outputData) {
+			case "post-start called\n":
+				return specerror.NewError(specerror.ProcImplement, fmt.Errorf("The runtime MUST run the user-specified program, as specified by `process`"), rspec.Version)
+			case "process called\n":
+				fmt.Fprintln(os.Stderr, "WARNING: The poststart hook invoke fails")
+				return nil
+			case "post-start called\nprocess called\n":
+				return specerror.NewError(specerror.PoststartTiming, fmt.Errorf("The post-start hooks MUST be called after the user-specified process is executed"), rspec.Version)
+			case "process called\npost-start called\n":
+				return nil
+			default:
+				return fmt.Errorf("Unsupported output information: %v", string(outputData))
+			}
+		},
+	}
+
+	err := util.RuntimeLifecycleValidate(g, config)
+	if err != nil {
+		diagnostic := map[string]string{
+			"error": err.Error(),
+		}
+		if e, ok := err.(*exec.ExitError); ok {
+			if len(e.Stderr) > 0 {
+				diagnostic["stderr"] = string(e.Stderr)
+			}
+		}
+		t.YAML(diagnostic)
+	}
+
+	t.AutoPlan()
+}
--- a/validation/poststop.go
+++ b/validation/poststop.go
@@ -1,0 +1,109 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"time"
+
+	tap "github.com/mndrix/tap-go"
+	rspec "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/opencontainers/runtime-tools/specerror"
+	"github.com/opencontainers/runtime-tools/validation/util"
+	uuid "github.com/satori/go.uuid"
+)
+
+func main() {
+	t := tap.New()
+	t.Header(0)
+
+	g := util.GetDefaultGenerator()
+
+	var output string
+	config := util.LifecycleConfig{
+		Actions: util.LifecycleActionCreate | util.LifecycleActionStart | util.LifecycleActionDelete,
+		PreCreate: func(r *util.Runtime) error {
+			r.SetID(uuid.NewV4().String())
+
+			output = filepath.Join(r.BundleDir, g.Spec().Root.Path, "output")
+			poststop := rspec.Hook{
+				Path: fmt.Sprintf("%s/%s/bin/sh", r.BundleDir, g.Spec().Root.Path),
+				Args: []string{
+					"sh", "-c", fmt.Sprintf("echo 'post-stop called' >> %s", output),
+				},
+			}
+
+			g.AddPostStopHook(poststop)
+			g.SetProcessArgs([]string{"sh", "-c", fmt.Sprintf("echo 'process called' >> %s", "/output")})
+			r.SetConfig(g)
+			return nil
+		},
+		PostCreate: func(r *util.Runtime) error {
+			outputData, err := ioutil.ReadFile(output)
+			if err == nil {
+				if strings.Contains(string(outputData), "post-stop called") {
+					return specerror.NewError(specerror.PoststopTiming, fmt.Errorf("The post-stop hooks MUST be called after the container is deleted"), rspec.Version)
+				} else if strings.Contains(string(outputData), "process called") {
+					return specerror.NewError(specerror.ProcNotRunAtResRequest, fmt.Errorf("The user-specified program (from process) MUST NOT be run at this time"), rspec.Version)
+				}
+				return fmt.Errorf("File %v should not exist", output)
+			}
+			return nil
+		},
+		PreDelete: func(r *util.Runtime) error {
+			util.WaitingForStatus(*r, util.LifecycleStatusStopped, time.Second*10, time.Second)
+			outputData, err := ioutil.ReadFile(output)
+			if err != nil {
+				return specerror.NewError(specerror.ProcImplement, fmt.Errorf("The runtime MUST run the user-specified program, as specified by `process`"), rspec.Version)
+			}
+			switch string(outputData) {
+			case "post-stop called\n":
+				return fmt.Errorf("%v\n%v", specerror.NewError(specerror.PoststopTiming, fmt.Errorf("The post-stop hooks MUST be called after the container is deleted"), rspec.Version), specerror.NewError(specerror.ProcImplement, fmt.Errorf("The runtime MUST run the user-specified program, as specified by `process`"), rspec.Version))
+			case "process called\n":
+				return nil
+			case "post-stop called\nprocess called\n", "process called\npost-stop called\n":
+				return specerror.NewError(specerror.PoststopTiming, fmt.Errorf("The post-stop hooks MUST be called after the container is deleted"), rspec.Version)
+			default:
+				return fmt.Errorf("Unsupported output information: %v", string(outputData))
+			}
+		},
+		PostDelete: func(r *util.Runtime) error {
+			outputData, err := ioutil.ReadFile(output)
+			if err != nil {
+				return fmt.Errorf("%v\n%v", specerror.NewError(specerror.PoststopHooksInvoke, fmt.Errorf("The poststop hooks MUST be invoked by the runtime"), rspec.Version), specerror.NewError(specerror.ProcImplement, fmt.Errorf("The runtime MUST run the user-specified program, as specified by `process`"), rspec.Version))
+			}
+			switch string(outputData) {
+			case "post-stop called\n":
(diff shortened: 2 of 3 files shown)


liangchenye:
>> The pre-start hooks MUST be called after the start operation is called but before the user-specified program command is executed.

Before 'start', we have 'create' operation, so I think this spec means:
1. when 'create', pre-start hooks SHOULD not be called
2. only when 'start', pre-start hooks will be called and be called after process

How about make a testcase like this:
1. in 'pre-start hook' we can set a command ```echo "pre-start called" >> $bundle/$rootfs/output```
2. in 'process' we can set a command ```echo "process called" >> $bundle/$rootfs/output```
   
so when 'create', the $bundle/$rootfs/output should be empty,
when 'start, the $bundle/$rootfs/output should look like this:
         ```pre-start called
            process called 
         ```

liangchenye:
I think `RuntimeOutsideValidate` is not suitable for lifecycle test.
We need to add 'preStart', 'postStart', 'postStop' functions as the arguments.
I try to do it in https://github.com/opencontainers/runtime-tools/pull/557.

q384566678:
I've already seen #557, and the setting of `LifecycleFuncs` is good. I will wait for #557 to merge and then update this patch.

q384566678:
According to the description of [lifecycle](https://github.com/opencontainers/runtime-spec/blob/master/runtime.md#lifecycle), I have determined the order of calls for each part:

 create -> start -> prestart -> process -> poststart -> delete -> poststop

The code is updated according to this call order.
@liangchenye @wking @Mashimiao PTAL

validation/lifecycle.go
```diff
@@ -0,0 +1,92 @@
...
+	var output = "$bundle/$rootfs/output"
+	config := util.LifecycleConfig{
+		Actions: util.LifecycleActionCreate | util.LifecycleActionStart | util.LifecycleActionDelete,
+		PreCreate: func(r *util.Runtime) error {
+			outputData, err := ioutil.ReadFile(output)
+			if err != nil {
+				return err
+			}
+			if string(outputData) != "" {
+				return errors.New("Wrong call")
```
q384566678: I am not quite sure what to say about this type of error. Any suggestions?

validation/lifecycle.go
```diff
@@ -0,0 +1,92 @@
...
+	rspec "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/opencontainers/runtime-tools/validation/util"
+)
+
+func main() {
+	t := tap.New()
+	t.Header(0)
+
+	prestart := rspec.Hook{
+		Path: "/usr/bin/prestart",
```
initlove: Where does `/usr/bin/prestart` come from?

validation/lifecycle.go
wking: > I am not quite sure what to say about this type of error.

The associated `specerror` entries start around [here][1].

[1]: https://github.com/opencontainers/runtime-tools/blob/eaecd9f609c4d68ec22e4737f6ecb0063cffc4b2/specerror/config.go#L69

validation/lifecycle.go
q384566678: About this `path`, I thought it was created automatically when added.
Maybe this idea is wrong.
Do you have any suggestions for the `path`?

wking: > Do you have any suggestions for the `path`?

The echo on our BusyBox tarball?

validation/lifecycle.go
q384566678: Depending on what is read, the specific error is different. It would be very troublesome to be accurate to every error condition. So instead of making a further distinction, I want to output the error directly.

Do you think it is necessary to distinguish between each error?

liangchenye: Yes, each error should explicitly reflect the relevant 'specerror'.

validation/lifecycle.go
liangchenye: I'm writing a demo to show my idea, I'll link to you soom.

liangchenye:
Following your idea, I made some changes: https://github.com/liangchenye/ocitools/blob/gotest/validation/lifecycle.go
1. using 'sh' and 'echo' in the bundle
2. compose hooks in 'preCreate'

About the 'Wrong Error", I suggest we may split the lifecycle.go to 'prestart.go', 'poststart.go' and the etc. Each hook has multiply spec errors, it will make it easier to maintain and cover all the errors. How do you think ? @q384566678

q384566678:
@liangchenye That sounds good. I'll update patch in this way. Thank you for your advice.

q384566678:
updated, @liangchenye @wking PTAL.

validation/poststart.go
```diff
@@ -0,0 +1,86 @@
...
+			}
+
+			g.AddPostStartHook(poststart)
+			g.SetProcessArgs([]string{"sh", "-c", fmt.Sprintf("echo 'process called' >> %s", output)})
+			r.SetConfig(g)
+			return nil
+		},
+		PostCreate: func(r *util.Runtime) error {
+			outputData, err := ioutil.ReadFile(output)
+			if err != nil {
```
liangchenye: It is expected to get an 'err', since the 'output' will not exist.

validation/poststart.go
```diff
@@ -0,0 +1,86 @@
...
+				return err
+			}
+			switch string(outputData) {
+			case "":
```
liangchenye: It should not happen, we can move it to 'default' part.

validation/poststart.go
```diff
@@ -0,0 +1,86 @@
...
+				return nil
+			case "post-start called\n":
```
liangchenye: We can make it simpler:
```
if strings.Contains(string(outputData), "post-start") {
     return specerror.NewError(specerror.PoststartTimin
} else if strings.Contains(string(outputData), "process") {
     return  specerror.NewError(specerror.ProcNotRunAtResRequest)
}
```

validation/poststart.go
```diff
@@ -0,0 +1,86 @@
...
+			case "post-start called\nprocess called\n", "process called\npost-start called\n":
+				return fmt.Errorf("%v\n%v", specerror.NewError(specerror.PoststartTiming, fmt.Errorf("The post-start hooks MUST be called before the `start` operation returns"), rspec.Version), specerror.NewError(specerror.ProcNotRunAtResRequest, fmt.Errorf("The user-specified program (from process) MUST NOT be run at this time"), rspec.Version))
+			default:
+				return fmt.Errorf("Unsupported output information: %v", string(outputData))
+			}
+		},
+		PreDelete: func(r *util.Runtime) error {
+			outputData, err := ioutil.ReadFile(output)
+			if err != nil {
+				return err
```
liangchenye: Also
```
return fmt.Errorf("%v\n%v", specerror.NewError(specerror.PoststartHooksInvoke, 
```

liangchenye:
I commented in prestart.go and it missing one case:
```
 If any prestart hook fails, the runtime MUST generate an error, stop the container, and continue the lifecycle at step 9.
```

validation/poststop.go
```diff
@@ -0,0 +1,105 @@
...
+			}
+
+			g.AddPostStopHook(poststop)
+			g.SetProcessArgs([]string{"sh", "-c", fmt.Sprintf("echo 'process called' >> %s", output)})
+			r.SetConfig(g)
+			return nil
+		},
+		PostCreate: func(r *util.Runtime) error {
+			outputData, err := ioutil.ReadFile(output)
+			if err != nil {
```
liangchenye: Similar to 'poststart.go', we should get an error: the output file is not exist.

validation/prestart.go
```diff
@@ -0,0 +1,86 @@
...
+			}
+
+			g.AddPreStartHook(prestart)
+			g.SetProcessArgs([]string{"sh", "-c", fmt.Sprintf("echo 'process called' >> %s", output)})
+			r.SetConfig(g)
+			return nil
+		},
+		PostCreate: func(r *util.Runtime) error {
+			outputData, err := ioutil.ReadFile(output)
+			if err != nil {
```
liangchenye: same, should get an error.

validation/prestart.go
```diff
@@ -0,0 +1,86 @@
...
+				return fmt.Errorf("Unsupported output information: %v", string(outputData))
+			}
+		},
+		PreDelete: func(r *util.Runtime) error {
+			outputData, err := ioutil.ReadFile(output)
+			if err != nil {
+				return err
+			}
+			switch string(outputData) {
+			case "":
```
liangchenye: `""` could merge with `default`

validation/poststop.go
```diff
@@ -0,0 +1,105 @@
...
+				return fmt.Errorf("Unsupported output information: %v", string(outputData))
+			}
+		},
+		PostDelete: func(r *util.Runtime) error {
+			outputData, err := ioutil.ReadFile(output)
+			if err != nil {
+				return err
+			}
+			switch string(outputData) {
+			case "":
```
liangchenye: a part of 'default'

validation/poststart.go
```diff
@@ -0,0 +1,86 @@
...
+			}
+			switch string(outputData) {
+			case "":
```
liangchenye: a part of 'default'

validation/prestart.go
```diff
@@ -0,0 +1,86 @@
...
+			r.SetConfig(g)
+			return nil
+		},
+		PostCreate: func(r *util.Runtime) error {
+			outputData, err := ioutil.ReadFile(output)
+			if err != nil {
+				return err
+			}
+			switch string(outputData) {
+			case "":
```
liangchenye: part of 'default'

validation/prestart.go
```diff
@@ -0,0 +1,86 @@
...
+				return nil
+			case "pre-start called\n":
+				return specerror.NewError(specerror.PrestartTiming, fmt.Errorf("Pre-start hooks MUST be called after the `start` operation is called"), rspec.Version)
+			case "process called\n":
+				return specerror.NewError(specerror.ProcNotRunAtResRequest, fmt.Errorf("The user-specified program (from process) MUST NOT be run at this time"), rspec.Version)
+			case "pre-start called\nprocess called\n", "process called\npre-start called\n":
```
liangchenye: remove this, change 
```
case "pre-start called\n" 
```
to
```
strings.Contain
```

validation/poststop.go
```diff
@@ -0,0 +1,105 @@
...
+			r.SetConfig(g)
+			return nil
+		},
+		PostCreate: func(r *util.Runtime) error {
+			outputData, err := ioutil.ReadFile(output)
+			if err != nil {
+				return err
+			}
+			switch string(outputData) {
+			case "":
```
liangchenye: a part of default

liangchenye:
Same to other hooks.
```
 If any poststart hook fails, the runtime MUST log a warning, but the remaining hooks and lifecycle continue as if the hook had succeeded.
```
```
if any poststop hook fails, the runtime MUST log a warning, but the remaining hooks and lifecycle continue as if the hook had succeeded.
```

q384566678:
@liangchenye @wking updated, PTAL.

liangchenye:
It looks to me.  (but I'm waiting for another day to merge it.)
In my test, I found runc has three issues:
1. prestart/poststart hook is called in 'create'
2. when the poststop hook was set, the main process never work

I'll doubt check the runc code.

liangchenye:
I saw the problem, the process 'output' filepath is wrong.
```g.SetProcessArgs([]string{"sh", "-c", fmt.Sprintf("echo 'process called' >> %s", output)})
```

validation/poststart.go
```diff
@@ -0,0 +1,90 @@
...
+			output = filepath.Join(r.BundleDir, g.Spec().Root.Path, "output")
+			poststart := rspec.Hook{
+				Path: fmt.Sprintf("%s/%s/bin/sh", r.BundleDir, g.Spec().Root.Path),
+				Args: []string{
+					"sh", "-c", fmt.Sprintf("echo 'post-start called' >> %s", output),
+				},
+			}
+
+			g.AddPostStartHook(poststart)
+			g.SetProcessArgs([]string{"sh", "-c", fmt.Sprintf("echo 'process called' >> %s", output)})
```
liangchenye: Change to:
g.SetProcessArgs([]string{"sh", "-c", fmt.Sprintf("echo 'process called' >> %s", "/output")})

validation/poststart.go
```diff
@@ -0,0 +1,90 @@
...
+				if strings.Contains(string(outputData), "post-start called") {
+					return specerror.NewError(specerror.PoststartTiming, fmt.Errorf("The post-start hooks MUST be called before the `start` operation returns"), rspec.Version)
+				} else if strings.Contains(string(outputData), "process called") {
+					return specerror.NewError(specerror.ProcNotRunAtResRequest, fmt.Errorf("The user-specified program (from process) MUST NOT be run at this time"), rspec.Version)
+				}
+				return fmt.Errorf("File %v should not exist", output)
+			}
+			return nil
+		},
+		PreDelete: func(r *util.Runtime) error {
```
liangchenye: I suggest we can add
```
util.WaitingForStatus(*r, util.LifecycleStatusStopped, time.Second * 10, time.Second)
```
 'start' will return immediately, although 'echo' runs fast, but we are not sure how many time it costs. (for example, in runc/cc/Kata project)

validation/poststart.go
```diff
@@ -0,0 +1,90 @@
...
+
+			g.AddPostStartHook(poststart)
+			g.SetProcessArgs([]string{"sh", "-c", fmt.Sprintf("echo 'process called' >> %s", output)})
+			r.SetConfig(g)
+			return nil
+		},
+		PostCreate: func(r *util.Runtime) error {
+			outputData, err := ioutil.ReadFile(output)
+			if err == nil {
+				if strings.Contains(string(outputData), "post-start called") {
```
liangchenye: I always got it in 'runc'!

validation/prestart.go
```diff
@@ -0,0 +1,89 @@
...
+			output = filepath.Join(r.BundleDir, g.Spec().Root.Path, "output")
+			prestart := rspec.Hook{
+				Path: fmt.Sprintf("%s/%s/bin/sh", r.BundleDir, g.Spec().Root.Path),
+				Args: []string{
+					"sh", "-c", fmt.Sprintf("echo 'pre-start called' >> %s", output),
+				},
+			}
+
+			g.AddPreStartHook(prestart)
+			g.SetProcessArgs([]string{"sh", "-c", fmt.Sprintf("echo 'process called' >> %s", output)})
```
liangchenye: Same, the output is wrong.

validation/prestart.go
```diff
@@ -0,0 +1,89 @@
...
+				} else if strings.Contains(string(outputData), "process called") {
+					return specerror.NewError(specerror.ProcNotRunAtResRequest, fmt.Errorf("The user-specified program (from process) MUST NOT be run at this time"), rspec.Version)
+				}
+
+				return fmt.Errorf("File %v should not exist", output)
+			}
+			return nil
+		},
+		PreDelete: func(r *util.Runtime) error {
+			outputData, err := ioutil.ReadFile(output)
```
liangchenye: Same, I suggest to add 'Wait.."

validation/prestart.go
```diff
@@ -0,0 +1,89 @@
...
+
+			g.AddPreStartHook(prestart)
+			g.SetProcessArgs([]string{"sh", "-c", fmt.Sprintf("echo 'process called' >> %s", output)})
+			r.SetConfig(g)
+			return nil
+		},
+		PostCreate: func(r *util.Runtime) error {
+			outputData, err := ioutil.ReadFile(output)
+			if err == nil {
+				if strings.Contains(string(outputData), "pre-start called") {
```
liangchenye: I always get this in 'runc'.

validation/poststart.go
```diff
@@ -0,0 +1,90 @@
...
+			if err != nil {
+				return fmt.Errorf("%v\n%v", specerror.NewError(specerror.PoststartHooksInvoke, fmt.Errorf("The poststart hooks MUST be invoked by the runtime"), rspec.Version), specerror.NewError(specerror.ProcImplement, fmt.Errorf("The runtime MUST run the user-specified program, as specified by `process`"), rspec.Version))
+			}
+			switch string(outputData) {
+			case "post-start called\n":
+				return specerror.NewError(specerror.ProcImplement, fmt.Errorf("The runtime MUST run the user-specified program, as specified by `process`"), rspec.Version)
+			case "process called\n":
+				fmt.Fprintln(os.Stderr, "WARNING: The poststart hook invoke fails")
+				return nil
+			case "post-start called\nprocess called\n":
```
liangchenye: since 'post-start' runs in 'create', the sequence here is wrong.

validation/poststop.go
```diff
@@ -0,0 +1,107 @@
...
+			output = filepath.Join(r.BundleDir, g.Spec().Root.Path, "output")
+			poststop := rspec.Hook{
+				Path: fmt.Sprintf("%s/%s/bin/sh", r.BundleDir, g.Spec().Root.Path),
+				Args: []string{
+					"sh", "-c", fmt.Sprintf("echo 'post-stop called' >> %s", output),
+				},
+			}
+
+			g.AddPostStopHook(poststop)
+			g.SetProcessArgs([]string{"sh", "-c", fmt.Sprintf("echo 'process called' >> %s", output)})
```
liangchenye: same, wrong output

liangchenye:
https://github.com/opencontainers/runc/issues/1574

liangchenye:
poststart hook issue: https://github.com/opencontainers/runc/issues/1135

liangchenye:
Report an issue in 'runc' project: https://github.com/opencontainers/runc/issues/1710.

liangchenye:
LGTM

PR #560, opened 2018-01-24
validate: CheckLinux is platform dependent

vbatts:
Signed-off-by: Vincent Batts <vbatts@hashbangbash.com>


liangchenye:
Hi @vbatts , I think 'validate.go' built in a non-Linux environment could also validate the Linux struct.

wking:
On Wed, Jan 24, 2018 at 08:37:17AM +0000, (Liang Chenye) wrote:
> I think 'validate.go' built in a non-Linux environment could also
> validate the Linux struct.

+1, as long as ‘HostSpecific’ wasn't set (in which case we should
error out).  I'm in favor of moving anything that won't compile on
other platforms (and is only used for HostSpecific checks) over to
Linux-only helper functions, but I'd rather not move *all* of
CheckLinux.

liangchenye:
LGTM
I just reviewed https://github.com/opencontainers/runtime-tools/pull/561. I feel it will be clearer and easier to maintain if we move all the Linux checking/generating works out.

wking:
> I feel it will be clearer and easier to maintain if we move all the Linux checking/generating works out.

But you lose the ability to do host-sgnostic, cross-platform validation.  More on that in #445.  Is cross-platform validation not a goal of this project?

wking:
I've filed #565 with a start at the narrower change I think we want to see, allowing for Windows hosts to compile `validate` (like this PR) but retaining the ability of Windows hosts to perform most host-agnostic validation of the `linux` config (unlike this PR).

q384566678:
Need rebase @vbatts .

vbatts:
updated. PTAL.

q384566678:
LGTM

PR #561, opened 2018-01-24
generate/seccomp: platform independent values

vbatts:
This default seccomp profile may need to be used/generated from
non-linux platforms, though the use of syscall package confines the
compile to linux only

Signed-off-by: Vincent Batts <vbatts@hashbangbash.com>

### DIFF
--- a/generate/seccomp/seccomp_default_linux.go
+++ b/generate/seccomp/seccomp_default_linux.go
@@ -1,0 +1,15 @@
+// +build linux
+
+package seccomp
+
+import "syscall"
+
+// System values passed through on linux
+const (
+	CloneNewIPC  = syscall.CLONE_NEWIPC
+	CloneNewNet  = syscall.CLONE_NEWNET
+	CloneNewNS   = syscall.CLONE_NEWNS
+	CloneNewPID  = syscall.CLONE_NEWPID
+	CloneNewUser = syscall.CLONE_NEWUSER
+	CloneNewUTS  = syscall.CLONE_NEWUTS
+)
--- a/generate/seccomp/seccomp_default_unsupported.go
+++ b/generate/seccomp/seccomp_default_unsupported.go
@@ -1,0 +1,15 @@
+// +build !linux
+
+package seccomp
+
+// These are copied from linux/amd64 syscall values, as a reference for other
+// platforms to have access to
+const (
+	CloneNewIPC    = 0x8000000
+	CloneNewNet    = 0x40000000
+	CloneNewNS     = 0x20000
+	CloneNewPID    = 0x20000000
+	CloneNewUser   = 0x10000000
+	CloneNewUTS    = 0x4000000
+	CloneNewCgroup = 0x02000000
+)
--- a/generate/seccomp/seccomp_default.go
+++ b/generate/seccomp/seccomp_default.go
@@ -2,7 +2,6 @@ package seccomp
 
 import (
 	"runtime"
-	"syscall"
 
 	"github.com/opencontainers/runtime-spec/specs-go"
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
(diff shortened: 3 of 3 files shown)


generate/seccomp/seccomp_default_linux.go
```diff
@@ -0,0 +1,14 @@
+// +build linux
+
+package seccomp
+
+import "syscall"
+
+const (
+	CLONE_NEWIPC  = syscall.CLONE_NEWIPC
```
wking: `golint` [suggests `CloneNewIPC`][1], etc.  Breaking consistency with `syscall` would be unfortunate, but I'd prefer `CloneNewIPC` over hoop-jumping to make `golint` accept `CLONE_NEWIPC`.

[1]: https://travis-ci.org/opencontainers/runtime-tools/jobs/332713642#L537

liangchenye:
please update the naming issue as golint suggest 'CLONE_NEWIPC ', other wise it looks good to me.

generate/seccomp/seccomp_default_unsupported.go
```diff
@@ -0,0 +1,14 @@
+// +build !linux
+
+package seccomp
+
+// These are copied from linux/amd64 syscall values, as a reference for other
+// platforms to have access to
```
wking: If these values were amd64-specific, that would be a problem, because folks generating configs for other arches from non-Linux systems would get the wrong values.  But it turns out that these values are [arch-independent][1].  Maybe adjust this comment to point that out, possibly linking to the kerrnel?

[1]: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/tree/include/uapi/linux/sched.h?h=v4.14.15#n5

generate/seccomp/seccomp_default_unsupported.go
```diff
@@ -0,0 +1,14 @@
...
+const (
+	CLONE_NEWIPC  = 0x8000000
+	CLONE_NEWNET  = 0x40000000
+	CLONE_NEWNS   = 0x20000
+	CLONE_NEWPID  = 0x20000000
+	CLONE_NEWUSER = 0x10000000
+	CLONE_NEWUTS  = 0x4000000
```
wking: You may want to add [`CloneNewCgroup = 0x02000000`][1].  We support the cgroup namespace since opencontainers/runtime-spec#397.  I'm not sure why the seccomp lib doesn't need it (yet?), but I'd rather try to keep full new-namespace coverage.

[1]: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/tree/include/uapi/linux/sched.h?h=v4.14.15#n25

vbatts: it's not even in the go syscall yet. I'm going to hold on this addition.

generate/seccomp/seccomp_default_linux.go
vbatts: fair, but I was hoping to stick to `syscall` format

vbatts:
updated. PTAL

generate/seccomp/seccomp_default_linux.go
```diff
@@ -0,0 +1,15 @@
+// +build linux
+
+package seccomp
+
+import "syscall"
```
wking: `syscall` is [end-of-life][1]:

> NOTE: This package is locked down. Code outside the standard Go repository should be migrated to use the corresponding package in the golang.org/x/sys repository…

We already vendor `golang.org/x/sys` here; can we use that instead?

[1]: https://golang.org/pkg/syscall/

generate/seccomp/seccomp_default_unsupported.go
wking: > it's not even in the go syscall yet. I'm going to hold on this addition.

`syscall` is end-of-life.  It's [in golang.org/x/sys/unix][1] since golang/sys@b44883b47 (2016-06-09).

[1]: https://godoc.org/golang.org/x/sys/unix

liangchenye:
LGTM

PR #562, opened 2018-01-25
add 'state' test; using WaitingForStatus in insideValidation

liangchenye:

### DIFF
--- a/validation/util/test.go
+++ b/validation/util/test.go
@@ -194,8 +194,10 @@ func RuntimeInsideValidate(g *generate.Generator, f PreFunc) (err error) {
 		return err
 	}
 
-	// FIXME: wait until the container exits and collect its exit code.
-	time.Sleep(1 * time.Second)
+	err = WaitingForStatus(r, LifecycleStatusStopped, 10*time.Second, 1*time.Second)
+	if err != nil {
+		return err
+	}
 
 	stdout, stderr, err := r.ReadStandardStreams()
 	if err != nil {
--- a/validation/util/test.go
+++ b/validation/util/test.go
@@ -26,6 +26,8 @@ var (
 type LifecycleAction int
 
 const (
+	// LifecycleActionNone does nothing
+	LifecycleActionNone = 0
 	// LifecycleActionCreate creates a container
 	LifecycleActionCreate = 1 << iota
 	// LifecycleActionStart starts a container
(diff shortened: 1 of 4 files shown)


liangchenye:
1. more diagnostic information
    'r.State' returns an extra stderr 
    PreCreate/PostCreate/PreDelete/PostDelete returns an extra stderr
2. move SetID out of lifecycle function
   Because container ID is also a testing item. 
   But leave bundle prepare in the lifecycle function.
3. Add LifecycleActionNone
   So we can test cases like 'state/start' with invalid/non-exist ID
4. Add state.go 
   Similar approach to current 'create.go'
5. Split 'Delete' from 'r.Clean'
    We don't always need to 'r.Delete' if the container is never not created, so the 'r.Delete' in r.Clean is not always necessary.
    It is up to the developer to 'delete' the container inside the lifecycle function or 'delete' it in other way.

validation/util/test.go
```diff
@@ -60,10 +62,10 @@ var lifecycleStatusMap = map[string]LifecycleStatus{
 // 2. Four phases for user to add his/her own operations.
 type LifecycleConfig struct {
 	Actions    LifecycleAction
-	PreCreate  func(runtime *Runtime) error
-	PostCreate func(runtime *Runtime) error
-	PreDelete  func(runtime *Runtime) error
-	PostDelete func(runtime *Runtime) error
+	PreCreate  func(runtime *Runtime) ([]byte, error)
```
wking: I'd rather not assume that these functions will include something that has a stderr.  Can we use [`*ExitError`][1] instead, and have the test include that stderr in their YAML diagnostic if the error casts to `*ExitError`?  I'll work up a PR with that narrow pivot to help show what I'm suggesting.

[1]: https://golang.org/pkg/os/exec/#ExitError

wking: > I'll work up a PR with that narrow pivot to help show what I'm suggesting.

Filed as #564.

liangchenye:
need to rebase after https://github.com/opencontainers/runtime-tools/pull/563

liangchenye:
Updated. 
The second commit removes the 'fixme' in RuntimeInsideValidation by using 'WaitingForStatus'.

PTAL @wking @Mashimiao @q384566678

q384566678:
LGTM

PR #564, opened 2018-01-26
validation/util/container: Use ExitError for stderr

wking:
Lean on the stdlib's [`ExitError`][1] instead of passing stderr around separately.  This makes it easy to keep all the parts of the error together, without having to worry about whether a particular command may have executed a runtime call.

[1]: https://golang.org/pkg/os/exec/#ExitError

### DIFF
--- a/validation/create.go
+++ b/validation/create.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"os/exec"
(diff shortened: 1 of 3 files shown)


liangchenye:
LGTM

q384566678:
LGTM

PR #565, opened 2018-01-26
validate: Move syscalls into validate_posix

wking:
So the rest of the validation will compile on Windows.  There might be a way to do these `FileInfo` comparisons on Windows, but I haven't had the time to figure that out yet.

This is a narrower alternative to #560 which still allows a Windows host to perform the bulk of the host-agnostic validation for a Linux config.


wking:
With #6560 merged, I'm not sure what the plan is for cross-platform, host-agnostic validation.  Is that no longer a goal?

PR #567, opened 2018-02-05
nil config support in lifecycle validate

liangchenye:


liangchenye:
The meaningless config will be saved in current hooks tests, this PR improves it.

q384566678:
LGTM

PR #569, opened 2018-02-06
add tests when prestart/poststart/poststop hooks fail

liangchenye:

### DIFF
--- a/validation/poststart_fail.go
+++ b/validation/poststart_fail.go
@@ -1,0 +1,70 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"time"
+
+	tap "github.com/mndrix/tap-go"
+	rspec "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/opencontainers/runtime-tools/specerror"
+	"github.com/opencontainers/runtime-tools/validation/util"
+	uuid "github.com/satori/go.uuid"
+)
+
+func main() {
+	t := tap.New()
+	t.Header(0)
+
+	bundleDir, err := util.PrepareBundle()
+	if err != nil {
+		return
+	}
+	defer os.RemoveAll(bundleDir)
+
+	g := util.GetDefaultGenerator()
+	output := filepath.Join(bundleDir, g.Spec().Root.Path, "output")
+	poststart := rspec.Hook{
+		Path: filepath.Join(bundleDir, g.Spec().Root.Path, "/bin/false"),
+		Args: []string{"false"},
+	}
+	g.AddPostStartHook(poststart)
+	poststartOK := rspec.Hook{
+		Path: filepath.Join(bundleDir, g.Spec().Root.Path, "/bin/sh"),
+		Args: []string{
+			"sh", "-c", fmt.Sprintf("echo 'post-start called' >> %s", output),
+		},
+	}
+	g.AddPostStartHook(poststartOK)
+	g.SetProcessArgs([]string{"true"})
+	config := util.LifecycleConfig{
+		Config:    g,
+		BundleDir: bundleDir,
+		Actions:   util.LifecycleActionCreate | util.LifecycleActionStart | util.LifecycleActionDelete,
+		PreCreate: func(r *util.Runtime) error {
+			r.SetID(uuid.NewV4().String())
+			return nil
+		},
+		PreDelete: func(r *util.Runtime) error {
+			util.WaitingForStatus(*r, util.LifecycleStatusStopped, time.Second*10, time.Second)
+			return nil
+		},
+	}
+
+	runErr := util.RuntimeLifecycleValidate(config)
+	outputData, _ := ioutil.ReadFile(output)
+
+	// if runErr is not nil, it means the runtime generates an error
+	// if outputData is not equal to the expected content, it means there is something wrong with the remaining hooks and lifecycle
+	if runErr != nil || string(outputData) != "post-start called\n" {
+		err := specerror.NewError(specerror.PoststartHookFailGenWarn, fmt.Errorf("if any poststart hook fails, the runtime MUST log a warning, but the remaining hooks and lifecycle continue as if the hook had succeeded"), rspec.Version)
+		diagnostic := map[string]string{
+			"error": err.Error(),
+		}
+		t.YAML(diagnostic)
+	}
+
+	t.AutoPlan()
+}
--- a/validation/poststop_fail.go
+++ b/validation/poststop_fail.go
@@ -1,0 +1,70 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"time"
+
+	tap "github.com/mndrix/tap-go"
+	rspec "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/opencontainers/runtime-tools/specerror"
+	"github.com/opencontainers/runtime-tools/validation/util"
+	uuid "github.com/satori/go.uuid"
+)
+
+func main() {
+	t := tap.New()
+	t.Header(0)
+
+	bundleDir, err := util.PrepareBundle()
+	if err != nil {
+		return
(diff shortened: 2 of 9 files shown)


liangchenye:
It should after #567. 
PTAL @q384566678 @wking

validation/poststart_fail.go
```diff
@@ -0,0 +1,48 @@
...
+	t := tap.New()
+	t.Header(0)
+
+	config := util.LifecycleConfig{
+		Actions: util.LifecycleActionCreate | util.LifecycleActionStart | util.LifecycleActionDelete,
+		PreCreate: func(r *util.Runtime) error {
+			r.SetID(uuid.NewV4().String())
+			g := util.GetDefaultGenerator()
+			poststart := rspec.Hook{
+				Path: fmt.Sprintf("%s/%s/bin/false", r.BundleDir, g.Spec().Root.Path),
```
wking: Can we use [`filepath.Join`][1] here?

[1]: https://golang.org/pkg/path/filepath/#Join

validation/poststart_fail.go
```diff
@@ -0,0 +1,48 @@
...
+				Args: []string{"false"},
+			}
+			g.AddPostStartHook(poststart)
```
wking: I think you also want a second hook which writes to a file, and a post-`RuntimeLifecycleValidate` check to ensure that file was never written.  Similarly with `SetProcessArgs`, you'll want to set something there so we can confirm that the runtime didn't actually execute the user-specified program.

validation/poststart_fail.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/opencontainers/runtime-tools/validation/util"
+	uuid "github.com/satori/go.uuid"
+)
+
+func main() {
+	t := tap.New()
+	t.Header(0)
+
+	config := util.LifecycleConfig{
+		Actions: util.LifecycleActionCreate | util.LifecycleActionStart | util.LifecycleActionDelete,
```
wking: No need to set `util.LifecycleActionDelete` or the `PreDelete` code below, because [the `start` call should exit nonzero][1] after the hook failure.  That means `RuntimeLifecycleValidate` will exit [here][2], before any of its delete-related code fires.

[1]: https://github.com/opencontainers/runtime-tools/blob/3eba113f4e36cca8795c9835a437a0f12b8fdb89/docs/command-line-interface.md#start
[2]: https://github.com/opencontainers/runtime-tools/blob/3eba113f4e36cca8795c9835a437a0f12b8fdb89/validation/util/test.go#L319

validation/poststart_fail.go
```diff
@@ -0,0 +1,48 @@
...
+		},
+		PreDelete: func(r *util.Runtime) error {
+			util.WaitingForStatus(*r, util.LifecycleStatusStopped, time.Second*10, time.Second)
+			return nil
+		},
+	}
+
+	err := util.RuntimeLifecycleValidate(nil, config)
+	if err != nil {
+		err := specerror.NewError(specerror.PoststartHookFailGenWarn, fmt.Errorf("if any poststart hook fails, the runtime MUST log a warning, but the remaining hooks and lifecycle continue as if the hook had succeeded"), rspec.Version)
```
wking: Instead of assuming any error is the error we're trying to trigger, we should add a `PostCreate` action that flips a boolean:

```go
successfulCreate := false
config := util.LifecycleConfig{
  …
  PostCreate: func(r *util.Runtime) error {
    successfulCreate = true
    return nil
  },
}
…
if !successfulCreate {
  …fail the test…
} else if err != nil {
  …what you have here…
}
```

validation/poststart_fail.go
liangchenye: https://github.com/opencontainers/runtime-spec/blob/master/runtime.md#lifecycle 
In poststart fail test, it will continue to work. 'start' should not exit.

Also with 'ActionDelete' set, even there is no chance to get 'r.Delete', the container will be deleted in the 'defer' function.
https://github.com/opencontainers/runtime-tools/blob/3eba113f4e36cca8795c9835a437a0f12b8fdb89/validation/util/test.go#L294

validation/poststart_fail.go
liangchenye: OK, I'll update it.

validation/poststart_fail.go
liangchenye: +1 for this, I'll update these.

validation/poststart_fail.go
liangchenye: We don't need to check this, we just need to verify if the 'poststart_fail', the remain hook and the lifecycle is working as expected. The sequence tests of 'create' and 'hooks' are covered in prestart.go/poststart.go/poststop.go .

liangchenye:
1. change the RuntimeLifecycleValidation by
1.1 add a 'BundleDir' to config
      it will be easier to test output out of the lifecycleValidation function
1.2 add a 'generator' to config
      we don't need to have an extra parameter
2. updates testing code by checking the outputs

liangchenye:
One issue found in 'runc':  once a hook fail in 'poststart' or 'poststop', the remain hooks will not be excused.

One issue found in our generator: we cannot have hooks with a same Path. https://github.com/opencontainers/runtime-tools/blob/master/generate/generate.go#L918
The spec does not forbid it.

validation/poststart_fail.go
```diff
@@ -0,0 +1,77 @@
...
+			"sh", "-c", fmt.Sprintf("echo 'post-start called' >> %s", output),
+		},
+	}
+	g.AddPostStartHook(poststartOK)
+	poststop := rspec.Hook{
+		Path: filepath.Join(bundleDir, g.Spec().Root.Path, "/bin/sh"),
+		Args: []string{
+			"sh", "-c", fmt.Sprintf("echo 'post-stop called' >> %s", output),
+		},
+	}
```
q384566678: I don't think this `poststop` is needed. If the final output is not `post-start called\npost-stop called\n`, it's not immediately clear that it's poststart's error, and further judgment is needed.  So I think this `poststop` should be removed to remove unnecessary effects.

validation/poststop_fail.go
```diff
@@ -0,0 +1,70 @@
...
+
+	g := util.GetDefaultGenerator()
+	output := filepath.Join(bundleDir, g.Spec().Root.Path, "output")
+	poststop := rspec.Hook{
+		Path: filepath.Join(bundleDir, g.Spec().Root.Path, "/bin/false"),
+		Args: []string{"false"},
+	}
+	g.AddPostStopHook(poststop)
+	poststopOK := rspec.Hook{
+		Path: filepath.Join(bundleDir, g.Spec().Root.Path, "/bin/echo"),
```
q384566678: /bin/echo -> /bin/sh

validation/poststart_fail.go
liangchenye: you are right, updated.

validation/poststop_fail.go
liangchenye: thanks

liangchenye:
updated PTAL @q384566678 

1. Remove the unnecessary 'poststop' hook check in poststart_fail.
2. Correct the wrong `path`

validation/poststop_fail.go
```diff
@@ -0,0 +1,71 @@
...
+			return nil
+		},
+	}
+
+	runErr := util.RuntimeLifecycleValidate(config)
+	outputData, _ := ioutil.ReadFile(output)
+	// if runErr is not nil, it means the runtime generates an error
+	// if outputData is not equal to the expected content, it means there is something wrong with the remaining hooks and lifecycle
+	//	if runErr != nil || string(outputData) != "post-stop called\n" {
+	if runErr != nil || string(outputData) != "" {
```
q384566678: I  think the line you commented out is correct.
```
//	if runErr != nil || string(outputData) != "post-stop called\n" {

liangchenye: ops, big mistake.
Updated

q384566678:
LGTM

PR #571, opened 2018-02-07
don't overwrite hook which has a same path

liangchenye:


liangchenye:
The issue is found in https://github.com/opencontainers/runtime-tools/pull/569.

For history reason, we add a hook and add hook env separately, so the 'path' is the unique ID to add env correctly. Now we are adding a full hook with json format, so we don't need to have this limitation. 
Also, in the spec, it is OK to have hooks with the same 'path'.

liangchenye:
PTAL @q384566678 @Mashimiao

q384566678:
LGTM

PR #577, opened 2018-02-08
validation: implement PosixHooksCalledInOrder test

q384566678:

### DIFF
--- a/validation/hooks.go
+++ b/validation/hooks.go
@@ -1,0 +1,114 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os/exec"
+	"path/filepath"
+	"time"
+
+	tap "github.com/mndrix/tap-go"
+	rspec "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/opencontainers/runtime-tools/specerror"
+	"github.com/opencontainers/runtime-tools/validation/util"
+	uuid "github.com/satori/go.uuid"
+)
+
+func main() {
+	t := tap.New()
+	t.Header(0)
+
+	var output string
+	config := util.LifecycleConfig{
+		Actions: util.LifecycleActionCreate | util.LifecycleActionStart | util.LifecycleActionDelete,
+		PreCreate: func(r *util.Runtime) error {
+			r.SetID(uuid.NewV4().String())
+			g := util.GetDefaultGenerator()
+			output = filepath.Join(r.BundleDir, g.Spec().Root.Path, "output")
+			shPath := filepath.Join(r.BundleDir, g.Spec().Root.Path, "/bin/sh")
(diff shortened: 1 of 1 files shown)


liangchenye:
1. I prefer to rename it to hooksorder.go .
2.  We need to verify 'poststart/poststop' as well. 
```    
   echo prestart1 > output_prestart
   echo prestart2 > output_prestart
   verify prestart hook order
   echo poststart1 > output_poststart
   echo poststart2 > output_poststart
   verify poststart hook order
   echo poststop1 > output_poststop
   echo poststop2 > output_poststop
   verify poststop hook order
```

liangchenye:
I think in the order tests, we have the assumption: every hook ('echo') will be executed collect， we just test the 'order'.

q384566678:
updated, and I think hook.go would be better.

validation/hooks.go
```diff
@@ -0,0 +1,101 @@
...
+			r.SetID(uuid.NewV4().String())
+			g := util.GetDefaultGenerator()
+			output = filepath.Join(r.BundleDir, g.Spec().Root.Path, "output")
+			prestart1 := rspec.Hook{
+				Path: filepath.Join(r.BundleDir, g.Spec().Root.Path, "/bin/sh"),
+				Args: []string{
+					"sh", "-c", fmt.Sprintf("echo 'pre-start1 called' >> %s", output),
+				},
+			}
+			g.AddPreStartHook(prestart1)
```
wking: No need for a `prestart1` temporary variable, you can just use:

```go
err := g.AddPreStartHook(rspec.Hook{
  …
})
if err != nil {
  return err
}
```

That also checks this call's returned error, which you're currently ignoring.

validation/hooks.go
```diff
@@ -0,0 +1,101 @@
...
+
+	var output string
+	config := util.LifecycleConfig{
+		Actions: util.LifecycleActionCreate | util.LifecycleActionStart | util.LifecycleActionDelete,
+		PreCreate: func(r *util.Runtime) error {
+			r.SetID(uuid.NewV4().String())
+			g := util.GetDefaultGenerator()
+			output = filepath.Join(r.BundleDir, g.Spec().Root.Path, "output")
+			prestart1 := rspec.Hook{
+				Path: filepath.Join(r.BundleDir, g.Spec().Root.Path, "/bin/sh"),
```
wking: You use this `Join` a lot in this test.  Maybe store it in a temporary variable:

```go
shPath := filepath.Join(r.BundleDir, g.Spec().Root.Path, "/bin/sh")
…
  Path: shPath,
```

q384566678: updated, thanks.

liangchenye:
LGTM

PR #578, opened 2018-02-08
implement start operation tests

liangchenye:

### DIFF
--- a/validation/start.go
+++ b/validation/start.go
@@ -1,0 +1,104 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/mndrix/tap-go"
+	rspecs "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/opencontainers/runtime-tools/specerror"
+	"github.com/opencontainers/runtime-tools/validation/util"
+	uuid "github.com/satori/go.uuid"
+)
+
+func main() {
+	t := tap.New()
+	t.Header(0)
+
+	bundleDir, err := util.PrepareBundle()
+	if err != nil {
+		util.Fatal(err)
+	}
+	defer os.RemoveAll(bundleDir)
+
+	containerID := uuid.NewV4().String()
+
+	r, err := util.NewRuntime(util.RuntimeCommand, bundleDir)
+	if err != nil {
+		util.Fatal(err)
+	}
+	g := util.GetDefaultGenerator()
+	g.SetProcessArgs([]string{"sh", "-c", fmt.Sprintf("echo 'process called' >> %s", "/output")})
+	r.SetConfig(g)
+	output := filepath.Join(bundleDir, g.Spec().Root.Path, "output")
+
+	// start without id
+	err = r.Start()
+	util.SpecErrorOK(t, err != nil, specerror.NewError(specerror.StartWithoutIDGenError, fmt.Errorf("start` operation MUST generate an error if it is not provided the container ID"), rspecs.Version), err)
+
+	// set id for the remaining tests
+	r.SetID(containerID)
+
+	// start a not `created` container - case one: non-exist container
+	err = r.Start()
+	util.SpecErrorOK(t, err != nil, specerror.NewError(specerror.StartNotCreatedGenError, fmt.Errorf("attempting to `start` a container that is not `created` MUST generate an error"), rspecs.Version), err)
+
+	err = r.Create()
+	if err != nil {
+		util.Fatal(err)
+	}
+	// start a `created` container
+	err = r.Start()
+	if err != nil {
+		util.SpecErrorOK(t, false, specerror.NewError(specerror.StartProcImplement, fmt.Errorf("`start` operation MUST run the user-specified program as specified by `process`"), rspecs.Version), err)
+	} else {
+		err := util.WaitingForStatus(r, util.LifecycleStatusStopped, time.Second*10, time.Second*1)
+		if err != nil {
+			util.Fatal(err)
+		}
+		outputData, outputErr := ioutil.ReadFile(output)
+		// check the output
+		util.SpecErrorOK(t, outputErr == nil && string(outputData) == "process called\n", specerror.NewError(specerror.StartProcImplement, fmt.Errorf("`start` operation MUST run the user-specified program as specified by `process`"), rspecs.Version), outputErr)
+	}
+
+	// start a not `created` container - case two: exist and `stopped`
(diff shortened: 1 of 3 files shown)


liangchenye:
1. rename spec error code, 'NotCreated' is more accurate 
   'StartNonCreateHaveNoEffect' to 'StartNotCreatedHaveNoEffect'
   'StartNonCreateGenError' to 'StartNotCreatedGenError'
2. test the five cases of 'start' operation, two cases might be controversial
2.1 StartNotCreatedHaveNoEffect
      I understand it in this way: if a container is not 'created', the process should not be called again.
      So the output file MUST remain 'process called\n'
2.2 StartWithProcUnsetGenError
      In the spec, a runtime could apply some of the 'process' configs, so if 'process' is set to nil, this runtime might return an error. In this case, we can never reach 'StartWithProcUnsetGenError', so I skip this test if runtime fails in 'create' with nil process. Only if the runtime COULD create successfully, we can test this.

validation/start.go
```diff
@@ -0,0 +1,110 @@
...
+	"time"
+
+	"github.com/mndrix/tap-go"
+	rspecs "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/opencontainers/runtime-tools/specerror"
+	"github.com/opencontainers/runtime-tools/validation/util"
+	uuid "github.com/satori/go.uuid"
+)
+
+func wrap(t *tap.T, expected bool, specErr error, detailedErr error) {
```
wking: This is probably useful/generic enough to be a public function in `validation/util/test.go`.  Maybe `SpecErrorOk()`?

validation/start.go
```diff
@@ -0,0 +1,110 @@
...
+	r.SetID(containerID)
+
+	// start a not `created` container - case one: non-exist container
+	err = r.Start()
+	wrap(t, err != nil, specerror.NewError(specerror.StartNotCreatedGenError, fmt.Errorf("attempting to `start` a container that is not `created` MUST generate an error"), rspecs.Version), err)
+
+	r.Create()
+	// start a `created` container
+	err = r.Start()
+	outputData, _ := ioutil.ReadFile(output)
```
wking: This is racy.  `start` should [block until the user-specified program has executed][1], but there's no guarantee about how far along that program is in its execution.  That means this read could be happening either before or after the user-specified program gets around to writing to `output`.  The coarse guard for this would be to sleep for some safe-enough time.  The better, but maybe Linux-specific,  guard would be to use `inotify(7)` to block until the file was updated (or some safe-enough time had passed) and compare the measured write time with the `start`-invocation time (the write time should be after `start` was invoked and also after any `prestart` hooks completed).

[1]: https://github.com/opencontainers/runtime-tools/blame/64e1103b1518c53911bd163765ae6391c3d54b19/docs/command-line-interface.md#L159

validation/start.go
```diff
@@ -0,0 +1,110 @@
...
+	// check the output
+	wrap(t, err == nil && string(outputData) == "process called\n", specerror.NewError(specerror.StartProcImplement, fmt.Errorf("`start` operation MUST run the user-specified program as specified by `process`"), rspecs.Version), err)
+
+	// start a not `created` container - case two: exist and `stopped`
+	util.WaitingForStatus(r, util.LifecycleStatusStopped, time.Second*10, time.Second*1)
+	err = r.Start()
+	// must generate an error
+	wrap(t, err != nil, specerror.NewError(specerror.StartNotCreatedGenError, fmt.Errorf("attempting to `start` a container that is not `created` MUST generate an error"), rspecs.Version), err)
+
+	outputData, _ = ioutil.ReadFile(output)
```
wking: You shouldn't be ignoring the returned error here (or in your other `ReadFile` calls).  If one of those errors, you can die with a non-zero exit.  node-tap will complain about that:

```console
$ cat validation/test 
#!/bin/sh
cat <<EOF
TAP version 13
ok 1 - ok
EOF
exit 1
$ tap ./validation/test 
./validation/test ..................................... 1/2
  not ok ./validation/test
    timeout: 30000
    file: ./validation/test
    command: ./validation/test
    args: []
    stdio:
      - 0
      - pipe
      - 2
    cwd: /home/wking/.local/lib/go/src/github.com/opencontainers/runtime-tools
    failures:
      - tapError: no plan
    exitCode: 1
…
```

validation/start.go
```diff
@@ -0,0 +1,110 @@
...
+	// must have no effect, it will not be something like 'process called\nprocess called\n'
+	wrap(t, string(outputData) == "process called\n", specerror.NewError(specerror.StartNotCreatedHaveNoEffect, fmt.Errorf("attempting to `start` a container that is not `created` MUST have no effect on the container"), rspecs.Version), err)
+
+	r.Delete()
```
wking: You have another ignored error here.

validation/start.go
```diff
@@ -0,0 +1,110 @@
...
+
+	g.Spec().Process = nil
+	r.SetConfig(g)
+	err = r.Create()
+	// 'create' could fail according to the spec when the process is nil
```
wking: Why “could fail”?  Is this the “[`create` can fail whenever it wants][1]” loophole (opencontainers/runtime-spec#813, opencontainers/runtime-spec#927)?  I don't see a need to call that out in this particular case.  If we want, we could adjust our `Create` wrapper to handle this in all cases.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/runtime.md#L106

validation/start.go
liangchenye: I'll add time sleep here.

validation/start.go
liangchenye: If the process is nil, the config.json is invalid. It is up to the runtime to validate it in the 'create' since:  `The runtime MAY validate config.json against this spec.`

If a runtime validates config.json in 'create', it might return error. In this case, the container is not created, we can never test 'StartWithProcUnsetGenError'.  That is why I use 'Skip' in this case.

liangchenye:
updated
1. add a new SpecErrorOK function to util/test.go
2. check unexpected errors

The 'StartWithProcUnsetGenError' is not changed, I replied in @wking 's comment.

validation/start.go
wking: > If the process is nil, the `config.json` is invalid.

That's not true since opencontainers/runtime-spec#701, which was part of runtime-spec v1.0.0-rc6 and later.

liangchenye: go it.

liangchenye:
updated PTAL @q384566678 @wking

q384566678:
LGTM

PR #580, opened 2018-02-12
implement kill tests

liangchenye:

### DIFF
--- a/validation/kill.go
+++ b/validation/kill.go
@@ -1,0 +1,75 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/mndrix/tap-go"
+	rspecs "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/opencontainers/runtime-tools/generate"
+	"github.com/opencontainers/runtime-tools/specerror"
+	"github.com/opencontainers/runtime-tools/validation/util"
+	uuid "github.com/satori/go.uuid"
+)
+
+func main() {
+	t := tap.New()
+	t.Header(0)
+	bundleDir, err := util.PrepareBundle()
+	if err != nil {
+		util.Fatal(err)
+	}
+	defer os.RemoveAll(bundleDir)
+
+	stoppedConfig := util.GetDefaultGenerator()
+	stoppedConfig.SetProcessArgs([]string{"true"})
+	runningConfig := util.GetDefaultGenerator()
+	runningConfig.SetProcessArgs([]string{"sleep", "30"})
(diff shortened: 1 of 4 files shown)


liangchenye:
- add 'kill' interface to container.go
  our interface docucment [--signal <SIGNAL>](https://github.com/opencontainers/runtime-tools/blob/master/docs/command-line-interface.md#kill) is different with current runc.
   I'm using runc way now.
- fix bug in test.go
- add kill.go 
  I did not find a good way to test this spec error ` Attempting to send a signal to a container that is neither created nor running MUST have no effect on the container`, so it is missing in this PR.
- add killsig.go
   now  'TERM' and 'KILL' are MUST supported, 'KILL' is verified in 'kill.go'.
   In killsig.go, we only verify 'TERM' now.
   In the future we might add more signal tests.  

Tested in my environment with 'runc' , it works.
PTAL  @wking @q384566678

validation/kill.go
```diff
@@ -0,0 +1,76 @@
...
+				r.SetID(c.id)
+				return nil
+			},
+			PreDelete: func(r *util.Runtime) error {
+				// waiting the 'stoppedConfig' to stop
+				// the running test case sleeps 30 seconds, so 10 seconds are enough for this case
+				util.WaitingForStatus(*r, util.LifecycleStatusCreated|util.LifecycleStatusStopped, time.Second*10, time.Second*1)
+				// KILL MUST be supported and KILL cannot be trapped
+				err := r.Kill("KILL")
+				// wait before the container been deleted
```
wking: I dont think the next line needs a comment; `util.WaitingForStatus(*r, util.LifecycleStatusStopped, ...)` is pretty clear on it's own.  And if you do want a comment, you probably want "been deleted" ->"has stopped".

liangchenye: I remove this line, updated.

liangchenye:
Another issue I found, 'util.Fatal' exits immediately, the 'defer' functions will not be called.
So the bunde dir might not be removed and the container might not be deleted.
It needs to be fixed.

validation/killsig.go
```diff
@@ -0,0 +1,60 @@
...
+
+var signals = []string{
+	"TERM",
+}
+
+func main() {
+	t := tap.New()
+	t.Header(0)
+	bundleDir, err := util.PrepareBundle()
+	fmt.Println(bundleDir)
```
q384566678: Should delete it?

liangchenye: updated

q384566678:
LGTM

PR #583, opened 2018-02-13
 add 'delete' tests

liangchenye:


liangchenye:
Based on #580  since it uses 'Kill' in the tests. So #580  should be accepted first.

It does not cover `Deleting a container MUST delete the resources that were created during the create step. Note that resources associated with the container, but not created by this container, MUST NOT be deleted.  ` since I don't have a good idea about 'resources' yet.

One issue found in 'runc'. 'runc' could delete a 'created' container, it does not follow `Attempting to delete a container that is not stopped MUST have no effect on the container and MUST generate an error. ` I checked the change log of runtime spec. In https://github.com/opencontainers/runtime-spec/pull/875, it suggests that a user should 'kill' a 'created' container before 'delete' it. But in runc, it will do the 'kill' operation for the user.  https://github.com/opencontainers/runc/blob/master/delete.go#L78

liangchenye:
updated, PTAL @q384566678

q384566678:
LGTM

PR #589, opened 2018-03-05
add hooks stdin test

liangchenye:

### DIFF
--- a/validation/hooks_stdin.go
+++ b/validation/hooks_stdin.go
@@ -1,0 +1,125 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"reflect"
+	"time"
+
+	tap "github.com/mndrix/tap-go"
+	rspecs "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/opencontainers/runtime-tools/specerror"
+	"github.com/opencontainers/runtime-tools/validation/util"
+	uuid "github.com/satori/go.uuid"
+)
+
+func stdinStateCheck(outputDir, hookName string, expectedState rspecs.State) error {
+	var state rspecs.State
+	data, err := ioutil.ReadFile(filepath.Join(outputDir, hookName))
+	if err != nil {
+		return err
+	}
+	err = json.Unmarshal(data, &state)
+	if err != nil {
+		return err
+	}
+
+	if state.ID != expectedState.ID {
+		return fmt.Errorf("wrong container ID %q in the stdin of %s hook, expected %q", state.ID, hookName, expectedState.ID)
+	}
+
+	if state.Bundle != expectedState.Bundle {
+		return fmt.Errorf("wrong bundle directory %q in the stdin of %s hook, expected %q", state.Bundle, hookName, expectedState.Bundle)
+	}
+
+	if hookName != "poststop" && state.Pid != expectedState.Pid {
+		return fmt.Errorf("wrong container process ID %q in the stdin of %s hook, expected %q", state.Version, hookName, expectedState.Version)
+	}
+
+	if !reflect.DeepEqual(state.Annotations, expectedState.Annotations) {
+		return fmt.Errorf("wrong annotations \"%v\" in the stdin of %s hook, expected \"%v\"", state.Annotations, hookName, expectedState.Annotations)
+	}
+	return nil
+}
+
+func main() {
+	t := tap.New()
+	t.Header(0)
+
+	bundleDir, err := util.PrepareBundle()
+	if err != nil {
+		util.Fatal(err)
+	}
+	containerID := uuid.NewV4().String()
+	defer os.RemoveAll(bundleDir)
+
+	var containerPid int
+
+	annotationKey := "org.opencontainers.runtime-tools"
+	annotationValue := "hook stdin test"
+	g := util.GetDefaultGenerator()
+	outputDir := filepath.Join(bundleDir, g.Spec().Root.Path)
+	timeout := 1
+	g.AddAnnotation(annotationKey, annotationValue)
+	g.AddPreStartHook(rspecs.Hook{
+		Path: filepath.Join(bundleDir, g.Spec().Root.Path, "/bin/sh"),
+		Args: []string{
+			"sh", "-c", fmt.Sprintf("cat > %s", filepath.Join(outputDir, "prestart")),
+		},
+		Timeout: &timeout,
+	})
+	g.AddPostStartHook(rspecs.Hook{
+		Path: filepath.Join(bundleDir, g.Spec().Root.Path, "/bin/sh"),
+		Args: []string{
+			"sh", "-c", fmt.Sprintf("cat > %s", filepath.Join(outputDir, "poststart")),
+		},
+		Timeout: &timeout,
+	})
+	g.AddPostStopHook(rspecs.Hook{
+		Path: filepath.Join(bundleDir, g.Spec().Root.Path, "/bin/sh"),
+		Args: []string{
+			"sh", "-c", fmt.Sprintf("cat > %s", filepath.Join(outputDir, "poststop")),
+		},
+		Timeout: &timeout,
+	})
+	g.SetProcessArgs([]string{"true"})
+	config := util.LifecycleConfig{
+		BundleDir: bundleDir,
+		Config:    g,
+		Actions:   util.LifecycleActionCreate | util.LifecycleActionStart | util.LifecycleActionDelete,
+		PreCreate: func(r *util.Runtime) error {
+			r.SetID(containerID)
+			return nil
+		},
+		PreDelete: func(r *util.Runtime) error {
+			state, err := r.State()
+			if err != nil {
+				return err
+			}
+			containerPid = state.Pid
+			util.WaitingForStatus(*r, util.LifecycleStatusStopped, time.Second*10, time.Second)
+			return nil
+		},
+	}
+
+	err = util.RuntimeLifecycleValidate(config)
+	if err != nil {
+		util.Fatal(err)
+	}
+
+	expectedState := rspecs.State{
+		Pid:         containerPid,
+		ID:          containerID,
+		Bundle:      bundleDir,
+		Annotations: map[string]string{annotationKey: annotationValue},
+	}
+	for _, file := range []string{"prestart", "poststart", "poststop"} {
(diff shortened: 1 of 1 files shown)


liangchenye:
It is a WIP PR to test the container state that passed to hooks over 'stdin':
```
The state of the container MUST be passed to hooks over stdin so that they may do work appropriate to the current state of the container.
```
I try to use `timeout 1 cat > output` to collect stdin content and compare the output file to do the test.
One problem is, it is not possible for the 'poststop' hook to get the container state since the container would be deleted before 'poststop hook' been called.
```
Poststop
The post-stop hooks MUST be called after the container is deleted but before the delete operation returns. Cleanup or debugging functions are examples of such a hook.
```

alban:
> One problem is, it is not possible for the 'poststop' hook to get the container state since the container would be deleted before 'poststop hook' been called.

It's not possible to call `funC state container-id` and compare with the output in the hook but we could still check that the [state](https://github.com/opencontainers/runtime-spec/blob/master/runtime.md#state) has the correct `id`, `bundle` and `annotations`.

I had a look at runc but it always gives an empty "status" (https://github.com/opencontainers/runc/issues/1057 & https://github.com/opencontainers/runc/pull/1741) and as you pointed out, it does not call the hooks at the correct time (https://github.com/opencontainers/runc/issues/1710).

Since the spec does not have a "status=deleted", I wonder what's the correct status to send on stdin in the poststop hook. Does the spec explain that?

liangchenye:
No, if it is deleted, there won't be a 'status=delete' since:
```
Once a container is deleted its ID MAY be used by a subsequent container
```
I think we can not send a status on stdin in the poststop hook.

liangchenye:
In the updated version, I check the 'ID', 'Bundle' and 'Annotations' of the container state in 'prestart', 'poststart' and 'poststop' hooks. The 'PID'，'status' and 'ociVersion' is unpredictable.

The 'poststop' hook is also checked here according to current spec requirement.

liangchenye:
PTAL @q384566678 @wking @alban

wking:
>I think we can not send a status on stdin in the poststop hook.

This is opencontainers/runtime-spec#958.  We may end up with a `deleted` state, but as it stands I think runtimes have to use `stopped` there.

> The 'PID'，'status' and 'ociVersion' is unpredictable.

Depending on how you read [the spec][1], `ociVersion` needs to match either the config version or the state version.  I have opencontainers/runtime-spec#903 in flight to settle on the latter.  But until runtime-spec cuts a release with a different/expanded state schema, I think we can safely use:

```go
if state.Version != "1.0.0" && state.Version != "1.0.1" {
    // complain
}
```

The container process can check its PID, and then this test can compare with the states it recieved.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/runtime.md#L12

liangchenye:
I'm working on the PID now.
About the spec version, my new update is just comparing the `Version:     rspecs.Version,`

validation/hooks_stdin.go
```diff
@@ -0,0 +1,126 @@
...
+
+	annotationKey := "org.opencontainers.runtime-tools"
+	annotationValue := "hook stdin test"
+	g := util.GetDefaultGenerator()
+	outputDir := filepath.Join(bundleDir, g.Spec().Root.Path)
+	g.AddAnnotation(annotationKey, annotationValue)
+	g.AddPreStartHook(rspecs.Hook{
+		Path: filepath.Join(bundleDir, g.Spec().Root.Path, "/bin/sh"),
+		Args: []string{
+			"sh", "-c", fmt.Sprintf("timeout -t 1 cat > %s", filepath.Join(outputDir, "prestart")),
```
wking: Where is your `timeout` from.  It's [not in POSIX][1], and [coreutils uses `-k` / `--kill-after`][2], not `-t`.

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/idx/utilities.html
[2]: http://man7.org/linux/man-pages/man1/timeout.1.html

validation/hooks_stdin.go
```diff
@@ -0,0 +1,126 @@
...
+		PreCreate: func(r *util.Runtime) error {
+			r.SetID(containerID)
+			return nil
+		},
+		PreDelete: func(r *util.Runtime) error {
+			state, err := r.State()
+			if err != nil {
+				return err
+			}
+			containerPid = state.Pid
```
wking: I think this would be more reliable if we dropped the `state` call and collected this from the container process itself (where you currently just use `true`).

liangchenye:
I'll revoke my version test. It is wrong.

validation/hooks_stdin.go
liangchenye: It is in the busybox 1.28, in the new rootfs-amd64.tar.gz.

wking: > It is in the busybox 1.28, in the new rootfs-amd64.tar.gz.

The lack of standardization is unfortunate, because hooks are called from the runtime namespace, so the version in the container doesn't matter.  We could work around that with `{bundle-path}/rootfs/bin/timeout` here, but it's probably better to set `Timeout` in the hook structure and skip the `timeout` command.

validation/hooks_stdin.go
liangchenye: `echo $$` will output the id inside the container. It is different with the state one - `It is 'seen' by host.`

validation/hooks_stdin.go
liangchenye: thanks, the timeout in hooks is much more rational.

validation/hooks_stdin.go
wking: > `echo $$` will output the id inside the container. It is different with the state one - `It is 'seen' by host.`

Ah right, you'd need the PID namespace too to translate, and that would break down for VM-based runtimes anyway.  So I'm on board with your current `state` call.

liangchenye:
Updated.
The 'version' test is dropped. I don't find a good definition of an expected version in the runtime spec.
We can get a 'Version' from r.State just like what we do to PID. But I don't see the meaning of doing this.

validation/hooks_stdin.go
```diff
@@ -0,0 +1,125 @@
...
+
+	expectedState := rspecs.State{
+		Pid:         containerPid,
+		ID:          containerID,
+		Bundle:      bundleDir,
+		Annotations: map[string]string{annotationKey: annotationValue},
+	}
+	for _, file := range []string{"prestart", "poststart", "poststop"} {
+		err := stdinStateCheck(outputDir, file, expectedState)
+		util.SpecErrorOK(t, err == nil, specerror.NewError(specerror.PosixHooksStateToStdin, fmt.Errorf("the state of the container MUST be passed to %q hook over stdin so that they may do work appropriate to the current state of the container", file), rspecs.Version), err)
```
wking: You can probably drop "so that they...of the container" from the error message.  That motivation is in the spec for folks who care, but you don't need it for the check.

liangchenye: updated.

q384566678:
LGTM

validation/hooks_stdin.go
```diff
@@ -0,0 +1,125 @@
...
+	if state.Bundle != expectedState.Bundle {
+		return fmt.Errorf("wrong bundle directory %q in the stdin of %s hook, expected %q", state.Bundle, hookName, expectedState.Bundle)
+	}
+
+	if hookName != "poststop" && state.Pid != expectedState.Pid {
+		return fmt.Errorf("wrong container process ID %q in the stdin of %s hook, expected %q", state.Version, hookName, expectedState.Version)
+	}
+
+	if !reflect.DeepEqual(state.Annotations, expectedState.Annotations) {
+		return fmt.Errorf("wrong annotations \"%v\" in the stdin of %s hook, expected \"%v\"", state.Annotations, hookName, state.Annotations)
```
q384566678: The last parameter should be `expectedState.Annotations`.

q384566678:
The following error occurred while I was testing, so is this normal?

```
TAP version 13
not ok 1 - the state of the container MUST be passed to "prestart" hook over stdin
  ---
  {
    "error": "wrong annotations \"map[]\" in the stdin of prestart hook, expected \"map[org.opencontainers.runtime-tools:hook stdin test]\"",
    "reference": "https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#posix-platform-hooks"
  }
  ...
not ok 2 - the state of the container MUST be passed to "poststart" hook over stdin
  ---
  {
    "error": "wrong annotations \"map[]\" in the stdin of poststart hook, expected \"map[org.opencontainers.runtime-tools:hook stdin test]\"",
    "reference": "https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#posix-platform-hooks"
  }
  ...
not ok 3 - the state of the container MUST be passed to "poststop" hook over stdin
  ---
  {
    "error": "wrong annotations \"map[]\" in the stdin of poststop hook, expected \"map[org.opencontainers.runtime-tools:hook stdin test]\"",
    "reference": "https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#posix-platform-hooks"
  }
  ...
1..3
```

validation/hooks_stdin.go
liangchenye: updated.

liangchenye:
@q384566678  which 'runc' version do you use? Annotations are supported in mine.
```
runc --version
...
commit: e9cdc3906f18...
...
```

q384566678:
```
runc --version
runc version 1.0.0-rc4+dev
commit: 91e979501348cb4cb13b5fb4437cc5d9ecd94b5d
spec: 1.0.0
```

q384566678:
It's ok when I update the `runc`.

q384566678:
LGTM

wking:
>It's ok when I update the `runc`.

We should really check `status` to notice runc failing becaue of opencontainers/runc#1710 and opencontainers/runc#1741.

liangchenye:
My understanding of the related spec is:
`prestart` hook status SHOULD not be 'creating/running'.
`poststart` hook status SHOULD be 'running'.
`poststop` hook status SHOULD not be 'creating/created/running/stopped'. (maybe a runtime defined status)

##Prestart
```
The pre-start hooks MUST be called after the start operation is called but before the user-specified program command is executed. ...
```

##Poststart
```
The post-start hooks MUST be called after the user-specified process is executed but before the start operation returns....
```

##Poststop
```
The post-stop hooks MUST be called after the container is deleted but before the delete operation returns....
```

PR #595, opened 2018-03-08
Fix condition in BlockIO test

alban:
Symptoms:
```
$ sudo validation/linux_cgroups_relative_blkio.t

 not ok 6 - blkio leafWeight for 8:0 is set correctly
 # expect: 300, actual: 300
```

After the patch:
```
 ok 6 - blkio leafWeight for 8:0 is set correctly
 # expect: 300, actual: 300
```


q384566678:
LGTM

PR #596, opened 2018-03-08
runtimetest: fix uid_map parsing

alban:
The fields of /proc/$pid/uid_map and /proc/$pid/gid_map were parsed in
the wrong order.

Related: https://github.com/opencontainers/runtime-spec/pull/956


liangchenye:
LGTM

PR #597, opened 2018-03-08
validation: LinuxUIDMapping: fix tests

alban:
Don't validate uid mappings and gid mappings separately: containers with
only user mappings or with only group mappings are not usable.

Additionally, don't rely on the runtime to create the directories to be
mounted. runc mounts them in the easy cases but it does not work with
user namespaces.

Marking as WIP/RFC because this is in discussion in https://github.com/opencontainers/runtime-spec/pull/955


wking:
On Thu, Mar 08, 2018 at 10:59:26AM -0800, Alban Crequy wrote:
> Additionnally, …

nit: typo; should be “Additionally”.  In the commit message too.

validation/linux_uid_mappings.go
```diff
@@ -1,14 +1,38 @@
...
 	g := util.GetDefaultGenerator()
 	g.AddOrReplaceLinuxNamespace("user", "")
-	g.AddLinuxUIDMapping(uint32(1000), uint32(0), uint32(3200))
-	err := util.RuntimeInsideValidate(g, nil)
+	g.AddLinuxUIDMapping(uint32(1000), uint32(0), uint32(2000))
+	g.AddLinuxGIDMapping(uint32(1000), uint32(0), uint32(3000))
+	err := util.RuntimeInsideValidate(g, func(path string) error {
+		_ = mkdir(path, "proc")
+		_ = mkdir(path, "dev")
+		_ = mkdir(path, "sys")
```
wking: If we need these (and it sounds like we do for now), I think they should be part of [`PrepareBundle`][1].  Until we have clearer spec wording around creation, I'd rather not rely on the details of runc's “sometimes we can do this” implementation.

[1]: https://github.com/opencontainers/runtime-tools/blob/v0.5.0/validation/util/test.go#L105-L122

alban:
I added `/proc`, `/dev`, `/sys` in the list of files for the rootfs and I had to update the Gentoo scripts.

The tests about uid/gid mappings now pass with runc:
```
$ sudo validation/linux_uid_mappings.t
TAP version 13
not ok 1 - root filesystem
  ---
  {
    "error": "rootfs must not be readonly\nRefer to: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#root"
  }
  ...
ok 2 - hostname
ok 3 - process
ok 4 - mounts
ok 5 - user
ok 6 - rlimits
ok 7 - capabilities
ok 8 - default symlinks
ok 9 - default file system
ok 10 - default devices
ok 11 - linux devices
ok 12 - linux process
ok 13 - masked paths
ok 14 - oom score adj
ok 1 # SKIP syscall action SCMP_ACT_ALLOW
ok 2 # SKIP syscall action SCMP_ACT_ALLOW
ok 3 # SKIP syscall action SCMP_ACT_ALLOW
ok 4 # SKIP syscall action SCMP_ACT_ALLOW
ok 5 # SKIP syscall action SCMP_ACT_ALLOW
ok 6 # SKIP syscall action SCMP_ACT_ALLOW
ok 15 - seccomp
ok 16 - read only paths
ok 17 - rootfs propagation
ok 18 - sysctls
ok 19 - uid mappings
ok 20 - gid mappings
1..20
```

contrib/rootfs-builder/get-stage3.sh
```diff
@@ -39,9 +39,9 @@ fi
 
 BASE_ARCH_URL="${BASE_ARCH_URL:-${MIRROR}releases/${BASE_ARCH}/autobuilds/}"
 LATEST=$(wget -O - "${BASE_ARCH_URL}latest-stage3.txt")
-DATE=$(echo "${LATEST}" | sed -n "s|/stage3-${STAGE3_ARCH}-[0-9]*[.]tar[.]bz2.*||p")
-ARCH_URL="${ARCH_URL:-${BASE_ARCH_URL}${DATE}/}"
-STAGE3="${STAGE3:-stage3-${STAGE3_ARCH}-${DATE}.tar.bz2}"
+DATE=$(echo "${LATEST}" | sed -n "s|/.*stage3-${STAGE3_ARCH}-systemd-[0-9]*[.]tar[.]bz2.*||p" | tail -1)
```
wking: Thanks for pointing out the current breakage :).

I don't think we want to use the systemd image, since we're just grabbing the BusyBox binary.  It's [currently][1] quite a bit bigger:

```
20180308T214502Z/stage3-amd64-20180308T214502Z.tar.xz 188851072
…
20180303/systemd/stage3-amd64-systemd-20180303.tar.bz2 281873162
```

I'm in the process of updating the rootfs builder for the new format, but haven't gotten through the variable compression (xz, bz2, …) yet.  I'll file a PR once I've got it working again.
[1]: http://distfiles.gentoo.org/releases/amd64/autobuilds/latest-stage3.txt

contrib/rootfs-builder/Makefile
```diff
@@ -20,7 +20,7 @@ rootfs/%/bin/busybox: downloads/stage3-%-systemd-current.tar.bz2 rootfs-files
 
 .PRECIOUS: rootfs/%/bin/echo
 rootfs/%/bin/echo: rootfs/%/bin/busybox
-	sudo sh -c 'COMMANDS=$$($< --list) || exit 1; for COMMAND in $${COMMANDS}; do \
+	sudo sh -c 'COMMANDS=$$($< --list | grep -v "^busybox$$") || exit 1; for COMMAND in $${COMMANDS}; do \
```
wking: Why wxclude it?  It's a valid command in its own right.

alban: Because the Makefile will call `ln -rs busybox busybox` and it will fail:

```
sudo sh -c 'COMMANDS=$(rootfs/amd64/bin/busybox --list) || exit 1; for COMMAND in ${COMMANDS}; do \
	test -L "rootfs/amd64/bin/${COMMAND}" || ln -rs rootfs/amd64/bin/busybox "rootfs/amd64/bin/${COMMAND}" || exit; \
done'
ln: failed to create symbolic link 'rootfs/amd64/bin/busybox': File exists
```

The busybox command doesn't need a symlink.

contrib/rootfs-builder/get-stage3.sh
alban: Ok, I'll rebase my PR when yours get merged.

alban:
```
$ sudo validation/linux_uid_mappings.t
TAP version 13
not ok 1 - root filesystem
  ---
  {
    "error": "rootfs must not be readonly\nRefer to: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#root"
  }
```

This error is addressed separately in https://github.com/opencontainers/runtime-tools/pull/599

alban:
@wking I rebased this PR on master (with #599 and #598) and updated the commitmsg.

q384566678:
LGTM

PR #594, opened 2018-03-08
runtimetest: count correctly TAP tests

alban:
Don't create a new tap.T in the middle of a test but reuse the existing
one instead.

Symptoms:
```
$ sudo validation/linux_readonly_paths.t
TAP version 13
ok 1 - root filesystem
ok 2 - hostname
ok 3 - process
ok 4 - mounts
ok 5 - user
ok 6 - rlimits
ok 7 - capabilities
ok 8 - default symlinks
ok 9 - default file system
ok 10 - default devices
ok 11 - linux devices
ok 12 - linux process
ok 13 - masked paths
ok 14 - oom score adj
ok 1 # SKIP syscall action SCMP_ACT_ALLOW
ok 2 # SKIP syscall action SCMP_ACT_ALLOW
ok 3 # SKIP syscall action SCMP_ACT_ALLOW
ok 4 # SKIP syscall action SCMP_ACT_ALLOW
ok 5 # SKIP syscall action SCMP_ACT_ALLOW
ok 6 # SKIP syscall action SCMP_ACT_ALLOW
ok 15 - seccomp
ok 16 - read only paths
ok 17 - rootfs propagation
ok 18 - sysctls
ok 19 - uid mappings
ok 20 - gid mappings
1..20
```

After this patch is applied:
```
$ sudo validation/linux_readonly_paths.t
TAP version 13
ok 1 - root filesystem
ok 2 - hostname
ok 3 - process
ok 4 - mounts
ok 5 - user
ok 6 - rlimits
ok 7 - capabilities
ok 8 - default symlinks
ok 9 - default file system
ok 10 - default devices
ok 11 - linux devices
ok 12 - linux process
ok 13 - masked paths
ok 14 - oom score adj
ok 15 # SKIP syscall action SCMP_ACT_ALLOW
ok 16 # SKIP syscall action SCMP_ACT_ALLOW
ok 17 # SKIP syscall action SCMP_ACT_ALLOW
ok 18 # SKIP syscall action SCMP_ACT_ALLOW
ok 19 # SKIP syscall action SCMP_ACT_ALLOW
ok 20 # SKIP syscall action SCMP_ACT_ALLOW
ok 21 - seccomp
ok 22 - read only paths
ok 23 - rootfs propagation
ok 24 - sysctls
ok 25 - uid mappings
ok 26 - gid mappings
1..26
```

Fixes https://github.com/opencontainers/runtime-tools/issues/593

Signed-off-by: Alban Crequy <alban@kinvolk.io>

PR #599, opened 2018-03-09
runtimetest: fix root readonly check

alban:
Instead of trying to write a file on the rootfs, check the "ro"
per-mount option on the root mountpoint. The rootfs might not be
readable despite spec.Root.Readonly being false. Example: the rootfs
belong to an unmapped uid.


wking:
> The rootfs might not be readable despite spec.Root.Readonly being false. Example: the rootfs belong to an unmapped uid.

Checking [the spec][1] again, it doesn't actually say anything about how the runtime should handle the `readonly` `false` and unset cases.  So I'm in favor of something like:

```go
if spec.Root.Readonly {
  err := testWriteAccess("/")
  if err == nil {
    return specerror.NewError(specerror.RootReadonlyImplement, fmt.Errorf("rootfs must be readonly"), rspec.Version)
} else {
  err := testWriteAccess("/")
  if err != nil {
    // print a Skip-level warning, which may depend on #594 or other *tap.T passing
  }
}
```

That way we don't have to have an opinion on unspecified implementation details (perhaps the runtime doesn't bother with a `ro` mount if it knows the ID mapping will cover read-only-ness?).

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/config.md#L45

wking:
> I just skip the test on userns if `spec.Root.Readonly == false`...

I agree that the unmapped user-namespace situation should not be a spec violation.  But I don't see a need for special handling there, because I don't see spec grounds for [this error][1] at all.  Do you?

> I don't think I need to use `*tap.T` from #594 right now, it will just `t.Pass()` when err == nil. Do we actually have a "Skip-level warning"? Do you mean using "OPTIONAL"?

I don't think we can use OPTIONAL, or even MAY, unless we're quoting spec wording, and the spec has nothing in that line for unset/false `readonly`.  So I'd be ok with just:

```go
if spec.Root.Readonly {
   // what's in master now
} // no need to check the else case
```

But I also think that users may find this surprising, and a TAP skip or stderr log are places where we can warn the user "hey, you might have expected setting `readonly` false would give you a writeable root, but yours isn't writeable.  This isn't a spec volation, but we wanted to let you know anyway.".

[1]: https://github.com/opencontainers/runtime-tools/blob/cb96426c39451df19ef3f00ba024470e26171f4d/cmd/runtimetest/main.go#L334

alban:
> I agree that the unmapped user-namespace situation should not be a spec violation. But I don't see a need for special handling there, because I don't see spec grounds for this error at all. Do you?

No, I don't.

> I don't think we can use OPTIONAL, or even MAY, unless we're quoting spec wording, and the spec has nothing in that line for unset/false readonly. So I'd be ok with just:

Ok, I updated the patch that way. I also deleted validation/root_readonly_false.go since that would not test anything.

> But I also think that users may find this surprising, and a TAP skip or stderr log are places where we can warn the user "hey, you might have expected setting readonly false would give you a writeable root, but yours isn't writeable. This isn't a spec volation, but we wanted to let you know anyway.".

That would be good indeed. I didn't do that in this PR because that would need more changes for the TAP skip.

liangchenye:
LGTM

PR #601, opened 2018-03-13
Validate_linux: Modify the returned error

q384566678:

### DIFF
--- a/validate/validate_linux.go
+++ b/validate/validate_linux.go
@@ -180,7 +180,7 @@ func (v *Validator) CheckLinux() (errs error) {
 		}
 
 		if _, exists := devTypeList[devID]; exists {
-			logrus.Warnf("type:%s, major:%d and minor:%d for linux devices is duplicated", device.Type, device.Major, device.Minor)
+			logrus.Warnf("%v", specerror.NewError(specerror.DevicesErrorOnDup, fmt.Errorf("type:%s, major:%d and minor:%d for linux devices is duplicated", device.Type, device.Major, device.Minor), rspec.Version))
 		} else {
 			devTypeList[devID] = true
 		}


validate/validate_linux.go
```diff
@@ -180,7 +180,7 @@ func (v *Validator) CheckLinux() (errs error) {
 		}
 
 		if _, exists := devTypeList[devID]; exists {
-			logrus.Warnf("type:%s, major:%d and minor:%d for linux devices is duplicated", device.Type, device.Major, device.Minor)
+			logrus.Warnf("%v", specerror.NewError(specerror.DevicesErrorOnDup, fmt.Errorf("type:%s, major:%d and minor:%d for linux devices is duplicated", device.Type, device.Major, device.Minor), rspec.Version))
```
liangchenye: +1

validate/validate_linux.go
```diff
@@ -94,7 +94,7 @@ func (v *Validator) CheckLinux() (errs error) {
 	for index := 0; index < len(v.spec.Linux.Devices); index++ {
 		device := v.spec.Linux.Devices[index]
 		if !deviceValid(device) {
-			errs = multierror.Append(errs, fmt.Errorf("device %v is invalid", device))
+			errs = multierror.Append(errs, specerror.NewError(specerror.DevicesAvailable, fmt.Errorf("device %v is invalid", device), rspec.Version))
```
liangchenye: I think it is not 'DevicesAvailable' error, it is simply 'device is invalid'.
DevicesAvailable is done here: 'https://github.com/opencontainers/runtime-tools/blob/master/cmd/runtimetest/main.go#L437'

liangchenye: BTW, the 'deviceValid' function should move to validate_linux.go.
It might be in another PR.

validate/validate_linux.go
q384566678: updated, PTAL.

liangchenye:
LGTM

PR #602, opened 2018-03-14
validation: Implement ProcArgsApplyUntilStart

q384566678:

### DIFF
--- a/validation/start.go
+++ b/validation/start.go
@@ -53,6 +53,10 @@ func main() {
 	if err != nil {
 		util.Fatal(err)
 	}
+	_, err = os.Stat(output)
+	// check the existence of the output file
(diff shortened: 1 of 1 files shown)


validation/start.go
```diff
@@ -53,6 +53,10 @@ func main() {
 	if err != nil {
 		util.Fatal(err)
 	}
+	outputData, outputErr := ioutil.ReadFile(output)
+	// check the output
+	util.SpecErrorOK(t, outputErr != nil || string(outputData) != "process called\n", specerror.NewError(specerror.ProcArgsApplyUntilStart, fmt.Errorf("`process.args` MUST NOT be applied until triggered by the start operation"), rspecs.Version), outputErr)
```
liangchenye: If `outputErr is nil`, it means the process is applied.
In this case, when `string(outputData) != "process called\n"`, we could only get this conclusion:
```
the process is applied, but not running as expected.
```
It is different with ProcArgsApplyUntilStart.
I suggest we can just check the existence of the output file.

q384566678: updated, PTAL.

liangchenye:
LGTM

PR #604, opened 2018-03-14
validation: Implement ConfigUpdatesWithoutAffect

q384566678:

### DIFF
--- a/validation/config_updates_without_affect.go
+++ b/validation/config_updates_without_affect.go
@@ -1,0 +1,89 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+
+	tap "github.com/mndrix/tap-go"
+	"github.com/mrunalp/fileutils"
+	rspec "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/opencontainers/runtime-tools/generate"
+	"github.com/opencontainers/runtime-tools/specerror"
+	"github.com/opencontainers/runtime-tools/validation/util"
+	uuid "github.com/satori/go.uuid"
+)
+
+func main() {
+	t := tap.New()
+	t.Header(0)
+
+	bundleDir, err := util.PrepareBundle()
+	if err != nil {
+		util.Fatal(err)
+	}
+	defer os.RemoveAll(bundleDir)
+
+	r, err := util.NewRuntime(util.RuntimeCommand, bundleDir)
+	if err != nil {
+		util.Fatal(err)
+	}
+
+	testPath := filepath.Join(bundleDir, "test.json")
+	r.SetID(uuid.NewV4().String())
+	// generate a config has all the testing properties
+	g, err := util.GetDefaultGenerator()
+	if err != nil {
+		util.Fatal(err)
+	}
+	g.SetProcessArgs([]string{"/runtimetest", "--path=/test.json"})
+	g.AddLinuxMaskedPaths("/proc/kcore")
+	g.AddLinuxReadonlyPaths("/proc/fs")
+	g.AddLinuxSysctl("net.ipv4.ip_forward", "1")
+	g.SetProcessOOMScoreAdj(100)
+	g.AddProcessRlimits("RLIMIT_NOFILE", 1024, 1024)
+	g.SetLinuxRootPropagation("shared")
+
+	err = r.SetConfig(g)
+	if err != nil {
+		util.Fatal(err)
+	}
+
+	err = g.SaveToFile(testPath, generate.ExportOptions{})
+	if err != nil {
+		util.Fatal(err)
+	}
+
+	err = fileutils.CopyFile("runtimetest", filepath.Join(r.BundleDir, "runtimetest"))
+	if err != nil {
+		util.Fatal(err)
+	}
+
+	err = r.Create()
+	if err != nil {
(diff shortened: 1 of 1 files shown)


liangchenye:
It proves that the process change after the `create` operation will not have an effect on the container and 
It is a part of 'any change', so this PR looks good to me.

But it does not prove other property changes, like 'env', 'mount' and the etc.
I think it is not ready to mark this  ConfigUpdatesWithoutAffect spec item to 'DONE'.

To prove 'any change...', how about this:
1. create a config.json (configA) with 'runtimetest' as the process
2. save configA to a different dir for `runtimetest` to read
3. create a container
4. create another config.json(configB) and make any changes (except process)
5. start the container
     ---- if runtimetest detects any property is different to configA, we can get
            the conclusion:
5.1 runtime cannot apply properties correctly  --- if we verify all the properties, it will not happen.
5.2 runtime updates to new config after 'create' --- this is what we need

wking:
> create another config.json(configB) and make any changes (except process)

Why not change `process` too?  The [spec][1] has no such qualification.

I like @liangchenye's general approach though.  Can we adjust our test helper to *always* clobber `config.json` with a dummy config (maybe [this one][2]?) immediately after a successful `create`?

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/runtime.md#L111
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/schema/test/config/good/minimal.json

liangchenye:
> Why not change process too? The spec has no such qualification.
The process is already verified in this commit :)

wking:
> The process is already verified in this commit :)

Do we want to keep that?  `poststop.go` seems like a strange place for that check.  If we always clobber via our test helpers, almost all our tests would enforce this requirement.

liangchenye:
You are right. When we use `https://github.com/opencontainers/runtime-spec/blob/v1.0.1/schema/test/config/good/minimal.json`, if it fails to start, it means the 'process' is updated; if it succeeds to start but fails in checking any properties, it means the 'other properties' are updated(assuming those properties are already verified).

q384566678:
It looks like I need to write a new file for this test.

q384566678:
@liangchenye @wking updated, PTAL.

liangchenye:
```
Any updates to config.json after this step MUST NOT affect the container.
```
## inside test
For example, if we want to test if hostname updates will affect the container, we need to 
1. set the config.json (A) with `hostname == good` and with `process == runtimetest`
2. create a container
3. set the config.json (B) with `hostname == bad` 
4. start the container
    the container process (runtimetest) will verify the hostname in `validateHostname`.
    if os.Hostname() is equal to `good`, it means `the hostname updates DOES NOT affect the container`.
    Or we can output a spec error.

Now `bundledir/config.json` is both config.json for the runtime and the `expected properties` for `runtimetest`, we will lose the 'expected properties' if we doing the 3rd step. We can save a copy of (A) to `bundledir/test/config.json`  (C) and tell 'runtimetest' to check with this file.
If runtimetest does not detect a hostname difference, it means the update of 'hostname' DOES not affect the container.

It is the same approach when we check other updates. 
A `lazy` way is we can do it in RuntimeInsideValidate.

q384566678:
@liangchenye @wking I updated this pr, and I added a runtimetest to test the `configA` command in process.

But when I test it, it's going to quit when it's doing `r.Create()`, I'm not sure why.
```
TAP version 13
exit status 1
```

liangchenye:
@q384566678  I'm debugging it now.

liangchenye:
Seems you forget to copy 'runtimetest' to the bundle.

q384566678:
I try to put this test in RuntimeInsideValidate, but I found a problem, other tests are executed in call RuntimeInsideValidate, there is no guarantee that the config is correct, so when the start failure, cannot confirm is affected by the later config, was originally the config is wrong.
So I think it's better to put this test alone.
@liangchenye @wking PTAL.

liangchenye:
The current implementation (commit#4cf14f6) is clearer to figure out the real problem of a runtime.
I think if we can guarantee the original config (configA) has all the testing properties, commit#4cf14f6 for the inside test will be sufficient.

q384566678:
updated, I added properties to config.json.

q384566678:
@liangchenye PTAL

validation/config_updates_without_affect.go
```diff
@@ -0,0 +1,86 @@
...
+		util.Fatal(err)
+	}
+	defer os.RemoveAll(bundleDir)
+
+	r, err := util.NewRuntime(util.RuntimeCommand, bundleDir)
+	if err != nil {
+		util.Fatal(err)
+	}
+
+	testPath := filepath.Join("/", "test", "config.json")
```
wking: This rooted path fails when you use it for `g.SaveToFile(…)` below:

```console
$ ./validation/config_updates_without_affect.t 
TAP version 13
open /test/config.json: no such file or directory
```

I agree with @liangchenye that [you should set this path to something inside the bundle][1], at least for the host-mount-namespace `SaveToFile` call.

[1]: https://github.com/opencontainers/runtime-tools/pull/604#discussion_r175666390

liangchenye:
It works in my test and according to previous discussion, I'll merge it.
LGTM

PR #605, opened 2018-03-14
Fix error messages in validation cgroup tests

alban:
Symptoms:
```
 ok 3 - blkio weight is set correctly
 # expect: 500, actual: 842352267696
```

After the patch:
```
 ok 3 - blkio weight is set correctly
 # expect: 500, actual: 500
```

I found those issues by patching runc to introduce "off-by-one" bugs and
checking if they are correctly caught by the validation tests:
https://github.com/kinvolk/runc/commits/alban/off-by-one-bugs


liangchenye:
LGTM

PR #606, opened 2018-03-15
contrib/rootfs-builder: Use $(cat rootfs-files)

wking:
The previous

    $(< rootfs-files)

is a Bashism that's not part of [POSIX command substitution][1].  For Bash, the `$(< …)` form is slightly faster (no need to exec `cat`), but not everyone uses Bash.

@liangchenye, I bet this is what was [happening to you][2].

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_03
[2]: https://github.com/opencontainers/runtime-tools/pull/598#issuecomment-372991838


liangchenye:
It solves the problem.

But I'm still not sure why the original script failed.
If I run 'tar .... $(< rootfs-files)' manually in my terminal, it works.
But if I run 'make ...', it fails.

liangchenye:
LGTM

PR #607, opened 2018-03-16
add test case for KillNonCreateRunHaveNoEffect

liangchenye:

### DIFF
--- a/validation/kill_no_effect.go
+++ b/validation/kill_no_effect.go
@@ -1,0 +1,62 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"reflect"
+	"time"
+
+	"github.com/mndrix/tap-go"
+	rspecs "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/opencontainers/runtime-tools/specerror"
+	"github.com/opencontainers/runtime-tools/validation/util"
+	uuid "github.com/satori/go.uuid"
+)
+
+func main() {
+	t := tap.New()
+	t.Header(0)
+	bundleDir, err := util.PrepareBundle()
+	if err != nil {
+		util.Fatal(err)
+	}
+	defer os.RemoveAll(bundleDir)
+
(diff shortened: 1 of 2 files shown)


liangchenye:
Currently we have 4 status: creating/created/running/stopped. Since  'created/running' is out of this testing and it is hard to catch 'creating' status, we can only test the 'stopped' container.

If there were any 'effect' when kill a 'stopped' container, the state might change. I don't have an idea on other 'effect' so in this PR I check the state changes.

q384566678:
looks good. @wking WDYT?

validation/kill_no_effect.go
```diff
@@ -0,0 +1,62 @@
...
+		PreDelete: func(r *util.Runtime) error {
+			err := util.WaitingForStatus(*r, util.LifecycleStatusStopped, time.Second*5, time.Second*1)
+			if err != nil {
+				return err
+			}
+			currentState, err := r.State()
+			if err != nil {
+				return err
+			}
+			r.Kill("KILL")
```
wking: You probably want to ensure that this call returns an error, based on [this][1] and [this][2].

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/runtime.md#L125
[2]: https://github.com/opencontainers/runtime-tools/blame/v0.5.0/docs/command-line-interface.md#L211

liangchenye:
```
* *Exit code:* Zero if the signal was successfully sent to the container process and non-zero on errors.
```
I found there is no general exit code,  there is golang syscall pkg, but just works in posix system.
So do we still need this 'Exit code' session in command-line-interface?
```
if status, ok := exiterr.Sys().(syscall.WaitStatus); ok {
                log.Printf("Exit Status: %d", status.ExitStatus())
}
```

liangchenye: The return error is check in `kill.go`. There is a bug in the previous kill.go, it displays the wrong spec error code. The new commit 0ddb5cd solves that.

q384566678:
LGTM

PR #608, opened 2018-03-19
check the status of the state passed to hooks over stdin

liangchenye:

### DIFF
--- a/validation/hooks_stdin.go
+++ b/validation/hooks_stdin.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -9,6 +10,7 @@ import (
 	"reflect"
 	"time"
 
+	multierror "github.com/hashicorp/go-multierror"
 	tap "github.com/mndrix/tap-go"
 	rspecs "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/runtime-tools/specerror"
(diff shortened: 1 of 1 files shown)


liangchenye:
It is a successor PR of https://github.com/opencontainers/runtime-tools/pull/589.
It only checks prestart status and poststart status.

The poststop status is not meaningful since the container will be deleted and the id might be reused.
I skip the poststop check.

validation/hooks_stdin.go
```diff
@@ -42,6 +42,22 @@ func stdinStateCheck(outputDir, hookName string, expectedState rspecs.State) err
...
+	case "prestart":
+		if state.Status != "created" {
+			return fmt.Errorf("wrong state %q in the stdin of %s hook, expected %q", state.Status, hookName, "created")
+		}
+	case "poststart":
+		if state.Status != "running" {
+			return fmt.Errorf("wrong state %q in the stdin of %s hook, expected %q", state.Status, hookName, "running")
+		}
+	case "poststop":
+		// Not defined by runtime spec
```
wking: I think the current spec wants this to be `stopped` (more discussion in opencontainers/runtime-spec#958), although more spec clarity would be nice.  But what the spec is clear about is that [`status` must be set to *something*][1].

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/runtime.md#L16

liangchenye: I'll update the code since `'status' must be set to something code.` and wait for runtime-spec#958.

liangchenye:
Now we only output three 'OK or NOT' results, each result may have the detailed property errors information. I  join the state property errors of a hook together by using multiple errors.

dongsupark:
@liangchenye 
I happened to see this PR, as it's exactly what @alban and I were going to implement.
Though with this PR, the `hooks_stdin.t` test fails like it:
```
not ok 1 - the state of the container MUST be passed to "prestart" hook over stdin
  ---
  {
    "error": "1 error occurred:\n\n* wrong status \"\" in the stdin of prestart hook, expected \"created\"",
    "reference": "https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#posix-platform-hooks"
  }
  ...
```
It looks like `Status` is always being set to an empty string.

dongsupark:
@wking Thanks for the info.
I'll try to test it with the runc fix later. Anyway I'm fine with merging this PR without waiting for runc.

q384566678:
looks good, But when I run the test, I get the error.
@liangchenye Did you test successfully?

```
➜  runtime-tools git:(7c6996f) ✗ sudo ./hooks_stdin
TAP version 13
not ok 1 - the state of the container MUST be passed to "prestart" hook over stdin
  ---
  {
    "error": "1 error occurred:\n\n* wrong status \"\" in the stdin of prestart hook, expected \"created\"",
    "reference": "https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#posix-platform-hooks"
  }
  ...
not ok 2 - the state of the container MUST be passed to "poststart" hook over stdin
  ---
  {
    "error": "1 error occurred:\n\n* wrong status \"\" in the stdin of poststart hook, expected \"running\"",
    "reference": "https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#posix-platform-hooks"
  }
  ...
not ok 3 - the state of the container MUST be passed to "poststop" hook over stdin
  ---
  {                             
    "error": "1 error occurred:\n\n* status in the stdin of poststop hook should not be empty",
    "reference": "https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#posix-platform-hooks"
  }
  ...
1..3
```

liangchenye:
@q384566678  Yes, it does not work because of the [runc bug](https://github.com/opencontainers/runtime-tools/pull/608)

PR #612, opened 2018-03-21
validation: add linux_process validation

q384566678:


validation/linux_process.go
```diff
@@ -0,0 +1,15 @@
+package main
+
+import (
+	"github.com/opencontainers/runtime-tools/validation/util"
+)
+
+func main() {
+	g := util.GetDefaultGenerator()
+	g.SetProcessNoNewPrivileges(true)
```
wking: It looks like this is the only thing specific to this test.  If so, maybe it should be `process_nonewprivileges.go` with a check to `t.Skip` if `runtime.GOOS != "linux"`?  If the goal is to exercise all of the Linux-specific `process` properties in one file, you also need to touch [`apparmorProfile`, etc][1].

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/config.md#linux-process

q384566678: I prefer to integrate them into one file. updated.

wking: > I prefer to integrate them into one file. updated.

You still aren't testing `apparmorProfile` or `selinuxLabel`.  That's technically ok, because the spec places no runtime conditions on them, but we may want to set them here anyway and leave it to `runtimetest` to `Skip` warnings for any surprising (but still technically conformant) runtime behavior).

q384566678: updated, @wking @liangchenye PTAL.

alban:
When testing this branch, I get the following:
```
$ sudo validation/linux_process/linux_process.t
selinux label is specified in config, but selinux is disabled or not supported
exit status 1
```
(Tested on Fedora with SELinux disabled)

Could we have `SetProcessSelinuxLabel` in a separate test, so that we could test other basic operations (`SetupPrivileges`, `SetProcessNoNewPrivileges`) separately?

We should also have a way to skip SELinux tests when it is not available (in the same way we have Windows-specific or Linux-specific tests).

PR #613, opened 2018-03-22
[WIP] validation: add test for NSProcInPath

alban:
I initially tried to add the checks in the container process
'runtimetest' by adding annotations prefixed with "runtimetest/".  But
that proved impractical with TAP outputs because I wanted to have
several tests for each namespace.

This patch now validates the namespaces outside the container with
util.RuntimeOutsideValidate().

TODO/notes:
- [ ] Fix the synchronisation with "sleep".
- [ ] Is it ok to have a dependency on 'util-linux'?
- [ ] Is it acceptable to require Linux >= 4.12? I guess not, so I would need to implement that differently.
- [ ] Unclear spec about user namespaces: https://github.com/opencontainers/runtime-spec/issues/961
- [ ] Fix about cgroupns in runc: https://github.com/opencontainers/runc/pull/1763

### DIFF
--- a/validation/linux_ns_path.go
+++ b/validation/linux_ns_path.go
@@ -12,6 +12,21 @@ import (
 	"github.com/opencontainers/runtime-tools/validation/util"
 )
 
+func getRuntimeToolsNamespace(ns string) string {
+	// Deal with exceptional cases of "net" and "mnt", because those strings
+	// cannot be recognized by mapStrToNamespace(), which actually expects
+	// "network" and "mount" respectively.
+	switch ns {
+	case "net":
+		return "network"
+	case "mnt":
+		return "mount"
+	}
+
+	// In other cases, return just the original string
+	return ns
+}
+
 func testNamespacePath(t *tap.T, ns string, unshareOpt string) error {
 	// Calling 'unshare' (part of util-linux) is easier than doing it from
 	// Golang: mnt namespaces cannot be unshared from multithreaded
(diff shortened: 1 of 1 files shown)


validation/linux_ns_path.go
```diff
@@ -0,0 +1,121 @@
...
+	// Golang: mnt namespaces cannot be unshared from multithreaded
+	// programs.
+	cmd := exec.Command("unshare", unshareOpt, "--fork", "sleep", "10000")
+	err := cmd.Start()
+	if err != nil {
+		return fmt.Errorf("cannot run unshare: %s", err)
+	}
+	defer func() {
+		if cmd.Process != nil {
+			cmd.Process.Kill()
```
alban: The sleep processes are surviving the death of the `unshare` parent unfortunately.

I tried `cmd.Process.Signal(syscall.SIGTERM)` but it does not help because:
- unshare does not catch the signal so it dies without forwarding the SIGTERM to its children
- processes in a different pid namespaces would not receive the signal anyway because of signal rules in pid namespaces (see [man 7 pid_namespaces](http://man7.org/linux/man-pages/man7/pid_namespaces.7.html))

dongsupark:
I created a new version of PR https://github.com/opencontainers/runtime-tools/pull/628, based on this PR #613.

For the record, I'll go through each TODO item.

* I think `Fix sync with sleep` and `stale child processes` are already resolved.
* `Is it ok to have a dependency on util-linux?`: I think yes, as util-linux has been a de facto standard library since many years.
* `/proc/pid/ns/pid_for_children` for kernel >= 4.12: we don't need to care about it, because it's solved by using setpgid.
* `unclear spec about user namespaces`: I'm not sure if there's anything left to do
* `Fix about cgroupns in runc`: This is blocked by the ancient PR https://github.com/opencontainers/runc/pull/1184. Nothing to do right now.

alban:
Superseded by #628.

PR #618, opened 2018-04-05
runtimetest: fix a validation test for unbindable mount

dongsupark:
`validation/linux_rootfs_propagation_unbindable.t` fails with the following message:

```
not ok 17 - rootfs propagation
  ---
  {
    "error": "rootfs expected to be unbindable, but not"
  }
```

That's because mount flag is set to `MS_BIND`. This should be actually `MS_UNBINDABLE`.

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -383,7 +383,7 @@ func validateRootfsPropagation(spec *rspec.Spec, t *tap.T) error {
 		}
 		return fmt.Errorf("rootfs should be %s, but not", spec.Linux.RootfsPropagation)
 	case "unbindable":
-		if err := unix.Mount("/", targetDir, "", unix.MS_BIND|unix.MS_REC, ""); err != nil {
+		if err := unix.Mount("/", targetDir, "", unix.MS_UNBINDABLE|unix.MS_REC, ""); err != nil {
 			if err == syscall.EINVAL {
 				return nil
 			}


alban:
@dongsupark I don't think that's correct: the test program `runtimetest` runs inside the container and it is not supposed to create an unbindable mount but to check that the rootfs has been setup by runc with the unbindable flag as specified in the config.json. It does not check by looking at the flags in `/proc/self/mountinfo` but by checking it has the correct effect, by trying to do a new bind mount and checking that the kernel successfully returns an error as it ought to.

runc has a issue with rootfs propagation mode with 'shared', so maybe the issue happens with the the 'unbindable' propagation mode too: https://github.com/opencontainers/runc/issues/1755

To test runc manually, you can run 
```
oci-runtime-tool generate --linux-rootfs-propagation=unbindable --process-terminal=true --rootfs-path=/home/alban/distro-trees/f26/ > config.json
sudo runc run mycontainer
cat /proc/self/mountinfo | grep unbindable
```

The flag MS_BIND is for creating a bind mount and requires two paths:
```
mount --bind /source /target
```

The flag MS_UNBINDABLE is for changing an existing mount propagation mode, to make it unbindable, and it requires 1 path:
```
mount --make-unbindable /target
```

In your patch, you use MS_UNBINDABLE but with 2 paths, so I am not sure what the kernel is supposed to do in that case.

dongsupark:
Thanks for the clarification, I'll look into runc.

PR #624, opened 2018-04-16
validation: Complete process test

q384566678:
I try to set `terminal` to true. But it failed when testing.

```
cannot allocate tty if runc will detach without setting console socket
exit status 1
```

### DIFF
--- a/validation/process.go
+++ b/validation/process.go
@@ -12,6 +12,8 @@ func main() {
 	if err != nil {
 		util.Fatal(err)
 	}
+
+	g.SetProcessConsoleSize(10, 80)
 	g.SetProcessCwd("/test")
 	g.AddProcessEnv("testa", "valuea")
 	g.AddProcessEnv("testb", "123")


validation/process.go
```diff
@@ -12,6 +12,8 @@ func main() {
 	if err != nil {
 		util.Fatal(err)
 	}
+
+	g.SetProcessConsoleSize(10, 80)
```
wking: The spec [says][1]:

> Runtimes MUST ignore `consoleSize` if `terminal` is `false` or unset. 

which makes sense.  We currently aren't testing that in `runtimetest` (and for `validation/process.go` the standard streams shouldn't be terminals anyway.  So I don't have a problem setting `consoleSize` here, but if we actually want to *test* something by doing that we need to:

* Setup our own pseudoterminal pair here with a *different* size and pass the slave in to `create` as stdin (at least).
* Add tests to `runtimetest` to check that that, when `consoleSize` is false/unset, the actual console size does not match the `config.json` value.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/config.md#process

PR #625, opened 2018-04-18
runtimetest: correctly check for a readable directory

dongsupark:
So far tests linux_masked_paths.t & linux_readonly_paths.t have failed with error messages like `cannot test read access for "/dirname"`. That's because `testReadAccess()` only checked if the given path is a regular file, returning an error for every other case. `testReadAccess()` should actually take care of another case of a given
path being a directory, to be able to correctly check for its readability.

Also run `testFileReadAccess()` or `testFileWriteAccess()` for all file types, not only a normal file, because the runtime spec does not mandate the type of masked files. It could be actually any file, including a character device like `/dev/null`, especially in case of runc.

Found by @alban.

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -36,6 +36,10 @@ var gitCommit = ""
 // VERSION file of the source code.
 var version = ""
 
+// errAccess will be used for defining either a read error or a write error
+// from any type of files.
+var errAccess error
+
 // PrGetNoNewPrivs isn't exposed in Golang so we define it ourselves copying the value from
 // the kernel
 const PrGetNoNewPrivs = 39
@@ -410,22 +414,34 @@ func testReadAccess(path string) (readable bool, err error) {
 		return false, err
 	}
 
-	// In case of a directory, we should check its readability in a special way.
-	// In other files, we should not check its Mode explicitly, because the runtime
-	// spec does not mandate the type of masked files. It could be a regular file
-	// or a character file (/dev/null), which is the case for runtimes like runc.
-	if fi.IsDir() {
+	// Check for readability in case of regular files, character device, or
+	// directory. Although the runtime spec does not mandate the type of
+	// masked files, we should check its Mode explicitly. A masked file
+	// could be represented as a character file (/dev/null), which is the
+	// case for runtimes like runc.
+	switch fi.Mode() & os.ModeType {
+	case 0, os.ModeDevice | os.ModeCharDevice:
+		return testFileReadAccess(path)
+	case os.ModeDir:
 		return testDirectoryReadAccess(path)
 	}
-	return testFileReadAccess(path)
+
+	errAccess = fmt.Errorf("cannot test read access for %q (mode %d)", path, fi.Mode())
+	return false, errAccess
 }
 
 func testDirectoryReadAccess(path string) (readable bool, err error) {
-	if files, err := ioutil.ReadDir(path); err != nil || len(files) == 0 {
-		// err from reading from a directory should not be considered as test failure,
-		// it just means that the test program successfully assessed that
-		// the directory is not readable.
+	files, err := ioutil.ReadDir(path)
+	if err == io.EOF || len(files) == 0 {
+		// Our validation/ tests only use non-empty directories for read-access
+		// tests. So if we get an EOF on the first read, the runtime did
+		// successfully block readability. So it should not be considered as test
+		// failure, it just means that the test program successfully assessed
+		// that the directory is not readable.
 		return false, nil
+	}
+	if err != nil {
+		return false, err
 	}
 	return true, nil
 }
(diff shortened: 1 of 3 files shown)


alban:
I think this needs 2 changes:
- in `validation/linux_masked_paths.go`: create a file in `testDir`, so that the directory is not empty
- in `cmd/runtimetest/main.go::testDirectoryReadAccess()`: don't ignore the return value of `ioutil.ReadDir()` with `_` and return false when `ioutil.ReadDir()` gives an empty array.

The rationale is the same as explained in the comment of `testFileReadAccess()` (see also [this discussion](https://github.com/opencontainers/runtime-tools/pull/308#discussion_r179779113)).

dongsupark:
@alban Done.
Created a file under `testDir`, and added a check for an empty array.

alban:
@dongsupark `linux_masked_paths.t` now works fine with runc but the test for `linux_masked_paths.t` became "skipped":
```
$ sudo validation/linux_masked_paths.t
...
ok 284 # SKIP linux.readonlyPaths not set
```

This is because `testDir` does not contain any files. Could you create a file in `linux_readonly_paths.go` in the same way you did in `linux_masked_paths.go`?

alban:
I didn't test correctly before... It works for me now:
```
$ sudo validation/linux_masked_paths.t|grep -E 'mask|read'
ok 275 - cannot read masked path "/masked-dir"
ok 276 - cannot read masked path "/masked-file"
ok 284 # SKIP linux.readonlyPaths not set
$ sudo validation/linux_readonly_paths.t|grep -E 'mask|read'
ok 275 # SKIP linux.maskedPaths not set
ok 283 - "/readonly-dir" (linux.readonlyPaths[0]) is not writable
ok 284 - "/readonly-file" (linux.readonlyPaths[1]) is not writable
```

cmd/runtimetest/main.go
```diff
@@ -409,10 +409,25 @@ func testReadAccess(path string) (readable bool, err error) {
 	if err != nil {
 		return false, err
 	}
-	if fi.Mode()&os.ModeType == 0 {
-		return testFileReadAccess(path)
+
+	// In case of a directory, we should check its readability in a special way.
+	// In other files, we should not check its Mode explicitly, because the runtime
+	// spec does not mandate the type of masked files. It could be a regular file
+	// or a character file (/dev/null), which is the case for runtimes like runc.
```
wking: I'd rather continue to test the mode.  For example, the runtime may use a permission change to make a symlink read-only (or not read/writable at all), and we don't want to use `testFileReadAccess` to test symlink readability.  I'd rather support `/dev/null` and other character devices with logic like:

```go
switch fi.Mode&os.ModeType {
case 0, os.ModeDevice|os.ModeCharDevice:
    return testFileReadAccess(path)
case os.ModeDir:
    return restDirectoryReadAccess(path)
}
return false, fmt.Errorf("cannot test read access for %q (mode %d)", path, fi.Mode())
```

cmd/runtimetest/main.go
```diff
@@ -439,12 +454,15 @@ func testWriteAccess(path string) (writable bool, err error) {
 	if err != nil {
 		return false, err
 	}
+
+	// In case of a directory, we should check its readability in a special way.
+	// In other files, we should not check its Mode explicitly, because the runtime
+	// spec does not mandate the type of masked files. It could be a regular file
+	// or a character file (/dev/null), which is the case for runtimes like runc.
```
wking: I prefer [keeping the explicit mode checks][1] to avoid calling `testFileWriteAccess` on types where its logic doesn't make sense.  And if runtimes seek to block write access by binding `/dev/null` over the path in question, we'll need a more sophisticated check than [what `testFileWriteAccess` does now][2], because you can certainly write a byte to `/dev/null`.

[1]: https://github.com/opencontainers/runtime-tools/pull/625#discussion_r182495793
[2]: https://github.com/opencontainers/runtime-tools/blob/9185c46fb7f02a165b0dac8379caab8261b4a74c/cmd/runtimetest/main.go#L459-L464

dongsupark:
@wking I pushed a commit to address your review comments as much as possible. I hope it looks ok to you.

cmd/runtimetest/main.go
dongsupark: @wking I'm not sure I understand. Can you please elaborate how we can do a more sophisticated check for writeability of `/dev/null`?
I don't see a general way to check if a character device can be written. AFAIK the read/write behavior varies a lot for each character device. The best we can do is to check for the file mode, isn't it?

validation/linux_readonly_paths.go
```diff
@@ -21,6 +21,12 @@ func main() {
 		if err != nil {
 			return err
 		}
+		// create a temp file to make testDir non-empty
+		tmpfile, err := ioutil.TempFile(testDir, "tmp")
+		if err != nil {
+			return err
+		}
+		defer tmpfile.Close()
```
q384566678: I didn't understand why do that. I tested it as if it had no effect on the results.

dongsupark: @q384566678 You're right.
What I wanted to do was `defer os.Remove(tmpfile.Name())`. Fixed it.

q384566678:
LGTM

@dongsupark Merge it first, and then you can rebase #644. Maybe we can also change `masked-path` test.

q384566678:
>  Maybe we can also change masked-path test.

Sorry, I didn't notice that you had already done it.

dongsupark:
@q384566678 Done, rebased #643 as well as #644. :-)

PR #626, opened 2018-04-18
validation: create: don't skip errors on state

alban:
If the runtime fails to give us the state, don't just ignore this and
skip the test but ensure we print "not ok" in the TAP output.

### DIFF
--- a/validation/create.go
+++ b/validation/create.go
@@ -70,14 +70,13 @@ func main() {
 
 		if err == nil {
 			state, err := r.State()
+			t.Ok(err == nil && state.ID == c.id, "'state' MUST return the state of a container")
 			if err == nil {
-				t.Ok(state.ID == c.id, "'state' MUST return the state of a container")
 				t.YAML(map[string]string{
 					"container ID": c.id,
 					"state ID":     state.ID,
 				})
 			} else {
-				t.Skip(1, "'state' MUST return the state of a container")
 				diagnostic = map[string]string{
 					"error": err.Error(),
 				}


validation/create.go
```diff
@@ -70,14 +70,13 @@ func main() {
...
 			state, err := r.State()
+			t.Ok(err == nil && state.ID == c.id, "'state' MUST return the state of a container")
 			if err == nil {
-				t.Ok(state.ID == c.id, "'state' MUST return the state of a container")
 				t.YAML(map[string]string{
 					"container ID": c.id,
 					"state ID":     state.ID,
 				})
 			} else {
-				t.Skip(1, "'state' MUST return the state of a container")
```
wking: This effectively rolls back part of d8d23968 (#564), but I didn't motivate the skip then, and can't remember why I'd felt it was a good idea.  I'm fine returning to to stricter handling like you've done here.

validation/create.go
```diff
@@ -70,14 +70,13 @@ func main() {
 
 		if err == nil {
 			state, err := r.State()
+			t.Ok(err == nil && state.ID == c.id, "'state' MUST return the state of a container")
```
wking: While we're touching this line, we may want to use the appropriate spec error.  In this case, a failed `state` call would be violating the combination of [`EntityOperSameContainer`][1], [`QueryStateImplement`][2], and, from the CLI spec, “[The runtime MUST support the entire API defined in this specification][3]” and “[Exit code: Zero if the state was successfully written to stdout and non-zero on errors][4]” (which should really have a MUST, I'll file a PR for that).  From among those, the most important seems like `QueryStateImplement`.  Following [this pattern][5] would give us something like:

```go
err = specerror.NewError(specerror.QueryStateImplement, errors.New("'state' returns the state of the container"), g.Config.Version)
runtimeError, ok := err.(*specerror.Error)
if !ok {
  return util.Fatal(fmt.Errorf("cannot convert %v to a runtime-spec error", err))
}
state, err := r.State()
t.Ok(err == nil && state.ID == c.id, runtimeError.Err.Err.Error())
diagnostic = map[string]string{
  "level": runtimeError.Err.Level.String(),
  "reference": runtimeError.Err.Refernce,
}
if err == nil {
  diagnostic["container ID"] = c.id
  diagnostic["state ID"] = state.ID
} else {
  diagnostic["error"] = err
  if e, ok := err.(*exec.ExitError); ok && len(e.Stderr) > 0 {
    diagnostic["stderr"] = string(e.Stderr)
  }
}
t.YAML(diagnostic)
```

Although we could make that a bit easier by adding a helper like `func NewRFCError(code Code, err error, version string) rfc2119.Error` to `specerror`.  I've filed #627 to add `NewRFCError`.

[1]: https://github.com/opencontainers/runtime-tools/blob/9185c46fb7f02a165b0dac8379caab8261b4a74c/specerror/runtime.go#L11-L12
[2]: https://github.com/opencontainers/runtime-tools/blob/9185c46fb7f02a165b0dac8379caab8261b4a74c/specerror/runtime.go#L53-L54
[3]: https://github.com/opencontainers/runtime-tools/blob/9185c46fb7f02a165b0dac8379caab8261b4a74c/docs/command-line-interface.md#global-usage
[4]: https://github.com/opencontainers/runtime-tools/blob/9185c46fb7f02a165b0dac8379caab8261b4a74c/docs/command-line-interface.md#state
[5]: https://github.com/opencontainers/runtime-tools/blob/9185c46fb7f02a165b0dac8379caab8261b4a74c/cmd/runtimetest/main.go#L361-L363

liangchenye:
LGTM

PR #627, opened 2018-04-18
specerror: Add NewRFCError and NewRFCErrorOrPanic

wking:
So we don't have to dance around with casting to extract structured information from `NewError`'s returned error.


alban:
LGTM (but I'm not a maintainer)

specerror/error.go
```diff
@@ -61,6 +61,34 @@ func (err *Error) Error() string {
...
+		Level:     template.Level,
+		Reference: reference,
+		Err:       err,
+	}, nil
+}
+
+// NewRFCErrorOrPanic creates an rfc2119.Error referencing a spec
+// violation and panics on failure.  This is handy for situations
+// where you can't be bothered to check NewRFCError for failure.
+func NewRFCErrorOrPanic(code Code, err error, version string) (*rfc2119.Error) {
```
liangchenye: go fmt fails at this line.
It should be
```
func NewRFCErrorOrPanic(code Code, err error, version string)  *rfc2119.Error {
```

validation/state.go
```diff
@@ -7,6 +7,7 @@ import (
 	"github.com/mndrix/tap-go"
 	rspecs "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/runtime-tools/specerror"
+	rfc2119 "github.com/opencontainers/runtime-tools/error"
```
liangchenye: go fmt fails at this line.
It should be moved one line up.
```
         rfc2119 "github.com/opencontainers/runtime-tools/error"
 	"github.com/opencontainers/runtime-tools/specerror"
```

wking:
I've fixed the`.gofmt` errors with dac7b67 → 7992f01.

q384566678:
LGTM

PR #628, opened 2018-04-23
validation: add test for NSProcInPath

dongsupark:
This PR validates `NSProcInPath`, i.e. `The runtime MUST place the container process in the namespace associated with that path`. It checks that Linux namespaces are created with a given path by making use of `util.RuntimeOutsideValidate()`. 

Since the [previous version of PR](https://github.com/opencontainers/runtime-tools/pull/613), we had to fix the following things:

* Fix a bug caused by namespace string mismatches, found by @alban.
* Kill the unshare process as well as child processes at once, making use of `setpgid()`.
* Use a ticker loop instead of a pure sleep call for general sync mechanism.
* Print out correct diagnostics based on specError.

This PR replaces https://github.com/opencontainers/runtime-tools/pull/613.
See also https://github.com/opencontainers/runtime-tools/issues/572.

### DIFF
--- a/validation/linux_ns_path.go
+++ b/validation/linux_ns_path.go
@@ -12,6 +12,21 @@ import (
 	"github.com/opencontainers/runtime-tools/validation/util"
 )
 
+func getRuntimeToolsNamespace(ns string) string {
+	// Deal with exceptional cases of "net" and "mnt", because those strings
+	// cannot be recognized by mapStrToNamespace(), which actually expects
+	// "network" and "mount" respectively.
+	switch ns {
+	case "net":
+		return "network"
+	case "mnt":
+		return "mount"
+	}
+
+	// In other cases, return just the original string
+	return ns
+}
+
 func testNamespacePath(t *tap.T, ns string, unshareOpt string) error {
 	// Calling 'unshare' (part of util-linux) is easier than doing it from
 	// Golang: mnt namespaces cannot be unshared from multithreaded
--- a/validation/linux_ns_path.go
+++ b/validation/linux_ns_path.go
@@ -107,15 +107,6 @@ func checkNamespacePath(unsharePid int, ns string) error {
 	rtns := getRuntimeToolsNamespace(ns)
 	g.AddOrReplaceLinuxNamespace(rtns, unshareNsPath)
 
-	// The spec is not clear about userns mappings when reusing an
-	// existing userns. Anyway in reality, we should set up uid/gid
-	// mappings, to make userns work in most runtimes.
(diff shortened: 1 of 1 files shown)


q384566678:
It failed when I tested:
```
➜  runtime-tools git:(dongsupark/NSProcInPath) ✗ sudo ./linux_ns_path
TAP version 13
namespace {"cgroup" "/proc/27701/ns/cgroup"} does not exist
not ok 1 - set cgroup namespace by path
  ---
  {
    "actual": "err == exit status 1",
    "expected": "err == nil",
    "level": "MUST",
    "namespace type": "cgroup",
    "reference": "https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config-linux.md#namespaces"
  }
  ...
ok 2 - set ipc namespace by path
ok 3 - set mnt namespace by path
ok 4 - set net namespace by path
ok 5 - set pid namespace by path
User namespaces enabled, but no uid mappings found.
not ok 6 - set user namespace by path
  ---
  {
    "actual": "err == exit status 1",
    "expected": "err == nil",
    "level": "MUST",
    "namespace type": "user",
    "reference": "https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config-linux.md#namespaces"
  }
  ...
ok 7 - set uts namespace by path
1..7
```
For `user namespace` errors, I think that `uid` should be set, but I don't know why the `cgroups namespace` are wrong.

dongsupark:
@q384566678 Yeah, I'm aware of both failures.
Userns test fails because [UID/GID mappings are currently disabled](https://github.com/opencontainers/runtime-tools/pull/628/commits/8e6baa6231e7173b113595a4e4c91567b1a75161#diff-eb81adf2b36d762bb4489c9de5c6c834R69), due to unclear runtime-specs. According to the discussion in https://github.com/opencontainers/runtime-spec/issues/961, I suppose we can enable UID/GID mappings in this test, right?

Cgroupns test fails because runc has never supported cgroup namespaces. See the pending PR https://github.com/opencontainers/runc/pull/1184.

q384566678:
>According to the discussion in opencontainers/runtime-spec#961, I suppose we can enable UID/GID mappings in this test, right?

I think so.

>Cgroupns test fails because runc has never supported cgroup namespaces. See the pending PR opencontainers/runc#1184.

I think we can add some explanations to the cgroups to facilitate understanding.

dongsupark:
@q384566678 
Pushed with a new commit for enabling UID/GID mappings, as well as adding comments about cgroup namespaces.
Though userns test still fails. I'm not sure why. Will look into it later again.

q384566678:
LGTM

PR #629, opened 2018-04-24
cgroups_v1: Correction parameters

q384566678:
Instead of specifying the value, use the passed in parameter. (although they have the same value)


liangchenye:
+1 for this.
I think we should remove  RelCgroupPath in cgroups.go and add RelCgroupTestPath to validation/cgroup.go. This is just a testing variable.

dongsupark:
+1. Looks good.

liangchenye:
LGTM

PR #630, opened 2018-04-24
travis: fix fetch issue of golint

dongsupark:
As golint repo does not return metadata for `go get` to work correctly, every travis build fails. Let's install golint without using go get until it's fixed.

See also https://github.com/golang/lint/issues/397


q384566678:
LGTM

PR #636, opened 2018-05-15
validation: add a new test for NSPathMatchTypeError

dongsupark:
`checkNSPathMatchType` checks if the container returns an error when deliberately setting a wrong namespace type. Doing that, it is possible to verify `NSPathMatchTypeError`, i.e. `The runtime MUST generate an error if path is not associated with a namespace of type "type"`.

See also https://github.com/opencontainers/runtime-tools/issues/572
/cc @alban 

### DIFF
--- a/validation/linux_ns_path_type.go
+++ b/validation/linux_ns_path_type.go
@@ -1,0 +1,120 @@
+package main
+
+import (
+	"fmt"
+	"os/exec"
+	"runtime"
+	"syscall"
+
+	"github.com/mndrix/tap-go"
+	rspec "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/opencontainers/runtime-tools/specerror"
+	"github.com/opencontainers/runtime-tools/validation/util"
+)
+
+func getRuntimeToolsNamespace(ns string) string {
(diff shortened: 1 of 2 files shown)


validation/linux_ns_path_type.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		return fmt.Errorf("cannot get default config from generator: %v", err)
+	}
+
+	rtns := getRuntimeToolsNamespace(ns)
+	g.AddOrReplaceLinuxNamespace(rtns, unshareNsPath)
+
+	err = util.RuntimeOutsideValidate(g, func(config *rspec.Spec, state *rspec.State) error {
+		return nil
+	})
```
q384566678: I think this will be better: `err = util.RuntimeOutsideValidate(g, nil)`

dongsupark: @q384566678 Right. I've changed it as you suggested.

dongsupark:
Rebased, and added another commit to use an existing helper function `util.GetRuntimeToolsNamespaces()`.

q384566678:
LGTM

PR #637, opened 2018-05-16
validation: test cgroups with different input values

dongsupark:
This PR tries to cover more test cases by running cgroup tests with different combinations of input values. Also it tries to print out detailed output results in the TAP format.

To do that, first of all, we change prototype of `RuntimeOutsideValidate()` as well as `AfterFunc()`, updating all corresponding call sites. And then we update cgroups tests. Changed cgroups tests are:

* Memory 
* CPU
* Network
* Blkio
* HugeTLB

Especially I needed to update CPU cgroups tests a lot, by following the same semantics of memory cgroups. For that, I had to add helpers in `validation/util/linux_resources_cpus.go`.

/cc @alban

### DIFF
--- a/generate/config.go
+++ b/generate/config.go
@@ -94,7 +94,8 @@ func (g *Generator) initConfigLinuxResourcesBlockIO() {
 	}
 }
 
-func (g *Generator) initConfigLinuxResourcesCPU() {
+// InitConfigLinuxResourcesCPU initializes CPU of Linux resources
+func (g *Generator) InitConfigLinuxResourcesCPU() {
 	g.initConfigLinuxResources()
 	if g.Config.Linux.Resources.CPU == nil {
 		g.Config.Linux.Resources.CPU = &rspec.LinuxCPU{}
--- a/validation/linux_cgroups_blkio.go
+++ b/validation/linux_cgroups_blkio.go
@@ -1,16 +1,23 @@
 package main
 
 import (
+	"fmt"
+	"runtime"
+
 	"github.com/mndrix/tap-go"
 	"github.com/opencontainers/runtime-tools/cgroups"
 	"github.com/opencontainers/runtime-tools/validation/util"
 )
 
-func main() {
+func testBlkioCgroups(rate uint64, isEmpty bool) error {
 	var weight uint16 = 500
 	var leafWeight uint16 = 300
-	var major, minor int64 = 8, 0
-	var rate uint64 = 102400
+
+	// It's assumed that a device /dev/sda (8:0) exists on the test system.
+	// The minor number must be always 0, as it's only allowed to set blkio
+	// weights to a whole block device /dev/sda, not to partitions like sda1.
+	var major int64 = 8
+	var minor int64
 
 	t := tap.New()
 	t.Header(0)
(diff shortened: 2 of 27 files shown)


generate/config.go
```diff
@@ -94,7 +94,8 @@ func (g *Generator) initConfigLinuxResourcesBlockIO() {
 	}
 }
 
-func (g *Generator) initConfigLinuxResourcesCPU() {
+// InitConfigLinuxResourcesCPU initializes CPU of Linux resources
+func (g *Generator) InitConfigLinuxResourcesCPU() {
```
liangchenye: this change LGTM

validation/linux_cgroups_blkio.go
```diff
@@ -1,30 +1,100 @@
...
-func main() {
+func testBlkioCgroups() error {
+	var weight uint16 = 500
+	var leafWeight uint16 = 300
+
+	t := tap.New()
+	t.Header(0)
+	defer t.AutoPlan()
+
+	// It's assumed that a device /dev/sda (8:0) exists on the test system.
```
liangchenye: why do we need this assumption?

validation/linux_cgroups_blkio.go
```diff
@@ -1,30 +1,100 @@
...
+
+		err = util.RuntimeOutsideValidate(g, t, util.ValidateLinuxResourcesBlockIO)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func testEmptyBlkio() error {
```
liangchenye: I think it could be part of testBlkioCgroups. We can add more args to the ```cases struct```.  Take https://github.com/opencontainers/runtime-tools/blob/master/validation/state.go#L29 for example.

liangchenye:
>> This PR tries to cover more test cases by running cgroup tests with different combinations of input values. Also it tries to print out detailed output results in the TAP format.

The `combinations` idea looks cool. I just wonder if the testing values were representative since we are not doing performance/pressure test.

validation/linux_cgroups_blkio.go
dongsupark: @liangchenye 
I'm not sure why, as I'm not the original author of the blkio test. The test has always relied on static major/minor numbers, since the beginning:
https://github.com/opencontainers/runtime-tools/commit/f1e02ff6b3bca42a9fa33f918f55cf97da3ab193

What I'm doing with this PR is only to improve the existing test. Though during running the blkio test on my local system, I realized that the test fails because of a missing device `/dev/sda` (8:0). The local machine has only `/dev/nvme` (243:0), not `/dev/sda`. That's why I added the comment above.

Ideally we could improve this part so that it detects the root device at runtime, instead of hard-coding like that. Though I think that it should be done in a separate PR. If you want me to change so, I can make another PR.

validation/linux_cgroups_blkio.go
dongsupark: @liangchenye That sounds like a good idea. Will try to do that.

dongsupark:
@liangchenye Sure, in some cases testing values could not be optimal. But I don't have a better idea.

dongsupark:
Rebased on top of master, and updated CPU & blkio tests.

* Added more nil checks in CPU tests
* Squashed empty blkio tests into `testBlkioCgroups`.

dongsupark:
Updated the PR, to replace `util.Fatal` with `t.Fail`, as done in https://github.com/opencontainers/runtime-tools/pull/645.

validation/util/linux_resources_devices.go
```diff
@@ -48,6 +45,5 @@ func ValidateLinuxResourcesDevices(config *rspec.Spec, state *rspec.State) error
 		}
 	}
 
-	t.AutoPlan()
```
q384566678: According to your idea, there is `t.AutoPlan()` in this file that has not been deleted yet.

q384566678:
Looks good, WDYT @liangchenye .

validation/util/linux_resources_devices.go
dongsupark: @q384566678 Good catch.
I've removed `t.AutoPlan()` from `linux_resources_devices.go` as well as `linux_resources_blkio.go`.

q384566678:
LGTM

PR #640, opened 2018-05-23
validation: test validation test with an empty hostname

dongsupark:
Check if the hostname test works also with an empty hostname, i.e, `setHostname == false`.
Also print out TAP results.


validation/hostname.go
```diff
@@ -1,17 +1,45 @@
...
+		g.SetHostname("hostname-specific")
 	}
-	g.SetHostname("hostname-specific")
 	err = util.RuntimeInsideValidate(g, nil)
+	t.Ok(err == nil, "hostname is set correctly")
 	if err != nil {
+		t.Diagnosticf("expect: err == nil, actual: err != nil")
+	}
+
+	return err
```
liangchenye: We should return nil.
util.Fatal is used when an unexpected error detect, most caused by runtime-tools source code.

validation/hostname.go
```diff
@@ -1,17 +1,45 @@
...
+func main() {
+	t := tap.New()
+	t.Header(0)
+	defer t.AutoPlan()
+
+	if "linux" != runtime.GOOS {
+		t.Skip(1, fmt.Sprintf("linux-specific namespace test"))
+	}
+
+	if err := testHostname(t, false); err != nil {
```
liangchenye: Change these two tests to one loop?

validation/hostname.go
```diff
@@ -1,17 +1,45 @@
...
 import (
+	"fmt"
+	"runtime"
+
+	"github.com/mndrix/tap-go"
 	"github.com/opencontainers/runtime-tools/validation/util"
 )
 
-func main() {
+func testHostname(t *tap.T, setHostname bool) error {
```
liangchenye: We can use
```
func testHostname(t *tap.T, hostname string) error {
```
We don't need to check 
```
    if setHostname {
   }
```

liangchenye:
+1 for wrapping the testing code, it is good way to solve
https://github.com/opencontainers/runtime-tools/issues/582

validation/hostname.go
liangchenye: I think we need a patch to make RuntimeInsideValidate and RuntimeOutsideValidate stronger.
Now it does not different runtime error and runtime-tool/system error.

dongsupark:
@liangchenye Done.
Changed return value to nil, made a loop over the 2 different test cases, got hostname string passed instead of a bool value.

liangchenye:
Thanks, LGTM

PR #642, opened 2018-05-24
validation: add minor checks for ptmx and kill signal

dongsupark:
Inside a container, we need to check if `/dev/ptmx` exists, a symlink to `/dev/pts/ptmx`, as mentioned in the [runtime-spec](https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#default-devices).

Also in `validation/killsig.t`, Test not only SIGTERM, but also SIGUSR1 as well as SIGUSR2.


q384566678:
LGTM

PR #643, opened 2018-05-25
validation: add more test cases for masked paths tests

dongsupark:
Split out the test function into `checkMaskedPaths()`, and add more cases for masked paths like subdirectory, file under subdirectory, and directory under subdirectory.

Test inside container should return error if a relative path is given for masked paths.

Deliberately create an invalid symlink that points out of the container, to see if the test fails inside the container.

Create masked block device, char device, and fifo, to check if they are masked as expected.


q384566678:
LGTM

PR #644, opened 2018-05-31
validation: add more test cases for read-only paths tests

dongsupark:
Split out the test function into `checkReadonlyPaths()`, and add more cases for read-only paths like subdirectory, file under subdirectory, and directory under subdirectory.

Test inside container should return error if a relative path is given for read-only paths.

Deliberately create an invalid read-only symlink that points out of the container, to see if the test fails inside the container.

Create read-only block device, char device, and fifo, to check if they are read-only as expected.

### DIFF
--- a/validation/linux_readonly_paths.go
+++ b/validation/linux_readonly_paths.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


validation/linux_readonly_paths.go
```diff
@@ -28,13 +45,130 @@ func main() {
...
+}
+
+func checkReadonlyRelPaths() error {
+	g, err := util.GetDefaultGenerator()
+	if err != nil {
+		return err
+	}
+
+	// Deliberately set a relative path to be read-only, and expect an error
+	readonlyRelPath := "/etc/../readonly-relpath"
```
q384566678: Is it a relative path? I think it equal to `/readonly-relpath`. It's a absolute path.

dongsupark: @q384566678 Fixed the relative path.
Also fixed the same one in https://github.com/opencontainers/runtime-tools/pull/643.

q384566678: Why not  use `readonlyRelPath := "readonly-relpath"` ?

dongsupark: @q384566678 Done. Also done in #643.

q384566678:
LGTM

PR #645, opened 2018-06-01
validation: use t.Fail when checking for main test errors

dongsupark:
So far `util.Fatal()`, which calls `os.Exit(1)`, has been widely used to exit immediately when a critical error happened. Its downside is though that no further function can be called after that moment, even defer functions cannot be called. That's fine if there's nothing to clean up or the test is simple enough. Though it could be an issue, for example, when a test should print out TAP output by calling `t.AutoPlan()`, or when it should clean up something.

So let's try to use `util.Fatal()` only for errors from critical cases like errors from `util.GetDefaultGenerator()` or `util.PrepareBundle()`. In case of main test errors, use instead `t.Fail(err.Error())` to print out errors in TAP outputs.

Partly addresses https://github.com/opencontainers/runtime-tools/issues/582

/cc @liangchenye

### DIFF
--- a/validation/config_updates_without_affect/config_updates_without_affect.go
+++ b/validation/config_updates_without_affect/config_updates_without_affect.go
@@ -82,7 +82,7 @@ func main() {
 		err = r.Delete()
 	}
 	if err != nil {
-		util.Fatal(err)
+		t.Fail(err.Error())
 	}
 
 	t.AutoPlan()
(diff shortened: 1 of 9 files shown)


q384566678:
The same content in `start.go` needs to be modified.

dongsupark:
@q384566678 Done. Updated `start.go` as well.

dongsupark:
Rebased.

q384566678:
LGTM

PR #647, opened 2018-06-04
travis: add go 1.10

q384566678:


liangchenye:
The CI failed, Go 1.10 seems quite different from previous versions.

liangchenye:
https://github.com/opencontainers/runtime-tools/blob/master/api/socket/socket.go#L18
```
// Message is the normal data for messages passed on the console socket.
type Message struct {
	// Type of message being passed
	Type string `json:"type"`
}

// Response is the normal data for response messages.
type Response struct {
	Message

	// Message is a phrase describing the response.
	Message string `json:"message,omitempty"`
}
```
Error detected here: `api/socket/socket.go:22:2: duplicate field Message`
@wking  can we change `type Message struct` to `type Data struct` ?

q384566678:
The reason for the current error is that there are multiple files in the document validation in the `go vet` detection, which contain the main function. Now I think of setting up folders for them respectively. Do you have any good ideas?

```
➜  runtime-tools git:(add-go1.10) ✗ go vet ./validation/...
# github.com/opencontainers/runtime-tools/validation
validation/create.go:16:6: main redeclared in this block
        previous declaration at validation/config_updates_without_affect.go:18:6
validation/default.go:7:6: main redeclared in this block
        previous declaration at validation/create.go:16:6
validation/delete.go:16:6: main redeclared in this block
        previous declaration at validation/default.go:7:6
validation/hooks.go:17:6: main redeclared in this block
        previous declaration at validation/delete.go:16:6
validation/hooks_stdin.go:78:6: main redeclared in this block
        previous declaration at validation/hooks.go:17:6
validation/hostname.go:27:6: main redeclared in this block
        previous declaration at validation/hooks_stdin.go:78:6
validation/kill.go:16:6: main redeclared in this block
        previous declaration at validation/hostname.go:27:6
validation/kill_no_effect.go:16:6: main redeclared in this block
        previous declaration at validation/kill.go:16:6
validation/killsig.go:20:6: main redeclared in this block
        previous declaration at validation/kill_no_effect.go:16:6
validation/linux_cgroups_blkio.go:8:6: main redeclared in this block
        previous declaration at validation/killsig.go:20:6
validation/linux_cgroups_blkio.go:8:6: too many errors
```

dongsupark:
@q384566678 Yeah indeed.
`go vet` 1.10 does not allow multiple definitions of main function in a single directory. That actually looks like a good behavior.
I agree that we need to move each test file into a subdirectory, for example, from `validation/create.go` to `validation/create/create.go`.
Then probably I would also need to update my PRs. That's fine. :-)

q384566678:
> Then probably I would also need to update my PRs. That's fine. :-)

Before this PR is merged I will review your PRs first.

q384566678:
@liangchenye updated, PTAL.

PR #652, opened 2018-06-22
mountinfo: parse empty strings in source

alban:
The source of a mount in /proc/self/mountinfo can unfortunately be an
empty string. Before this patch, 'mount' and 'mountpoint' fail as
following:
```
  $ sudo mount -t tmpfs "" /tmp/bb
  $ mount
  mount: /proc/self/mountinfo: parse error at line 64 -- ignored
  $ mountpoint /tmp/bb
  /tmp/bb is not a mountpoint
```
This patch fixes the parsing.

-----

Similar patch in util-linux: https://www.spinics.net/lists/linux-fsdevel/msg128896.html
Similar PR in runc: https://github.com/opencontainers/runc/pull/1829
Related patch that would fix this kernel-side: https://patchwork.kernel.org/patch/10349095/

### DIFF
--- a/cgroups/cgroups.go
+++ b/cgroups/cgroups.go
@@ -44,7 +44,7 @@ func FindCgroup() (Cgroup, error) {
 		fields := strings.Split(text, " ")
 		// Safe as mountinfo encodes mountpoints with spaces as \040.
 		index := strings.Index(text, " - ")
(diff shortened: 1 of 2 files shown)


liangchenye:
LGTM
It works in my test.

PR #654, opened 2018-07-02
add resource validation after delete

q384566678:
```
DeleteResImplement: Deleting a container MUST delete the resources that were created during the create step.
DeleteOnlyCreatedRes: Note that resources associated with the container, but not created by this container, MUST NOT be deleted.
```
At present, `DeleteResImplement` is implemented, but it is not yet clear how to implement `DeleteOnlyCreatedRes`. (Write the container id to the existing `cgroup tasks`?)
@liangchenye @dongsupark @wking WDYT?

### DIFF
--- a/validation/delete_only_create_resources/delete_only_create_resources.go
+++ b/validation/delete_only_create_resources/delete_only_create_resources.go
@@ -1,0 +1,88 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strconv"
+	"time"
+
+	tap "github.com/mndrix/tap-go"
+	"github.com/mrunalp/fileutils"
+	rspec "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/opencontainers/runtime-tools/specerror"
+	"github.com/opencontainers/runtime-tools/validation/util"
+	uuid "github.com/satori/go.uuid"
+)
+
+func main() {
+	t := tap.New()
+	t.Header(0)
+
+	// Create a cgroup
+	cgPath := "/sys/fs/cgroup"
+	testPath := filepath.Join(cgPath, "pids", "cgrouptest")
+	os.Mkdir(testPath, 0755)
+	defer os.RemoveAll(testPath)
+
+	bundleDir, err := util.PrepareBundle()
+	if err != nil {
+		util.Fatal(err)
+	}
+	defer os.RemoveAll(bundleDir)
+
+	r, err := util.NewRuntime(util.RuntimeCommand, bundleDir)
+	if err != nil {
+		util.Fatal(err)
+	}
+
+	r.SetID(uuid.NewV4().String())
+	g, err := util.GetDefaultGenerator()
+	if err != nil {
+		util.Fatal(err)
+	}
+
+	err = r.SetConfig(g)
+	if err != nil {
+		util.Fatal(err)
+	}
+	err = fileutils.CopyFile("runtimetest", filepath.Join(r.BundleDir, "runtimetest"))
+	if err != nil {
+		util.Fatal(err)
+	}
+
+	err = r.Create()
+	if err != nil {
+		util.Fatal(err)
+	}
+
+	state, err := r.State()
+	if err != nil {
+		util.Fatal(err)
+	}
+	// Add the container to the cgroup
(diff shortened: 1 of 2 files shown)


dongsupark:
From a quick look, this PR looks good.
About `DeleteOnlyCreatedRes`, could it be maybe written like this?

* Create a cgroup `cgrouptest`
* Run a container that joins the cgroup, by writing its PID to `/sys/fs/cgroup/pids/cgrouptest/tasks`.
* Delete the container
* See if the cgroup stays, without being deleted.

I've not tried it on my own though.
See also [the cgroups document in the Linux kernel](https://github.com/torvalds/linux/blob/master/Documentation/cgroup-v1/cgroups.txt).

q384566678:
@dongsupark thanks for your suggestion.
Currently both have been implemented. @liangchenye PTAL.

validation/deleate_only_create_resources/delte_only_create_resources.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		util.Fatal(err)
+	}
+
+	state, err := r.State()
+	if err != nil {
+		util.Fatal(err)
+	}
+	// Add the container to the cgroup
+	ioutil.WriteFile(filepath.Join(pidsPath, "cgrouptset/tasks"), []byte(strconv.Itoa(state.Pid)), 0644)
```
alban: typo: "cgrouptset" != "cgrouptest" from line 26. You could check the return value of `ioutil.WriteFile` and use a constant for that directory.

validation/deleate_only_create_resources/delte_only_create_resources.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+func main() {
+	t := tap.New()
+	t.Header(0)
+
+	// Create a cgroup
+	cgPath := "/sys/fs/cgroup"
+	pidsPath := filepath.Join(cgPath, "pids")
+	os.Mkdir(filepath.Join(pidsPath, "cgrouptest"), 0755)
```
alban: The test should cleanup (rmdir) that directory after the test is completed.

validation/deleate_only_create_resources/delte_only_create_resources.go
```diff
@@ -0,0 +1,84 @@
...
+	uuid "github.com/satori/go.uuid"
+)
+
+func main() {
+	t := tap.New()
+	t.Header(0)
+
+	// Create a cgroup
+	cgPath := "/sys/fs/cgroup"
+	pidsPath := filepath.Join(cgPath, "pids")
```
alban: This will not work with cgroup-v2 because the paths are different but that's a larger in OCI.

validation/deleate_only_create_resources/delte_only_create_resources.go
```diff
@@ -0,0 +1,84 @@
+package main
```
alban: typos in the filename: 
```
validation/deleate_only_create_resources/delte_only_create_resources.go
            ^^^^^                        ^^^^^
```

validation/deleate_only_create_resources/delte_only_create_resources.go
q384566678: I don't know much about cgroup-v2, so should we do validation here in two different cases?

I will learn about cgroup-v2 later. There is still much work to be done about cgroup-v2 in the runtime-tool.

validation/deleate_only_create_resources/delte_only_create_resources.go
q384566678: updated, thanks.

validation/deleate_only_create_resources/delte_only_create_resources.go
alban: A lot about the spec is cgroup-v1 only. I don't think we can do add support for v1 and v2 everywhere at once. So I don't think this should block this PR.

q384566678:
The two commit have resolved the remaining issues in `specerror`. @liangchenye @mrunalp PTAL

q384566678:
ping @liangchenye @mrunalp

q384566678:
@liangchenye @vbatts @mrunalp PTAL

dongsupark:
I've tested it, it works well.
LGTM. (but I'm not a maintainer)

vbatts:
LGTM

PR #658, opened 2018-07-17
Fix TAP output with multiple RuntimeInsideValidate

alban:
Getting proper TAP output with `RuntimeInsideValidate` was not possible when running several containers in a single test file. With this PR, we can optionally pass the TAP object in `RuntimeInsideValidate` and then `runtimetest` in the container will no longer directly print its own TAP output on stdout but it will be intercepted and encapsulated in the outer TAP output.


docs/devel_guidelines.md
```diff
@@ -7,9 +7,52 @@ Each validation test has a `.go` file in the `validation/` directory and can be
...
+```go
+err = util.RuntimeInsideValidate(g, nil, nil)
+if err != nil {
+        util.Fatal(err)
+}
+```
+
+#### Using `util.RuntimeInsideValidate` and encapsulation
+
+Similar to previously but the test consumes the output from `runtimetest` and re-emit a single TAP result for the container run.
```
dongsupark: Wouldn't it more precise to write `Similar to the passthrough variant but...`?

docs/devel_guidelines.md
```diff
@@ -7,9 +7,52 @@ Each validation test has a `.go` file in the `validation/` directory and can be
...
+So far, we have three kinds of validation tests and they print the TAP output differently:
+
+#### Using `util.RuntimeOutsideValidate`
+
+They create a container but without executing `runtimetest`. The test program itself must print the TAP output.
+
+Example:
+```go
+err = util.RuntimeOutsideValidate(g, t, func(config *rspec.Spec, t *tap.T, state *rspec.State) error {
+        err := test_foo()
```
dongsupark: NIT: `s/test_foo/testFoo/` for the idiomatic Go coding style?

alban:
Thanks for the review. Fixed & rebased.

dongsupark:
:+1:

PR #660, opened 2018-07-31
validation: mounts: fix condition of source & type check

alban:
The runtime-spec says that [mount source is optional](https://github.com/opencontainers/runtime-spec/blob/v1.0.1/config.md#mounts). So let's relax condition of the mount source path check, so that it only checks for an empty mount source. Ditto for the type field.

The 'mount' test now tries different mount options: bind and not bind, recursive or not, different mount propagation modes.

runc passes the test after https://github.com/opencontainers/runc/pull/1845

Closes: https://github.com/opencontainers/runtime-tools/issues/651

Based on work from: Dongsu Park <dongsu@kinvolk.io> @dongsupark 

### DIFF
--- a/cmd/runtimetest/main.go
+++ b/cmd/runtimetest/main.go
@@ -1091,14 +1091,32 @@ func mountMatch(configMount rspec.Mount, sysMount *mount.Info) error {
 		return fmt.Errorf("mount destination expected: %v, actual: %v", configMount.Destination, sys.Destination)
 	}
 
-	if configMount.Type != sys.Type {
+	isBind := false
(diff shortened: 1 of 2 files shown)


dongsupark:
LGTM (but I'm not maintainer)
Tested, and it works fine with the current master branch of runc.
Thanks for taking my code into this PR. :+1:

q384566678:
LGTM

q384566678:
@alban need rebase.

dongsupark:
@q384566678 Rebased. PTAL

q384566678:
LGTM

PR #662, opened 2018-09-11
generate: add short options

q384566678:
For easier use, add short options to `env` and `label` option.


liangchenye:
I don't think 'e' stands for 'env' and 'l' stands for 'label' are the universal naming habits.
I -1 to this.

q384566678:
Fine, closed.

PR #671, opened 2018-10-28
Bump Go versions

HaraldNordgren:


HaraldNordgren:
@q384566678 Fair enough. Close this if you want to.

PR #672, opened 2018-10-29
Add generate option

q384566678:
Add the corresponding generate option according to the newly updated spec.


q384566678:
This pr is a perfection of the generate function, and some verification will be done in the next pr.
@liangchenye @hqhq @vbatts @mrunalp PTAL.

q384566678:
A commit is added to verify the input value.
PTAL.

PR #674, opened 2018-12-06
 generate: fix capabilities add/drop option

### LINKED ISSUE #673: Generate subcommand is unable to add multiple capabilities to config with single flag
The description for the capability related flags for the generate subcommand made it seem like that I could add or remove multiple capabilities with one flag. For example, the description for --process-cap-add-ambient is just "Add Linux ambient capabilities". I've tried many varieties of --process-cap-add-ambient '["CAP_AUDIT_WRITE", "CAP_CHOWN", ...]' but none seem to work. They only way it seems to work is by specifying the flag each time for each capability, which leads to a very verbose shell command.

If multiple capabilities per flag is not supported, that would be a nice enhancement. If it is supported, having a doc example for this would help.

It would also help to have a top level --process-cap-add convenience flag for adding a capability to all 5 capability sets.

q384566678:
After modification, you can add or delete multiple capabilities with one command.

```
oci-runtime-tool generate  --process-cap-add-ambient CAP_FOWNER,CAP_FSETID
```
Fixes #673 


liangchenye:
LGTM

PR #680, opened 2019-01-16
update godeps

q384566678:
Temporarily update the third-party package first to avoid the version being too old. Later we can add a maintain policy as recommended by @liangchenye.


Godeps/Godeps.json
```diff
@@ -45,8 +46,8 @@
 		},
 		{
 			"ImportPath": "github.com/satori/go.uuid",
-			"Comment": "v1.1.0",
-			"Rev": "879c5887cd475cd7864858769793b2ceb0d44feb"
+			"Comment": "v1.2.0-12-gb2ce238",
```
thaJeztah, line 49: Same question here; would the last tagged version of this work instead?

```suggestion
			"Comment": "v1.2.0",
```

https://github.com/satori/go.uuid/compare/v1.2.0...master

Perhaps we should replace `github.com/satori/go.uuid` with `github.com/google/uuid`, which is commonly used in other projects (e.g. in containerd), and looks to be more actively maintained; https://github.com/google/uuid

Godeps/Godeps.json
```diff
@@ -64,8 +65,8 @@
 		},
 		{
 			"ImportPath": "github.com/urfave/cli",
-			"Comment": "v1.19.1",
-			"Rev": "0bdeddeeb0f650497d603c4ad7b20cfe685682f6"
+			"Comment": "v1.20.0-81-gb67dcf9",
+			"Rev": "b67dcf995b6a7b7f14fad5fcb7cc5441b05e814b"
```
thaJeztah, line 69: Same question here; was there a reason not to use a tagged version?

Godeps/Godeps.json
```diff
@@ -8,8 +8,8 @@
 	"Deps": [
 		{
 			"ImportPath": "github.com/blang/semver",
-			"Comment": "v3.5.0",
-			"Rev": "b38d23b8782a487059e8fc8773e9a5b228a77cb6"
+			"Comment": "v3.5.1-4-g3c10740",
+			"Rev": "3c1074078d32d767e08ab2c8564867292da86926"
```
thaJeztah, line 12: Was there a reason not to pick a tagged release for this?

thaJeztah (changes requested):

PR #681, opened 2019-01-16
generate: add oci-version option

q384566678:
Add oci-version option to generate to set the version of the Open Container Initiative runtime specification.


q384566678:
@liangchenye PTAL

mrunalp:
LGTM

PR #687, opened 2019-03-03
validation-tests: fix several tests

giuseppe:
this PR includes several minor fixes to the validation tests.  I've found these issues while preparing `crun` to be fully OCI compliant: https://github.com/giuseppe/crun/pull/16

More details are included in the commit message for each patch.

I've added a patch for the Makefile to print what validation-tests are used.  I want to use it on Travis to easily read the list of the tests and disabling some of them, namely some cgroup tests that require a special configuration (I could just repeat the same find query, but I'd prefer to keep this logic in the runtime-tools Makefile)

### DIFF
--- a/validation/process_capabilities_fail/process_capabilities_fail.go
+++ b/validation/process_capabilities_fail/process_capabilities_fail.go
@@ -20,7 +20,7 @@ func main() {
 	if err != nil {
 		util.Fatal(err)
 	}
-	g.AddProcessCapabilityBounding("CAP_TEST")
+	g.Config.Process.Capabilities.Bounding = append(g.Config.Process.Capabilities.Bounding, "CAP_TEST")
 	err = util.RuntimeInsideValidate(g, nil, nil)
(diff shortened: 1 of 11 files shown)


validation/process_capabilities_fail/process_capabilities_fail.go
```diff
@@ -20,7 +20,7 @@ func main() {
 	if err != nil {
 		util.Fatal(err)
 	}
-	g.AddProcessCapabilityBounding("CAP_TEST")
+	g.Config.Process.Capabilities.Bounding = append(g.Config.Process.Capabilities.Bounding, "CAP_TEST")
```
mrunalp: Is there a bug in the helper function that we should fix?

giuseppe: no, I think it is fine that the helper function checks whether the input is valid and rejects unknown capabilities

q384566678:
LGTM

PR #689, opened 2019-03-05
validate_test: add test to TestCheckPlatform

zhouhao3:

### DIFF
--- a/validate/validate_test.go
+++ b/validate/validate_test.go
@@ -656,6 +656,13 @@ func TestCheckPlatform(t *testing.T) {
 			platform: "windows",
 			expected: specerror.NonError,
 		},
+		{
+			val: rspec.Spec{
+				Version: "1.0.0",
+			},
+			platform: "solaris",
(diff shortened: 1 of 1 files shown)


validate/validate_test.go
```diff
@@ -656,6 +656,13 @@ func TestCheckPlatform(t *testing.T) {
 			platform: "windows",
 			expected: specerror.NonError,
 		},
+		{
+			val: rspec.Spec{
+				Version: "1.0.0",
```
kolyshkin, line 661: @zhouhao3 can you please change this to 1.0.2 (which is currently the minimum version required (as explained in commit 67884fc9e3b1a61a85db024c0cf33710e57cb37f)?

rhatdan:
@zhouhao3 Are you still interested in this, if not please close or update.

validate/validate_test.go
kolyshkin: @zhouhao3 ^^^

PR #693, opened 2019-04-10
Improve performance of AddProcessEnv

umohnani8:
AddProcessEnv was sequentially going through the env array
to check for duplicates, which would become very slow if we
add a huge amount of ENV vars.
Add a map to Generator to keep track of the ENV vars already
added, so that when AddProcessEnv checks for duplicates, it is
much faster.
Also add a new function called AddMultipleProcessEnv,
which takes a []string of ENVs and adds them to the config in one
go.
Add unit tests for these functions.

Signed-off-by: Urvashi Mohnani <umohnani@redhat.com>

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -29,6 +29,9 @@ var (
 type Generator struct {
 	Config       *rspec.Spec
 	HostSpecific bool
+	// This is used to keep a cache of the ENVs added to improve
+	// performance when adding a huge number of ENV variables
+	envMap map[string]int
 }
 
 // ExportOptions have toggles for exporting only certain parts of the specification
@@ -236,7 +239,12 @@ func New(os string) (generator Generator, err error) {
 		}
 	}
 
-	return Generator{Config: &config}, nil
+	envCache := map[string]int{}
+	if config.Process != nil {
+		envCache = createEnvCacheMap(config.Process.Env)
+	}
+
+	return Generator{Config: &config, envMap: envCache}, nil
 }
 
 // NewFromSpec creates a configuration Generator from a given
@@ -246,8 +254,14 @@ func New(os string) (generator Generator, err error) {
 //
 //   generator := Generator{Config: config}
 func NewFromSpec(config *rspec.Spec) Generator {
+	envCache := map[string]int{}
+	if config != nil && config.Process != nil {
+		envCache = createEnvCacheMap(config.Process.Env)
+	}
+
 	return Generator{
 		Config: config,
+		envMap: envCache,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


umohnani8:
@mrunalp PTAL

mrunalp:
LGTM

mrunalp:
@vbatts ptal

umohnani8:
@vbatts ping :)

generate/generate.go
```diff
@@ -456,21 +486,44 @@ func (g *Generator) ClearProcessEnv() {
...
+// any duplicates
+// This is called by both AddMultipleProcessEnv and AddProcessEnv
+func (g *Generator) addEnv(env, key string) {
+	if idx, ok := g.envMap[key]; ok {
+		// The ENV exists in the cache, so change its value in g.Config.Process.Env
+		g.Config.Process.Env[idx] = env
+	} else {
+		// else the env doesn't exist, so add it and add it's index to g.envMap
+		g.Config.Process.Env = append(g.Config.Process.Env, env)
+		g.envMap[key] = len(g.Config.Process.Env) - 1
```
rhatdan: Nit: Why not
g.envMap[key] = env

generate/generate_test.go
```diff
@@ -93,3 +94,65 @@ func TestRemoveMount(t *testing.T) {
...
+
+func TestMultipleEnvCaching(t *testing.T) {
+	// Start with empty ENV and add a few
+	g, err := generate.New("windows")
+	if err != nil {
+		t.Fatal(err)
+	}
+	newEnvs := []string{"k1=v1", "k2=v2"}
+	expected := []string{"k1=v1", "k2=v2"}
+	g.AddMultipleProcessEnv(expected)
```
rhatdan: Nit: This should be newEnvs

generate/generate.go
umohnani8: Because, we want to keep track of the position in the string array this env is at so we can easily go in and update it if an existing env is added again with another value.

generate/generate_test.go
umohnani8: Fixed!

generate/generate.go
```diff
@@ -273,11 +287,27 @@ func NewFromTemplate(r io.Reader) (Generator, error) {
...
+
 	return Generator{
 		Config: &config,
+		envMap: envCache,
 	}, nil
 }
 
+// createEnvCacheMap creates a hash map with the ENV variables given by the config
+func createEnvCacheMap(env []string) map[string]int {
+	envMap := make(map[string]int)
```
vbatts: if you're going to use make(), and you know the length of `env`, it seems that you might make it that length

generate/generate.go
```diff
@@ -456,21 +486,44 @@ func (g *Generator) ClearProcessEnv() {
 		return
 	}
 	g.Config.Process.Env = []string{}
+	// Clear out the env cache map as well
+	g.envMap = make(map[string]int)
```
vbatts: would `nil` accomplish this as well?

vbatts:
couple of thoughts, but otherwise LGTM

generate/generate.go
umohnani8: right, fixed! thanks :)

generate/generate.go
umohnani8: I got rid of the make, and just define the map. I think nil might cause issues, but hopefully this is better.

umohnani8:
@vbatts addresses comments, this should be ready now :)

vbatts:
LGTM

umohnani8:
Anything more we need to do to get this in?

PR #696, opened 2019-06-23
 Fix cgroup hugetlb size prefix for kB 

odinuge:
This is ~a cherry-pick of https://github.com/opencontainers/runc/pull/2065.

The hugetlb cgroup control files (introduced here in 2012:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=abb8206cb0773)
use "KB" and not "kB"
(https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/mm/hugetlb_cgroup.c\?h\=v5.0\#n349\).

The behavior in the kernel has not changed since the introduction, and
the current code using "kB" will therefore fail on devices with small
amounts of ram (see
kubernetes/kubernetes#77169) running a kernel
with config flag CONFIG_HUGETLBFS=y

As seen from the code in "mem_fmt" inside hugetlb_cgroup.c, only "KB",
"MB" and "GB" are used, so the others may be removed as well.

Here is a real world example of the files inside the
"/sys/kernel/mm/hugepages/" directory:
- "hugepages-64kB"
- "hugepages-2048kB"
- "hugepages-32768kB"
- "hugepages-1048576kB"

And the corresponding cgroup files:
- "hugetlb.64KB._____"
- "hugetlb.2MB._____"
- "hugetlb.32MB._____"
- "hugetlb.1GB._____"

### DIFF
--- a/cgroups/cgroups_v1.go
+++ b/cgroups/cgroups_v1.go
@@ -18,6 +18,11 @@ type CgroupV1 struct {
 	MountPath string
 }
 
+// HugePageSizeUnitList is a list of the units used by the linux kernel when
+// naming the HugePage control files.
+// https://www.kernel.org/doc/Documentation/cgroup-v1/hugetlb.txt
+var HugePageSizeUnitList = []string{"B", "KB", "MB", "GB", "TB", "PB"}
+
 func getDeviceID(id string) (int64, int64, error) {
 	elem := strings.Split(id, ":")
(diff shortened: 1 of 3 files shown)


odinuge:
cc @vbatts

vbatts:
LGTM

vbatts:

hqhq:
LGTM

PR #697, opened 2019-07-19
Switch interface to use LinuxNamespaceType rather then string

rhatdan:
The runtime-spec uses spec.LinuxNamespaceType for the different namespace
strings rather then string directly.

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -444,6 +444,12 @@ func (g *Generator) SetProcessUsername(username string) {
 	g.Config.Process.User.Username = username
 }
 
+// SetProcessUmask sets g.Config.Process.User.Umask.
+func (g *Generator) SetProcessUmask(umask uint32) {
+	g.initConfigProcess()
+	g.Config.Process.User.Umask = umask
+}
+
 // SetProcessGID sets g.Config.Process.User.GID.
 func (g *Generator) SetProcessGID(gid uint32) {
 	g.initConfigProcess()


rhatdan:
@mrunalp @wking PTAL

rhatdan:
@q384566678 PTAL

rhatdan:
@vbatts PTAL

generate/generate.go
```diff
@@ -1419,25 +1419,8 @@ func (g *Generator) DropProcessCapabilityPermitted(c string) error {
...
-		return rspec.LinuxNamespace{Type: rspec.UTSNamespace, Path: path}, nil
-	case "user":
-		return rspec.LinuxNamespace{Type: rspec.UserNamespace, Path: path}, nil
-	case "cgroup":
-		return rspec.LinuxNamespace{Type: rspec.CgroupNamespace, Path: path}, nil
-	default:
-		return rspec.LinuxNamespace{}, fmt.Errorf("unrecognized namespace %q", ns)
-	}
+func mapStrToNamespace(ns rspec.LinuxNamespaceType, path string) (rspec.LinuxNamespace, error) {
+	return rspec.LinuxNamespace{Type: ns, Path: path}, nil
```
q384566678: SGTM. However,  if you change this way, you can't judge whether the added type value is correct.

rhatdan: Suggestions on a better way?

PR #700, opened 2019-10-21
Add missing clone rule for s390x. Fixes #699

### LINKED ISSUE #699: Generated Seccomp profile rejects clone syscall on s390x.
The s390x architecture requires an [extra seccomp masked_eq rule](https://github.com/cri-o/cri-o/blob/master/vendor/github.com/seccomp/containers-golang/seccomp.json#L607-L632) to allow `clone` syscalls.  This rule is present in cri-o's default configuration, but is missing in the configuration generated by opencontainers/runtime-tools.  This causes syscalls to `clone` to be rejected by seccomp in cri-o subsystems which utilize this generated seccomp profile.

This issue affects the ose-pod container in openshift.  When cri-o tries to instantiate /usr/bin/pod on s390x during the openshifft bootstrap, there is a `clone` syscall in the golang runtime startup which fails with a Permission Denied error.  This prevents containers from being instantiated by the bootstrap kubelet, causing the openshift bootstrap to fail.

dbenoit17:
Fixes https://github.com/opencontainers/runtime-tools/issues/699

This PR adds the missing [clone masked_eq seccomp](https://github.com/cri-o/cri-o/blob/master/vendor/github.com/seccomp/containers-golang/seccomp.json#L607-L632) rule which causes clone syscalls to fail on s390x.

I have only tested this patch directly in cri-o so far.  Please consider this PR a work in progress while I become familiar with this project and its processes such as writing and running unit tests, etc.

Additionally, since this patch only affects s390x, please let me know if there is anything extra I should do to test the architecture-specific functionality.

Thanks,
DB

### DIFF
--- a/generate/seccomp/seccomp_default.go
+++ b/generate/seccomp/seccomp_default.go
@@ -566,6 +566,20 @@ func DefaultProfile(rs *specs.Spec) *rspec.LinuxSeccomp {
 			},
 		}...)
 		/* Flags parameter of the clone syscall is the 2nd on s390 */
+		syscalls = append(syscalls, []rspec.LinuxSyscall{
+			{
+				Names:  []string{"clone"},
+				Action: rspec.ActAllow,
+				Args: []rspec.LinuxSeccompArg{
+					{
+						Index:    1,
+						Value:    2080505856,
+						ValueTwo: 0,
+						Op:       rspec.OpMaskedEqual,
+					},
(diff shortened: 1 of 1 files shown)


Prashanth684:
can somebody lgtm this? we need this for s390x Openshift installations.

mrunalp:
LGTM

PR #713, opened 2021-02-12
validation: read pid in PostCreate

giuseppe:
do not read the container pid in PreDelete as it might be already
terminated ("true" is a short lived process).

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>


giuseppe:
@tianon @dqminh @hqhq PTAL

tianon (approved):
The logic seems sane, but I have to admit I'm not very familiar with this code.

hqhq:
LGTM

PR #725, opened 2021-09-24
Change /dev to be mounted by default with /noexec

rhatdan:
Podman had an issue, where someone was attemptig to mount all tmpfs
within the container as noexec. They were able to get most of it done
but "/dev", because it was done down in the runtime spec.

I can think of no reason why "/dev", should not be mounted with noexec
especially within a container.  I know it is not mounted by default in
Fedora that way, but I do not know why.

Debian looks like it has made the change, and only one bug a couple of
years ago showed issues, which would not apply to containers.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=940171

Anyways this would make containers "slightly" more secure, and I think it
is worth doing.


rhatdan:
@mrunalp PTAL

rhatdan:
@runtime-tools-maintainers PTAL

rhatdan:
@ caniszczyk, @crosbymichael, @dqminh, @hqhq, @liangchenye, @mashimiao, @mrunalp, @tianon, @vbatts,@vishh, @zhouhao3 PTAL

rhatdan:
For what it is worth, I ran this change through Podman's test suite with no failures.
https://github.com/containers/podman/pull/11349

tianon (approved):
LGTM, although I'm not sure whether many other projects are using this particular code? (I believe runc, containerd, etc all have their own versions of this)

rhatdan:
Well it is used in Podman, Buildah, CRI-O.

mrunalp:
LGTM

giuseppe (approved):
LGTM

vbatts:
bah, Travis needs to be removed and switched to GitHub actions

amye:
@caniszczyk ^ 
TravisCI is blocking things

PR #727, opened 2021-10-14
Migrate to go mod

kolyshkin:
This is a minimal migration from Godeps to native go modules, making it possible to use recent go versions.

Closes: #716
Closes: #704

Issue #726, opened 2021-09-29
:rotating_light: Overhaul CI steps


vbatts:
* [x] remove travis (it is now defunct and blocking merges)
* [x] switch to github actions
* [x] switch to Github CODEOWNERS
* [x] remove pullapprove

kolyshkin:
I have removed pullapprove integration, instead enabling master branch protection rules (two LGTMs from maintainers are required to merge a PR, no direct pushes are allowed, etc).

@vbatts can you please elaborate on CODEOWNERS? As per my older comment,

>  I read https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners but I don't quite get the benefits of using CODEOWNERS.

vbatts:
whoa sorry (since October 😓 )
CODEOWNERS is being used to migrate off of pullapprove (and whatever), to use the built in github approval configuration

PR #728, opened 2021-10-15
Switch to GHA CI

kolyshkin:
~~_This PR builds on and currently includes PR #727, thus the draft status._~~

This removes Travis CI config and adds GHA CI. You can see how it looks like at https://github.com/kolyshkin/runtime-tools/actions/runs/1361288633 and https://github.com/kolyshkin/runtime-tools/pull/1

Also:
 - fix many linter warnings from the default golangci-lint config
    (alas not all of them from errcheck -- there are way too many 😿, I gave up)
 - add golangci-lint, fix all its warnings
 - re-add commit subject length validation

Addresses: #726

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,6 +18,24 @@ jobs:
         with:
           version: v1.42
 
+  commit:
+    runs-on: ubuntu-20.04
+    # Only check commits on pull requests.
+    if: github.event_name == 'pull_request'
+    steps:
+      - name: get pr commits
+        id: 'get-pr-commits'
+        uses: tim-actions/get-pr-commits@v1.1.0
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: check subject line length
+        uses: tim-actions/commit-message-checker-with-regex@v0.3.1
+        with:
+          commits: ${{ steps.get-pr-commits.outputs.commits }}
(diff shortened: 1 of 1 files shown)


kolyshkin:
> This all looks sane to me -- do you think it would be useful for me to try pushing your branch to a new branch here on the origin so the PR can actually run the actions?

As noted in description, I have those ran at https://github.com/kolyshkin/runtime-tools/actions/runs/1344345504, don't think it makes any difference which repo it is.

This is still WIP; currently working on enabling golangci-lint and it's overwhelming :( Guess I'll have to disable errcheck linter :(

I'll let you know once it's ready @tianon, thanks for looking!

tianon:
Thanks to Travis it's going to require a repository administrator (which excludes me 😅).

@vbatts is that you? @caniszczyk?

IMO both PRs look great and merging this which contains both seems like a simpler path, but I don't have any issue with any solution that unblocks Kir and gets this repo able to progress again. 😁

caniszczyk:
I truly don't know how to disable the Travis CI check

kolyshkin:
@caniszczyk this is not about disabling Travis, this is about removing travis (and, maybe, pullapprove, too) from the list of jobs required to merge a PR. This is under Settings -> Branches -> Branch protection rules -> master -> edit -> Status checks that are required. Need to remove travis-ci from that list (and later add GHA CI jobs added by this PR).

PR #739, opened 2021-11-10
Add CODEOWNERS, remove .pullapprove.yml

jdolitsky:


tianon (approved):

tianon:
cc @opencontainers/runtime-tools-maintainers

CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @crosbymichael @dqminh @hqhq @liangchenye @mashimiao @mrunalp @q384566678 @tianon @vbatts @vishh
```
hqhq, line 1: Can we keep using `runtime-tools-maintainers` so we don't have to maintain multiple maintainer lists?

According to https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-syntax , I guest we can specify a team name here.

kolyshkin: @jdolitsky PTAL ^^

vbatts: I think making it a group name, and then just managing the group members is a solid choice

kolyshkin:
Closing in favor of https://github.com/opencontainers/runtime-tools/pull/756

PR #741, opened 2022-01-04
Fix hanging on runc create.

kolyshkin:
This reverts commit 0ab61aed92a67a6e72b68fc0dd6322c95e4bb05b, and also
* changes `ReadStandardStreams` to use `ioutils.ReadFile`;
* changes the uuid module used by the code (see commit abcb94dc55c6c4a).

Using `bytes.Buffer` instead of plain `os.File` results in go runtime
creating two sets of pipes and spawning two threads (for stdout and
stderr) that copy the data from the pipe to the buffers.

It is not working when we execute runc create, because it spawns a child
(runc init) whose stdout and stderr are kept connected to the write ends
of the above pipes. As a result, those copying routines never finish,
and `cmd.Run()` never returns as it waits for those routines.

Therefore, let's go back to using temporary files. Note that the new
code is using `ioutil.ReadAll` instead of reusing the opened file
descriptors, which seems a bit cleaner.

Closes: #735

### DIFF
--- a/validation/util/container.go
+++ b/validation/util/container.go
@@ -1,15 +1,16 @@
 package util
 
 import (
-	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"time"
 
+	"github.com/google/uuid"
 	rspecs "github.com/opencontainers/runtime-spec/specs-go"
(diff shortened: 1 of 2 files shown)


kolyshkin:
@dqminh @tianon @opencontainers/runtime-tools-maintainers PTAL

rhatdan:
LGTM

mrunalp (approved):
LGTM

Issue #723, opened 2021-09-23
Can't use TAP= to point make localvalidation to an alternate path for node-tap binary


dgibson:
Setting this environment variable from `make` apparently conflicts with a different usage of the `TAP` environment variable somewhere down in the stack:

```
$ sudo make RUNTIME=runc TAP=./node_modules/.bin/tap localvalidation
[...]

PR #743, opened 2022-01-06
Makefile: replace TAP with TAPTOOL

kolyshkin:
Apparently, TAP is used by node-tap, so when it is used to specify a
path to node-tap, this results in a failure:

	$ make TAP=/path/to/node-tap localvalidation
	...
	Error: Environment variable TAP must be set to 0 or 1 only

Fixes: #723

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 PREFIX ?= $(DESTDIR)/usr
 BINDIR ?= $(DESTDIR)/usr/bin
-TAP ?= tap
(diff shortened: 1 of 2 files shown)


dgibson (approved):
Looks good, thanks for making this change.

tianon (approved):

kolyshkin:
Can we please have a second LGTM on this, @opencontainers/runtime-tools-maintainers ?

hqhq (approved):

PR #744, opened 2022-01-13
Add syscall "statx" in seccomp to fix Operation not permitted

CooperLi:
syscall "statx" was first introduced by kernel at version 4.11, which
is a new implementation of stat, caller can request specific
information(file size only) to speed up the call. Docker has the statx
in seccomp config as well.

Signed-off-by: CooperLi <1866858@gmail.com>


tianon (approved):
LGTM

For reference, https://github.com/moby/moby/pull/36417 is the PR which added it in Docker's policy.

kolyshkin (approved):
LGTM

PR #746, opened 2022-04-22
Add generator support for FreeBSD

dfr:
Signed-off-by: Doug Rabson <dfr@rabson.org>

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -42,7 +42,7 @@ type ExportOptions struct {
 // New creates a configuration Generator with the default
 // configuration for the target operating system.
 func New(os string) (generator Generator, err error) {
-	if os != "linux" && os != "solaris" && os != "windows" {
+	if os != "linux" && os != "solaris" && os != "windows" && os != "freebsd" {
 		return generator, fmt.Errorf("no defaults configured for %s", os)
 	}
 
@@ -72,7 +72,7 @@ func New(os string) (generator Generator, err error) {
 		}
 	}
 
-	if os == "linux" || os == "solaris" {
+	if os == "linux" || os == "solaris" || os == "freebsd" {
 		config.Process.User = rspec.User{}
 		config.Process.Env = []string{
 			"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -237,6 +237,21 @@ func New(os string) (generator Generator, err error) {
 			},
 			Seccomp: seccomp.DefaultProfile(&config),
 		}
+	} else if os == "freebsd" {
+		config.Mounts = []rspec.Mount{
+			{
+				Destination: "/dev",
+				Type:        "devfs",
+				Source:      "devfs",
+				Options:     []string{"ruleset=4"},
+			},
+			{
+				Destination: "/dev/fd",
+				Type:        "fdescfs",
+				Source:      "fdesc",
+				Options:     []string{},
+			},
+		}
 	}
 
 	envCache := map[string]int{}


tianon (approved):
Seems OK to me, but I'm going to CC a few other OCI-FreeBSD-interested folks too. :sweat_smile:

@samuelkarp @gizahNL (gonna post this in Slack too)

tianon:
(FWIW, there is a `#freebsd` channel in the OCI Slack: https://opencontainers.org/community/overview/#chat)

dfr:
> (FWIW, there is a `#freebsd` channel in the OCI Slack: https://opencontainers.org/community/overview/#chat)

Thanks for the link but I wasn't able to join this slack - chat.opencontainers.org doesn't seem to exist:
```
$ host chat.opencontainers.org
Host chat.opencontainers.org not found: 3(NXDOMAIN)
```

dfr:
I took a look at the check failure above and as far as I can tell, the lint test fails even without this change.

tianon:
cc @crosbymichael @vishh @mrunalp @vbatts @dqminh @hqhq @liangchenye @Mashimiao @zhouhao3 @kolyshkin 

(@zhouhao3 it looks like we need to update your GitHub handle in the `MAINTAINERS` file :smile:)

dfr:
Could I please get a little feedback on this PR - this change is needed to support my buildah and podman ports to FreeBSD (both projects depend on runtime-tools to generate OCI configs). If necessary I can reduce it to just the part which allows generator to run at all but I do believe that having a minimal /dev is necessary for almost all containers.

Since opening the PR, I have been working through a long list of changes to modules which are vendored into buildah and podman and this is the last piece I need before I can move the ports forward to code review.

generate/generate.go
```diff
@@ -238,6 +238,16 @@ func New(os string) (generator Generator, err error) {
 			Seccomp: seccomp.DefaultProfile(&config),
 		}
 	}
+	if os == "freebsd" {
```
samuelkarp, line 241:
```suggestion
	} else if os == "freebsd" {
```

I think the pattern of 

```
if os == "linux" {
	// ...
}
if os == "freebsd" {
	// ...
}
```

is a bit odd, since those conditionals are mutually exclusive.  An if/else if pattern or a switch might be more appropriate here.

generate/generate.go
```diff
@@ -238,6 +238,16 @@ func New(os string) (generator Generator, err error) {
...
+		config.Mounts = []rspec.Mount{
```
samuelkarp, line 242: For containerd, I also added fdescfs by default.  Not sure whether you'd want that here (or if maybe it should be removed from containerd's default).

samuelkarp (commented):
I'm not a maintainer, but I've left a couple suggestions.

generate/generate.go
dfr: Good point, I added fdescfs as a default mount.

samuelkarp (dismissed):
Not a maintainer but LGTM

dfr:
I wanted to try to figure out why CI is broken. I believe this is broken on master and not due to this PR. Debugging it seems impossible since I can't run the actions here and haven't figured out the magic incantation to run them in my own fork.

gizahNL:
Not sure if generator (not very familiar with the code) would be the responsible party: what happens if the host OS is FreeBSD, but the container OS is Linux?
If the goal is to allow buildah and other consumers of the runtime-tools to run Linux containers on FreeBSD then at least some of the Linux specific compat mounts should be added.

dfr:
Running linux containers on freebsd is not a current goal but in that case, the caller would pass "linux" as the os argument to New which would trigger the linux sections and add those elements.

My main goal is running freebsd containers on freebsd hosts and my work in buildah and podman will pass "freebsd" as os which will add freebsd-specific elements such as the devfs mount.

giuseppe (dismissed):
LGTM

dfr:
Rebased to pick up the CI fixes

mrunalp (approved):
LGTM

PR #747, opened 2022-04-22
generate: NewFromSpec() remove deprecated comment

### LINKED ISSUE #720: NewFromSpec is deprecated, but required to avoid addEnv panic from nil envMap
The addition of `Generator.envMap` and its use for caching env vars requires it to be non-nil, or a call to `AddProcessEnv()` will panic within `addEnv()` at:

https://github.com/opencontainers/runtime-tools/blob/59cdde06764be8d761db120664020f0415f36045/generate/generate.go#L532

`envMap` is initialized in `New()` or `NewFromSpec()`. However, `NewFromSpec()` is marked deprecated:

```
// NewFromSpec creates a configuration Generator from a given
// configuration.
//
// Deprecated: Replace with:
//
//   generator := Generator{Config: config}
```

The replacement of creating the `Generator` struct directly will not initialize `envMap`, so a panic is possible.

Possibly use of `envMap` should be protected in case it is nil, and/or the non-deprecated method of creating with a config should permit initializing it?

Luap99:
The recommended replacement does not work because other code paths expect
an initialized envCacheMap.

Fixes #720

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -249,10 +249,6 @@ func New(os string) (generator Generator, err error) {
 
 // NewFromSpec creates a configuration Generator from a given
 // configuration.
-//
-// Deprecated: Replace with:
-//
-//   generator := Generator{Config: config}
 func NewFromSpec(config *rspec.Spec) Generator {
 	envCache := map[string]int{}
 	if config != nil && config.Process != nil {


rhatdan:
@giuseppe @mrunalp @AkihiroSuda @kolyshkin PTAL

giuseppe (approved):
LGTM

kolyshkin:
close/reopen to kick ci

kolyshkin (approved):
This reverts a tiny part of commit 84a62c6acd1576415e363c02702d1992f2ebcb07. After looking into all this I think this PR makes sense.

PR #748, opened 2022-06-13
Fix various golangci-lint typecheck errors ...

dfr:
... which cause CI to fail

Most of this seems to be trivial and/or false positive. I was unable to
fix typecheck errors around imports of "os/exec", "unicode" and
"unicode/utf8" and marked them as nolint. An alternative would be to
just disable the typecheck linter.

Signed-off-by: Doug Rabson <dfr@rabson.org>


dfr:
This problem is fixed in a better way in #750

PR #750, opened 2022-06-14
Use new golangci-lint

dmage:
The job `lint` has a lot of errors from typecheck that don't seem valid. The new golangci-lint with the new Go doesn't complain about the existing code.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,10 +13,13 @@ jobs:
   lint:
     runs-on: ubuntu-20.04
     steps:
-      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


dfr:
This is a much better approach than my attempt. LGTM but I'm not a maintainer.

tianon (approved):

dfr:
What is the next step for this PR? Is there any reason why it can't be merged?

giuseppe (approved):
LGTM

giuseppe:
> What is the next step for this PR? Is there any reason why it can't be merged?

merged :)

PR #755, opened 2022-09-08
generate, validate: isolate gojson* dependencies.

klihub:
Split out those few capability validation functions (`LastCap()`, `CapValid()`) which `Generator` depends on into a `validate/capabilities` subpackage of their own. This should prevent `github.com/xeipuuv/gojson*` from sneaking in to the dependencies of anyone who uses `Generator` for `OCI Spec` manipulation.

Those `gojsonschema` and related packages are often considered problematic because they do not seem to be actively maintained. At the time of this commit they received their last updates in 2018, 2019 and 2020. Excluding them from the dependency blast radius can lower the barrier for accepting/allowing import of generate.Generator.

### DIFF
--- a/validate/capabilities/validate_linux.go
+++ b/validate/capabilities/validate_linux.go
@@ -1,0 +1,16 @@
+package capabilities
+
+import (
+	"github.com/syndtr/gocapability/capability"
+)
+
+// LastCap return last cap of system
+func LastCap() capability.Cap {
+	last := capability.CAP_LAST_CAP
+	// hack for RHEL6 which has no /proc/sys/kernel/cap_last_cap
+	if last == capability.Cap(63) {
+		last = capability.CAP_BLOCK_SUSPEND
+	}
+
+	return last
+}
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -20,8 +20,8 @@ import (
 	"github.com/hashicorp/go-multierror"
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
 	osFilepath "github.com/opencontainers/runtime-tools/filepath"
+	capsCheck "github.com/opencontainers/runtime-tools/validate/capabilities"
 	"github.com/sirupsen/logrus"
-	"github.com/syndtr/gocapability/capability"
 
 	"github.com/opencontainers/runtime-tools/specerror"
 	"github.com/xeipuuv/gojsonschema"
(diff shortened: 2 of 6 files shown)


pohly:
> Those gojsonschema and related packages are often considered problematic because they do not seem to be actively maintained.

Even without that concern it is better to not pull in additional dependencies that aren't needed. For example, compilation becomes faster.

validate/validate.go
```diff
@@ -688,25 +688,7 @@ func (v *Validator) CheckAnnotations() (errs error) {
...
-			isValid = true
-			break
-		}
-	}
-
-	if !isValid {
-		return fmt.Errorf("invalid capability: %s", c)
-	}
-	return nil
+	return capsCheck.CapValid(c, hostSpecific)
```
pohly, line 691: A simpler solution is this:
```
// CapValid checks whether a capability is valid.
//
// DEPRECATED: Use [github.com/opencontainers/runtime-tools/validate/capabilities.CapValid] directly.
var CapValid = capsCheck.CapValid
```

Whether the alias should be marked as deprecated is open for debate, just as whether this alias is needed at all.

I think it makes sense, but I don't know how this package manages API stability.

validate/validate_linux.go
```diff
@@ -17,19 +17,14 @@ import (
...
 // LastCap return last cap of system
 func LastCap() capability.Cap {
-	last := capability.CAP_LAST_CAP
-	// hack for RHEL6 which has no /proc/sys/kernel/cap_last_cap
-	if last == capability.Cap(63) {
-		last = capability.CAP_BLOCK_SUSPEND
-	}
-
-	return last
+	return capsCheck.LastCap()
```
pohly, line 27: Same comment about an alias here.

validate/validate.go
pohly: True. Let's hear from the maintainers which style they prefer.

saschagrunert (approved):

pohly:
@saschagrunert: thanks. Is there a chance to get this into a tagged release soon(ish)?

saschagrunert:
@kolyshkin PTAL

validate/capabilities/validate_linux.go
```diff
@@ -0,0 +1,19 @@
+//go:build linux
+// +build linux
+
```
kolyshkin, line 3: nit: this is redundand, since there's already `_linux` in the file name.

validate/validate.go
```diff
@@ -687,26 +687,10 @@ func (v *Validator) CheckAnnotations() (errs error) {
 }
 
 // CapValid checks whether a capability is valid
+//
+// DEPRECATED: Use [github.com/opencontainers/runtime-tools/validate/capabilities.CapValid] directly.
```
kolyshkin, line 691: nit: `s/DEPRECATED: Use/Deprecated: use/`

kolyshkin: Also I'm not sure that `[ ]` in godoc makes any sense.

validate/validate_linux.go
```diff
@@ -11,26 +11,19 @@ import (
...
-	last := capability.CAP_LAST_CAP
-	// hack for RHEL6 which has no /proc/sys/kernel/cap_last_cap
-	if last == capability.Cap(63) {
-		last = capability.CAP_BLOCK_SUSPEND
-	}
-
-	return last
-}
+//
+// DEPRECATED: Use [github.com/opencontainers/runtime-tools/validate/capabilities.LastCap] directly.
```
kolyshkin, line 25: ditto

validate/validate.go
kolyshkin: I understand the concern that since this is a variable, it can be changed, but I guess we can live with it, since this stuff is deprecated and to be removed soon.

kolyshkin (changes requested):
Overall LGTM, left a few nits.

validate/capabilities/validate_linux.go
klihub: Removed build-tags.

validate/validate.go
klihub: Updated comment accordingly.

rhatdan:
@giuseppe @mrunalp PTAL

kolyshkin (approved):
LGTM

giuseppe (approved):
LGTM

elezar:
@saschagrunert @rhatdan what is still required to get these changes merged?

PR #767, opened 2023-05-04
dependency: integrate github.com/mndrix/tap-go as internal package

fmuyassarov:
Migrate the code of github.com/mndrix/tap-go as internal package with commit history as discussed in https://github.com/opencontainers/runtime-tools/pull/766#issuecomment-1498355392
Fixes subset of the dependency issues described in https://github.com/opencontainers/runtime-tools/issues/758.

I followed @thaJeztah suggestion of using git-filter-repo tool not to loose the git commit history while migrating the code and took the following steps.

```
git clone https://github.com/mndrix/tap-go.git
cd tap-go

git filter-repo 
      --path tap.go \
      --path yaml_yaml.go \
      --path yaml_json.go

git filter-repo \
     --path-rename yaml_json.go:util/yaml_json.go \
     --path-rename  yaml_yaml.go:util/yaml_yaml.go \
     --path-rename  tap.go:util/tap.go

# switch to the runtime-tools project
cd ../runtime-tools
git checkout -b integrate_tap_go
git fetch tap-go
git merge --allow-unrelated-histories --signoff -S tap-go/master
```
As the next step, I will replace function calls of the tap-go and remove imports of github.com/mndrix/tap-go as an external package.


fmuyassarov:
@thaJeztah shall I perhaps do the function call replacement on the same PR or would you prefer it to be on a separate PR?

thaJeztah:
Thanks!

I see this PR does not include the

- `test` directory and `Makefile` (perhaps something to consider; at least the tests to have some test coverage?)
- `README.md` (TBD, but could include a mention where this code was forked from)
- `LICENSE` (it has a different license than this repository, in which case I _guess_ the original license should be preserved); https://github.com/mndrix/tap-go/blob/master/LICENSE


> shall I perhaps do the function call replacement on the same PR or would you prefer it to be on a separate PR?

I think it's good to add those commits here;

- a commit that removes the `// import` comment in `tap.go` (and updates other references)

```patch
- package tap // import "github.com/mndrix/tap-go"
+ package tap
```

- a commit that replaces the uses (`github.com/mndrix/tap-go -> github.com/opencontainers/runtime-tools/tap`) and removes it from `go.mod` and `vendor`
- ☝️ with that included, this PR (as a whole) shows that it's migrating the old files to the new location (git / github shows them as a "move")

Other enhancements after that could be done in a follow-up

For the failing DCO sign-off, I think in the past we got an exemption from the legal team at CNCF, but it's probably worth checking if that's needed.

fmuyassarov:
> For the failing DCO sign-off, I think in the past we got an exemption from the legal team at CNCF, but it's probably worth checking if that's needed.

Thanks for the suggestions @thaJeztah. I have addressed above comments. Regarding DCO, would you mind to share who should I contact from the CNCF team?

fmuyassarov:
ping @thaJeztah

fmuyassarov:
Hi @thaJeztah . Please take a look when you have some time. Thanks

PR #771, opened 2023-09-06
update runtime-spec v1.1.0

thaJeztah:
- [x] depends on https://github.com/opencontainers/runtime-tools/pull/773

Update some dependencies to released versions;

### go.mod: github.com/opencontainers/runtime-spec v1.1.0

full diff: https://github.com/opencontainers/runtime-spec/compare/86290f6a00fb...v1.1.0

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/hashicorp/go-multierror v1.1.1
 	github.com/mndrix/tap-go v0.0.0-20171203230836-629fa407e90b
 	github.com/mrunalp/fileutils v0.5.0
-	github.com/opencontainers/runtime-spec v1.0.3-0.20220825212826-86290f6a00fb
+	github.com/opencontainers/runtime-spec v1.1.0
 	github.com/opencontainers/selinux v1.9.1
 	github.com/sirupsen/logrus v1.8.1
 	github.com/stretchr/testify v1.3.0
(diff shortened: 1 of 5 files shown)


thaJeztah:
@kolyshkin @mrunalp ptal 🤗

cyphar:
Can we do a release without this change, and then another release with this change? The lack of a release with `go.mod` means that you cannot use `go install` to get a `1.0.2`-compatible runtime-tool validator (0.9 doesn't have a go.mod so you pull the 1.1.0 spec which causes `runtime-tools` to think that it is a `1.1.0` spec validator). This is currently breaking umoci's CI, and will be an issue for historical users that need the older spec version.

cyphar:
Or, we can put the `golang.org/x/sys` update into a separate PR.

cyphar:
We could fork from 0.9 and release a 0.9.1 that "just" does the go.mod migration and upgrades to 1.0.2 (with cherry-picking if absolutely necessary(. That way, we won't have to worry about reverting things from `main` and `0.10.0` could be a strict upgrade.

That being said, given that 0.9 was also a release that had a non-released spec version I think it's not too crazy to just say that 0.10 can similarly just be a fruity intermediate runtime-spec version in keeping with all of the other releases of runtime-tools, and we will stop doing that in future releases.

cyphar:
Based on looking at the git logs, there was actually only one patch "needed" for us to support v1.0.2, so that approach seemed much simpler. See #774.

cyphar (approved):
On reflection, just doing this is a better idea.

kolyshkin (approved):
LGTM

PR #775, opened 2024-06-10
Move generate valid test

elezar:
This change moves the TestGenerateValid test to the validate package from the generate package. This decreases the transitive dependencies for clients of the generate package -- the more common use case.


kolyshkin (approved):
LGTM

PR #776, opened 2024-08-15
Add Go 1.22 and 1.23, rm old versions, bump CI actions, bump golangci-linter

kolyshkin:
Mostly a CI refresh. See commits for details.

Closes: #769

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,15 +11,15 @@ on:
 jobs:
 
   lint:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
-      - uses: actions/setup-go@v3
+      - uses: actions/setup-go@v5
         with:
-          go-version: 1.20.x
-      - uses: actions/checkout@v3
-      - uses: golangci/golangci-lint-action@v3
+          go-version: 1.x # latest
+      - uses: actions/checkout@v4
+      - uses: golangci/golangci-lint-action@v6
         with:
-          version: v1.51.2
+          version: v1.61
 
   commit:
     runs-on: ubuntu-22.04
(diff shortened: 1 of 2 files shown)


tianon (approved):

kolyshkin:
@opencontainers/runtime-tools-maintainers PTAL

.github/workflows/test.yml
```diff
@@ -44,14 +44,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: [1.19.x, 1.20.x]
+        go-version: [1.22.x, 1.23.x]
```
tianon, line 47: We should still include 1.19 here for go.mod, right? Or do something clever with the `go-version-file:` support in setup-go?

kolyshkin: I'm not sure it makes sense to test on older Go releases, unless we explicitly want to maintain compatibility with older Go versions (and for now I don't know a reason to do that).

go version in `go.mod` will raise naturally once we'll start using newer features.

kolyshkin: Besides, we're switching to go 1.21 as a minimum in #777

kolyshkin:
@opencontainers/runtime-tools-maintainers PTAL 🙏🏻

.github/workflows/test.yml
thaJeztah: Would it be ok to use;

- `min` version (go1.19 or go1.21 after #777)
- `latest`
- `latest - 1`

The last 2 being the currently supported versions, and the `min` version to verify that the minimum version specified in `go.mod` isn't lying.

kolyshkin: Updated; PTAL @thaJeztah

thaJeztah (approved):
LGTM, thanks!

thaJeztah:
Oh, interesting; I have a green check here, but I'm not in the maintainers list for this repo. Wondering if some permissions were set too relaxed?

kolyshkin:
> Oh, interesting; I have a green check here, but I'm not in the maintainers list for this repo. Wondering if some permissions were set too relaxed?

It looks like members of both runtime-tools-maintainers and runtime-spec-maintainers are admins here.

kolyshkin:
@tianon we need a "still lgtm" from you here 🧔🏻

kolyshkin:
@opencontainers/runtime-tools-maintainers PTAL 🙏🏻

tianon (approved):
Doh, sorry! 

To misquote one of my favorite people, "yep yep, looks like changes"

.github/workflows/test.yml
tianon: GitHub requires explicitly resolving conversations before merge; that's absolutely bonkers 🙃

PR #777, opened 2024-09-25
Switch to github.com/moby/sys/capability

kolyshkin:
Currently a draft pending
 - #776 merge;
 - moby/sys/capability@v0.3.0 release

----

The github.com/moby/sys/capability package is a fork of the original
one, which is apparently no longer maintained.

For changes since the fork took place, see
https://github.com/moby/sys/blob/main/capability/CHANGELOG.md

----

Note that "workaround for RHEL6" is removed for a number of reasons.
Feel free to choose the one you like the most, either is sufficient:

 1. /proc/sys/kernel/cap_last_cap is available since RHEL 6.7
    (kernel 2.6.32-573.el6), released 9 years ago (2015-07-22).

 2. It incorrectly returns CAP_BLOCK_SUSPEND (36), which was only added
    in kernel v3.5 and was never backported to RHEL6 kernels. The
    correct value for RHEL6 would be CAP_MAC_ADMIN (33).

 3. As far as upstream kernels go, /proc/sys/kernel/cap_last_cap was
    added in kernel v3.2, and a correct value depends on the kernel
    version. It could be CAP_WAKE_ALARM (35), added to kernel v3.0, or
    CAP_SYSLOG (34), added to kernel v2.6.38, or possibly a lesser value
    for even older kernels.

PR #778, opened 2024-10-02
replace hashicorp/go-multierror with errors.Join

thaJeztah:
- [x] depends on https://github.com/opencontainers/runtime-tools/pull/776


thaJeztah:
Ah, right, this requires go1.20 as minimum, which perhaps by now may be fine?

kolyshkin:
@thaJeztah #776 is merged; time to rebase

validate/validate.go
```diff
@@ -144,21 +144,21 @@ func (v *Validator) CheckJSONSchema() (errs error) {
 
 	url, err := JSONSchemaURL(strings.TrimSuffix(v.spec.Version, "-dev"))
 	if err != nil {
-		errs = multierror.Append(errs, err)
+		errs = errors.Join(errs, err)
```
thaJeztah, line 147: I need to check if I'm actually doing these right; I recently made a mistake with stdlib "multi-errors" because (IIRC) `errors.Join` is more an equivalent to `errors.Wrap`, and _nests_ a new error, not making it on the same "level" as other errors when using a loop.

kolyshkin:
With #777 we're switching to go 1.21 as a minimum so once merged we can merge this one as well.

thaJeztah:
Rebased after #777 was merged

kolyshkin (dismissed):
lgtm

tianon:
I think https://github.com/opencontainers/runtime-tools/graphs/contributors tells a pretty clear story about whether or not anyone is actively maintaining the code in this repository in 2025 (and frankly even as far back as 2019), and I'd honestly suggest that any project finding value from it should probably be thinking about either forking or adapting to some new library, if there's some code here that's still useful.

Is there a particular reason _you're_ looking at this, Seb?  Some transitive dependency pulling this into a project you maintain?

> oci-runtime-tool is a collection of tools for working with the OCI runtime specification.

Like honestly, this description is really `runc`, right?  I get that this repository does something slightly different, but to what end?  We had some lofty goals, but I don't think they were ever really realized.

thaJeztah:
Updated, but looks like CI started failing because golangci-lint doesn't like go1.26; 😂 

```
run golangci-lint
  Running [/home/runner/golangci-lint-2.6.2-linux-amd64/golangci-lint config path] in [/home/runner/work/runtime-tools/runtime-tools] ...
  Running [/home/runner/golangci-lint-2.6.2-linux-amd64/golangci-lint config verify] in [/home/runner/work/runtime-tools/runtime-tools] ...
  Running [/home/runner/golangci-lint-2.6.2-linux-amd64/golangci-lint run] in [/home/runner/work/runtime-tools/runtime-tools] ...
  panic: file requires newer Go version go1.26 (application built with go1.25) [recovered, repanicked]
```

thaJeztah:
@giuseppe @kolyshkin @cyphar rebased after golangci-lint was updated; PTAL 🤗

giuseppe (approved):
LGTM

PR #779, opened 2024-10-18
Create devcontainer.json

signal-coder:
go get github.com/tree-sitter/tree-sitter-javascript@latest


tianon:
Sorry, but this honestly looks like spam to me. If it's not, please comment with more justification for why we should consider adding this.

PR #801, opened 2025-11-26
Add support for the new FreeBSD section in the runtime config.

dfr:
The runtime-spec v1.3.0 release adds support for jail-based FreeBSD containers. This pull request adds support for these new fields in `generate` and `oci-runtime-tool`.

### DIFF
--- a/cmd/oci-runtime-tool/generate.go
+++ b/cmd/oci-runtime-tool/generate.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"net"
 	"os"
 	"runtime"
 	"strconv"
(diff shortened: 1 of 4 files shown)


kolyshkin (approved):
lgtm

dfr:
I just rebased this against the latest master branch. I would like to get these changes into runtime-tools to simplify improving the FreeBSD support in Podman and Buildah to follow the new FreeBSD config support in runtime-spec 1.3.0.

There are two commits here. The first adds helpers for the new freebsd config section to `generate/generate.go` and the second adds support for the new fields in `oci-runtime-tool`. Strictly speaking, I only need the first one for Podman/Buildah - would it be easier to review if I split this into two pull requests?

PR #803, opened 2026-01-07
Makefile: add EXTRA_LDFLAGS

sandy-lcq:
Add EXTRA_LDFLAGS to support add extra ldflags

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,8 +6,9 @@ BUILDTAGS=
 RUNTIME ?= runc
 COMMIT ?= $(shell git describe --dirty --long --always --tags 2> /dev/null)
 VERSION := ${shell cat ./VERSION}
-BUILD_FLAGS := -tags "$(BUILDTAGS)" -ldflags "-X main.gitCommit=$(COMMIT) -X main.version=$(VERSION)" $(EXTRA_FLAGS)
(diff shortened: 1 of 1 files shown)


kolyshkin:
@sandy-lcq why can't you use existing `EXTRA_FLAGS`? E.g.

```shell
make EXTRA_FLAGS="..."
```

sandy-lcq:
Hi,

with the old BUILD_FLAGS := -tags "$(BUILDTAGS)" -ldflags "-X main.gitCommit=$(COMMIT) -X main.version=$(VERSION)" $(EXTRA_FLAGS)

if user want to pass some config for -ldflags,  user can only pass like  EXTRA_FLAGS="-ldflags -buildid=",   in this way,   the latter option will overrides the first "-ldflag "-X main.gitCommit=$(COMMIT) -X main.version=$(VERSION)"",  unless I pass like 
EXTRA_FLAGS ="-ldflags -X main.gitCommit=$(COMMIT) -X main.version=$(VERSION)" -buildid=",   so I think maybe add another EXTRA_LDFLAGS is better. So user can only set EXTRA_LDFLAGS="-buildid="

kolyshkin (approved):
Makes sense.

thaJeztah (approved):
LGTM

PR #808, opened 2026-05-22
fix: correct lookup for environment variables in cache

### LINKED ISSUE #798: Bug in createEnvCacheMap()
When using `NewFromTemplate()`, or other similar functions, a call is made to `createEnvCacheMap()` to populate a map of the environment variables, used for later lookups:

```
func createEnvCacheMap(env []string) map[string]int {
	envMap := make(map[string]int, len(env))
	for i, val := range env {
		envMap[val] = i
	}
	return envMap
}
```

Unfortunately, this is keying the map on the name/value pair for the environment variable, rather than just the name, so when environment variables are later added with the same name, the updated variables are not added correctly:

```
// addEnv looks through adds ENV to the Process and checks envMap for
// any duplicates
// This is called by both AddMultipleProcessEnv and AddProcessEnv
func (g *Generator) addEnv(env, key string) {
	if idx, ok := g.envMap[key]; ok {
		// The ENV exists in the cache, so change its value in g.Config.Process.Env
		g.Config.Process.Env[idx] = env
	} else {
		// else the env doesn't exist, so add it and add it's index to g.envMap
		g.Config.Process.Env = append(g.Config.Process.Env, env)
		g.envMap[key] = len(g.Config.Process.Env) - 1
	}
}
```

Here, the assumption is made that `envMap` is keyed by the name only.

Really, `createEnvCacheMap()` should perform a `strings.SplitN(val, "=", 2)` and use the name only for its key.

Skywalkr-dev:
The current implementation keys the environment cache by the full "NAME=VALUE" string. This causes the generator to treat updates to existing variables as new entries, leading to duplicate environment variables in the final configuration.

This change switches the cache key to use only the variable name, ensuring that `AddProcessEnv` correctly updates existing values instead of appending them.

FIXES: #798 
Signed-off-by: Skywalkr-dev <snaveenbharath2005@gmail.com>

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -322,6 +322,7 @@ func NewFromTemplate(r io.Reader) (Generator, error) {
 func createEnvCacheMap(env []string) map[string]int {
 	envMap := make(map[string]int, len(env))
 	for i, val := range env {
+		val, _, _ = strings.Cut(val, "=")
 		envMap[val] = i
 	}
 	return envMap
--- a/generate/generate_test.go
+++ b/generate/generate_test.go
@@ -155,3 +155,41 @@ func TestMultipleEnvCaching(t *testing.T) {
 	g.AddMultipleProcessEnv([]string{})
 	assert.Equal(t, []string(nil), g.Config.Process.Env)
 }
+
+func TestEnvCachingOverrides(t *testing.T) {
+	// Test overriding default ENV variables to verify createEnvCacheMap correctly extracts keys
+	g, err := generate.New("linux")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Override existing default ENVs
+	g.AddProcessEnv("TERM", "vt100")
+	g.AddProcessEnv("PATH", "/minimal/path")
+	g.AddProcessEnv("FOO", "bar")
+
+	expected := []string{
+		"PATH=/minimal/path",
+		"TERM=vt100",
+		"FOO=bar",
+	}
+	assert.Equal(t, expected, g.Config.Process.Env)
+
+	// Test AddMultipleProcessEnv with overrides
+	g, err = generate.New("linux")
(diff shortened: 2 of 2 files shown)


kolyshkin (changes requested):
You need to fix the commit subject.

Skywalkr-dev:
@kolyshkin sorry about the commit, does the current commit subject work

generate/generate.go
```diff
@@ -322,7 +322,13 @@ func NewFromTemplate(r io.Reader) (Generator, error) {
 func createEnvCacheMap(env []string) map[string]int {
 	envMap := make(map[string]int, len(env))
 	for i, val := range env {
-		envMap[val] = i
+    name, _, found := strings.Cut(val, "=")
+        if found {
+            envMap[name] = i
+        } else {
+            // Fallback for variables without an '=' (rare, but possible)
+            envMap[val] = i
```
kolyshkin: nit: this can be simplified to a one-line patch:
```diff
 for i, val := range env {
+  val, _, _ = strings.Cut(val, "=")
   envMap[val] = i
```

Skywalkr-dev: > nit: this can be simplified to a one-line patch:
> 
> ```diff
>  for i, val := range env {
> +  val, _, _ = strings.Cut(val, "=")
>    envMap[val] = i
> ```

yup, that's much cleaner. I've updated the PR to use strings.Cut as suggested.

Skywalkr-dev:
Hey, @kolyshkin any update on this? It’s currently the blocker for the NRI environment variable regression in containerd. 
Appreciate your time.

kolyshkin:
@Skywalkr-dev the patch LGTM, can we have a unit test added (which fails before the fix)?

Skywalkr-dev:
@kolyshkin does the unit test seem fine ?

PR #809, opened 2026-05-23
generate: fix env cache keys

### LINKED ISSUE #798: Bug in createEnvCacheMap()

chansuke:
Closes #798 

## Summary

- Fix `createEnvCacheMap()` to cache environment variables by name instead of the full `name=value` pair
- Add regression coverage for env overrides after loading an existing spec through `NewFromTemplate()` and `NewFromSpec()`

## Testing

- `CGO_ENABLED=0 go test ./generate`

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -322,7 +322,8 @@ func NewFromTemplate(r io.Reader) (Generator, error) {
 func createEnvCacheMap(env []string) map[string]int {
 	envMap := make(map[string]int, len(env))
 	for i, val := range env {
-		envMap[val] = i
+		parts := strings.SplitN(val, "=", 2)
+		envMap[parts[0]] = i
 	}
 	return envMap
 }
--- a/generate/generate_test.go
+++ b/generate/generate_test.go
@@ -4,8 +4,10 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"testing"
 
+	rspec "github.com/opencontainers/runtime-spec/specs-go"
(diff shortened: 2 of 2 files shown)


elezar:
This is a duplicate of #808

chansuke:
Closing this as it's a duplicate of https://github.com/opencontainers/runtime-tools/pull/808.

